PR #16, opened 2024-04-16
fix identifying params

efriis:
renders in langsmith: https://smith.langchain.com/public/35edf4ec-dabb-4560-b83e-508d6185b9fa/r?runtab=2

### DIFF
--- a/libs/aws/langchain_aws/llms/bedrock.py
+++ b/libs/aws/langchain_aws/llms/bedrock.py
@@ -444,7 +444,7 @@ class BedrockBase(BaseLanguageModel, ABC):
         _model_kwargs = self.model_kwargs or {}
         return {
             "model_id": self.model_id,
-            "provider": self.provider,
+            "provider": self._get_provider(),
             "stream": self.streaming,
             "guardrails": self.guardrails,
             **_model_kwargs,
--- a/libs/aws/langchain_aws/llms/bedrock.py
+++ b/libs/aws/langchain_aws/llms/bedrock.py
@@ -19,9 +19,9 @@ from langchain_core.callbacks import (
     AsyncCallbackManagerForLLMRun,
     CallbackManagerForLLMRun,
 )
-from langchain_core.language_models.llms import LLM
+from langchain_core.language_models import LLM, BaseLanguageModel
 from langchain_core.outputs import GenerationChunk
-from langchain_core.pydantic_v1 import BaseModel, Extra, Field, root_validator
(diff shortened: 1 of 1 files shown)


libs/aws/langchain_aws/llms/bedrock.py
```diff
@@ -440,11 +440,14 @@ def validate_environment(cls, values: Dict) -> Dict:
...
+    def _identifying_params(self) -> Dict[str, Any]:
         _model_kwargs = self.model_kwargs or {}
         return {
-            **{"model_kwargs": _model_kwargs},
+            "model_id": self.model_id,
+            "provider": self._get_provider(),
+            "stream": self.streaming,
+            "guardrails": self.guardrails,
+            **_model_kwargs,
         }
```
3coins, line 451: With this change, would we need to call `_get_provider` in any of the other functions or can just use `self.provider`? For example:
https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/llms/bedrock.py#L462
https://github.com/langchain-ai/langchain-aws/blob/688a3a525594a592c6bae5e8009ad1da2bbb95f1/libs/aws/langchain_aws/llms/bedrock.py#L517

libs/aws/langchain_aws/llms/bedrock.py
```diff
@@ -617,7 +620,8 @@ def _prepare_input_and_invoke_stream(
 
             # stop sequence from _generate() overrides
             # stop sequences in the class attribute
-            _model_kwargs[self.provider_stop_sequence_key_name_map.get(provider)] = stop
+            if k := self.provider_stop_sequence_key_name_map.get(provider):
+                _model_kwargs[k] = stop
```
3coins, line 625: This is great. Ty.

libs/aws/langchain_aws/llms/bedrock.py
efriis: I think fixed! This might have been a comment from before I changed `self.provider` to `self._get_provider()`

efriis: Also just calling out - this moves the model_kwargs into a flat structure of invocation params to match most other model providers. Let me know if that sounds good to you.

efriis: I think I misunderstood the question. Still need to use `self._get_provider()` everywhere because provider isn't always set.

Although Likely worth adding a root validator that populates that field based on model string.

PR #61, opened 2024-05-28
Fix stop sequence key for Mistral model in BedrockBase

spigarmx:
Correct the stop sequence key for the Mistral model from 'stop_sequences' to 'stop' in the provider_stop_sequence_key_name_map. This ensures proper handling of stop sequences for Mistral.


spigarmx:
@3coins Is it possible to trade this PR? mistral is unusable like this.
https://github.com/langchain-ai/langchain/pull/20709

PR #68, opened 2024-06-04
[Langchain-AWS-67] Refactor ChatBedrock._combine_llm_outputs() to mat…

adarmiento:
…ch other Langchain model output

 # Changes

* Replaced `usage` with `token_usage`
* Replaced `model_id` with `model_name`

These change make the Runnable output of Bedrock hosted models consistent with other models supported by Langchain, such as OpenAI chat models.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -445,7 +445,7 @@ class ChatBedrock(BaseChatModel, BedrockBase):
                 **params,
             )
 
-        llm_output["model_id"] = self.model_id
+        llm_output["model_name"] = self.model_id
         return ChatResult(
             generations=[
                 ChatGeneration(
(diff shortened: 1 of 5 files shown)


libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -460,11 +460,11 @@ def _combine_llm_outputs(self, llm_outputs: List[Optional[dict]]) -> dict:
         final_output = {}
         for output in llm_outputs:
             output = output or {}
-            usage = output.get("usage", {})
-            for token_type, token_count in usage.items():
+            token_usage = output.get("usage", {})
+            for token_type, token_count in token_usage.items():
                 final_usage[token_type] += token_count
             final_output.update(output)
-        final_output["usage"] = final_usage
```
baskaryan, line 467: this is a breaking change.

what we probably want to do is use the new AIMessage.token_usage attribute which is meant to be the standardized way to record token usage: https://github.com/langchain-ai/langchain/pull/21944

ccurme: +1, the standard Langchain is moving toward is the [usage_metadata](https://api.python.langchain.com/en/latest/_modules/langchain_core/messages/ai.html#AIMessage) attribute on AIMessage, which is just a dict with keys `"input_tokens"`, `"output_tokens"`, `"total_tokens"`. If you're able to access the usage data within the `_generate` method, you can construct this dict and then populate `usage_metadata` on the `AIMessage` it generates. Note that we'd need to bump langchain-core >= 0.2.2 in the dependencies.

ccurme:
Believe usage_metadata is now correctly populated on output messages. Please let me know if more development is needed here.

PR #76, opened 2024-06-15
Updates __init__.py files for an easier import of ChatBedrockConverse model

userlerueda:
Since this is my first time submitting a PR here, I am not sure if for the `beta` models we add imports into __init__.py.

If so, kindly accept my PR.


3coins:
@userlerueda 
Can you provide some sample code that showcases the problem. The `_format_tools` duplicate calls that you have called out are in 2 different methods.

PR #77, opened 2024-06-16
Fix llama3 SSE int usage values

dannykok:
Currently when streaming from bedrock llama3, the usage numbers feedback from streaming chunks will cause error in Langchain core merge_dicts function.

Change:
- format `prompt_token_count` and `generation_token_count` from int to list[int]

This should enable the streaming from bedrock llama3 models.


dannykok:
Hi, can someone help review this PR please?

Issue #80, opened 2024-06-21
Recursion depth error: Tool calling with streaming model


munday-tech:
When setting the Bedrock model to `streaming=True` and tool calls enabled, it causes a recursion depth error due to stream and generate constantly pointing back to each other.

```
  File "C:\Users\..\AppData\Local\pypoetry\Cache\virtualenvs\..\lib\site-packages\langchain_aws\chat_models\bedrock.py", line 432, in _stream
    result = self._generate(
  File "C:\Users\..\AppData\Local\pypoetry\Cache\virtualenvs\..\lib\site-packages\langchain_aws\chat_models\bedrock.py", line 501, in _generate
    for chunk in self._stream(messages, stop, run_manager, **kwargs):
  File "C:\Users\..\AppData\Local\pypoetry\Cache\virtualenvs\..\lib\site-packages\langchain_aws\chat_models\bedrock.py", line 432, in _stream
    result = self._generate(
  File "C:\Users\..\AppData\Local\pypoetry\Cache\virtualenvs\..\lib\site-packages\langchain_aws\chat_models\bedrock.py", line 501, in _generate
    for chunk in self._stream(messages, stop, run_manager, **kwargs):
  File "C:\Users\..\AppData\Local\pypoetry\Cache\virtualenvs\..\lib\site-packages\langchain_aws\chat_models\bedrock.py", line 432, in _stream
    result = self._generate(
  File "C:\Users\..\AppData\Local\pypoetry\Cache\virtualenvs\..\lib\site-packages\langchain_aws\chat_models\bedrock.py", line 501, in _generate
    for chunk in self._stream(messages, stop, run_manager, **kwargs):
  File "C:\Users\..\AppData\Local\pypoetry\Cache\virtualenvs\..\lib\site-packages\langchain_aws\chat_models\bedrock.py", line 427, in _stream
    provider = self._get_provider()
  File "C:\Users\..\AppData\Local\pypoetry\Cache\virtualenvs\..\lib\site-packages\langchain_aws\llms\bedrock.py", line 570, in _get_provider
    if self.model_id.startswith("arn"):
RecursionError: maximum recursion depth exceeded while calling a Python object
```

`llm = ChatBedrock(client=bedrock_client, model_id=model, model_kwargs=model_kwargs, streaming=True)`

```
def _stream(
...

        if "claude-3" in self._get_model():
            if _tools_in_params({**kwargs}):
                result = self._generate(
[...]

PR #81, opened 2024-06-21
Bug: Fix recursion error when using streaming with tools

munday-tech:
Added additional param to allow independent turning off for streaming to ensure recursion errors don't occur when calling `generate` from `stream` with `streaming=True` on BedrockChat

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -430,7 +430,7 @@ class ChatBedrock(BaseChatModel, BedrockBase):
         if "claude-3" in self._get_model():
             if _tools_in_params({**kwargs}):
                 result = self._generate(
-                    messages, stop=stop, run_manager=run_manager, **kwargs
+                    messages, stop=stop, run_manager=run_manager, stream=False, **kwargs
                 )
                 message = result.generations[0].message
                 if isinstance(message, AIMessage) and message.tool_calls is not None:
@@ -488,15 +488,17 @@ class ChatBedrock(BaseChatModel, BedrockBase):
         messages: List[BaseMessage],
         stop: Optional[List[str]] = None,
         run_manager: Optional[CallbackManagerForLLMRun] = None,
+        stream: Optional[bool] = None,
         **kwargs: Any,
     ) -> ChatResult:
+        should_stream = stream if stream is not None else self.streaming
         completion = ""
         llm_output: Dict[str, Any] = {}
         tool_calls: List[Dict[str, Any]] = []
         provider_stop_reason_code = self.provider_stop_reason_key_map.get(
             self._get_provider(), "stop_reason"
         )
-        if self.streaming:
+        if should_stream:
             response_metadata: List[Dict[str, Any]] = []
             for chunk in self._stream(messages, stop, run_manager, **kwargs):
                 completion += chunk.text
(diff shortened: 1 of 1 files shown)


munday-tech:
Fixes #80

laithalsaadoon:
@3coins @bigbernnn can you please have a look? I confirmed this PR is working

laithalsaadoon (approved):
LGTM

bigbernnn:
@munday-tech Looks good! Please fix small linting error.

munday-tech:
@bigbernnn Should be fixed now.

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -439,7 +439,7 @@ def _stream(
         if "claude-3" in self._get_model():
             if _tools_in_params({**kwargs}):
                 result = self._generate(
-                    messages, stop=stop, run_manager=run_manager, **kwargs
+                    messages, stop=stop, run_manager=run_manager, stream=False, **kwargs
```
3coins, line 442: Isn't this hardcoding and disabling streaming for all tool calls? I might be missing some context here, is there a documented issue regarding this in Bedrock/Anthropic or documentation that refers to this?

deepakdalakoti:
Is there any update on this? I am facing the same issue

libs/aws/langchain_aws/chat_models/bedrock.py
3coins, line 442: Rather than adding a new parameter, can try adding to the kwargs. 
```python
kw = {**kwargs, "stream": False}
result = self._generate(
    messages, stop=stop, run_manager=run_manager, **kw
)
```

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -497,19 +497,23 @@ def _generate(
         messages: List[BaseMessage],
         stop: Optional[List[str]] = None,
         run_manager: Optional[CallbackManagerForLLMRun] = None,
+        stream: Optional[bool] = None,
```
3coins, line 500:
```suggestion
```

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -497,19 +497,23 @@ def _generate(
...
         **kwargs: Any,
     ) -> ChatResult:
+        should_stream = stream if stream is not None else self.streaming
```
3coins, line 503:
```suggestion
should_stream = (
    kwargs.pop("stream", self.streaming) if kwargs else self.streaming
)
```

3coins (changes requested):
@munday-tech 
Thanks for submitting this fix. Added suggestion to avoid adding a new parameter to `_generate`. Can you try this, and check if this works.

deepakdalakoti:
I did a quick test with this solution. While this fixes the recursion bug, the tokens are still not being streamed (for claude-3 models)

rsgrewal-aws:
@munday-tech  could you review the changes requested and also confirm if the tokens get streamed with this fix ?

3coins:
Working on another PR, that allows streaming to work as expected. The change here uses interplay between _stream and _generate and does not return the chunks independently. See https://github.com/langchain-ai/langchain-aws/pull/119

3coins:
Fixed by #119

PR #82, opened 2024-06-24
Bagatur/guardrails support

baskaryan:


3coins:
@supreetk
Can you review this PR, and make sure to test the changes. Thanks.

3coins:
@supreetk
Can you review this PR. Thanks.

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
```diff
@@ -87,7 +87,14 @@ def test__messages_to_bedrock() -> None:
     messages = [
         SystemMessage(content="sys1"),
         SystemMessage(content=["sys2"]),
-        SystemMessage(content=[{"text": "sys3"}, {"type": "text", "text": "sys4"}]),
+        SystemMessage(
+            content=[
+                {"text": "sys3"},
+                {"type": "text", "text": "sys4"},
+                {"guardContent": {"text": {"text": "sys5"}}},
+                {"type": "guard_content", "text": "sys6"},
```
3coins, line 95: The first form is understood, and matches with Bedrock guardrails documentation. Are we introducing the second form with `{"type": "guard_content"}`? Referring to this doc from the converse API.
https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails-use-converse-api.html

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -314,6 +314,24 @@ class Joke(BaseModel):
     config: Any = None
     """An optional botocore.config.Config instance to pass to the client."""
 
+    guardrail_config: Optional[Dict[str, Any]] = Field(None, alias="guardrails")
```
3coins, line 317: Should we limit this to specified values in the API than an arbitrary dict?
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_GuardrailConfiguration.html

3coins (commented):
@baskaryan 
Thanks for making these changes. Some minor comments about the changes.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -314,6 +314,24 @@ class Joke(BaseModel):
...
+    """Configuration information for a guardrail that you want to use in the request."""
+
+    additional_model_request_fields: Optional[Dict[str, Any]] = None
```
supreetkt, line 320: Is there a reason why `max_tokens`, `stop_sequences` etc. being passed as class variables and not a part of the `inference_config` dict? If its an `inference_config` dict, you can limit them to these specified values (`max_tokens`, `stop_sequences` etc.).

We have `additional_model_request_fields` and `additional_model_response_field_paths` here which are the inputs to the converse API call as well. So having `inference_config` would be in line with the converse API call and these other fields.

supreetkt: Also, if its `inference_config` instead of the individual inference parameters, then any changes to add or remove these parameters from `inferenceConfig` in the converse API call wouldn't have to be handled by us.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
baskaryan: for less common params like this generally prefer to do less validation on the langchain side and let boto handle validation. mainly because if we validate here then we need to actively stay up to date, eg add support for new parameters every time they're added. don't feel super strongly though

libs/aws/langchain_aws/chat_models/bedrock_converse.py
baskaryan: this is done for a nicer devx and for interoperatbility. max_tokens, stop_sequences, [few other params](https://github.com/langchain-ai/langchain/issues/20085) are standardized across all the main models, making it extra easy to swam models in your code

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
baskaryan: yea we are introducing {"type": "guard_content", ...}. this is meant to imitate the anthropic format, since we support anthropic style blocks otherwise. but since this isn't a type actually supported by anthropic id be ok dropping it if we don't like it

libs/aws/langchain_aws/chat_models/bedrock_converse.py
3coins: These properties are pretty common in all providers, it's ok to have these at the root. Alternately, `inferenceConfig` is also supported if users want to pass them in that fashion.

3coins (approved):
@baskaryan 
Looks good. We should go ahead and merge. 

One thing to note is that the usage from ChatBedrock is different which uses `guardrails` to set this config. We might want to align with the changes here and make `guardrail_config` as primary attribute, with `guardrails` as an alias there. But we can handle that in a separate PR. Thanks again on working on this.

rsgrewal-aws:
hello Team, if the entry point will be ChatBedrock with converse api -- what will it mean for Guardrails config ? will it work as is ?

3coins:
@rsgrewal-aws 
I believe it will not work at the moment, as ChatBedrock is using a different field, and flattening the guardrail properties to pass to the invoke function. We will need few more changes on ChatBedrock to make it work. For example, this is currently failing in my test.

```python
def test_guardrails() -> None:
    params = {
        "region_name": "us-west-2",
        "model_id": "anthropic.claude-3-sonnet-20240229-v1:0",
        "guardrails": {
            "guardrailIdentifier": "e7esbceow153",
            "guardrailVersion": "1",
            "trace": "enabled",
        },
        "beta_use_converse_api": True
    }
    chat_model = ChatBedrock(**params)  # type: ignore[arg-type]
    messages = [
        HumanMessage(
            content=[
                "Create a playlist of 2 heavy metal songs.",
                {
                    "guardContent": {
                        "text": {"text": "Only answer with a list of songs."}
                    }
                },
            ]
        )
    ]
    response = chat_model.invoke(messages)

    assert (
        response.content == "Sorry, I can't answer questions about heavy metal music."
    )
    assert response.response_metadata["stopReason"] == "guardrail_intervened"
    assert response.response_metadata["trace"] is not None

    stream = chat_model.stream(messages)
    response = next(stream)
    for chunk in stream:
        response += chunk

    assert (
        response.content[0]["text"]  # type: ignore[index]
        == "Sorry, I can't answer questions about heavy metal music."
    )
    assert response.response_metadata["stopReason"] == "guardrail_intervened"
    assert response.response_metadata["trace"] is not None

```

**Error**
```
E           botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the Converse operation: The model returned the following errors: The guardrail can't assess the content in the guardContent field. The guardrail configuration is missing. Specify the guardrailConfig field and try again.
```

I debugged this a bit, and not sure why the `_generate` in ChatBedrock is not passing the guardrails config in `kwargs`.
```
# ChatBedrock instance
self = ChatBedrock(client=<botocore.client.BedrockRuntime object at 0x109438250>, region_name='us-west-2', model_id='anthropi...rails={'guardrailIdentifier': 'e7esbceow153', 'guardrailVersion': '1', 'trace': 'enabled'}, beta_use_converse_api=True)

# Debugged params to the _generate call
messages = [HumanMessage(content=['Create a playlist of 2 heavy metal songs.', {'guardContent': {'text': {'text': 'Only answer with a list of songs.'}}}])]
stop = None, run_manager = <langchain_core.callbacks.manager.CallbackManagerForLLMRun object at 0x108d7c940>, kwargs = {}
```

3coins:
@baskaryan 
Can you take a look at the recent commit, this should fix calling the ChatBedrock with guardrails config. I have added an integration test to verify this.

supreetkt:
Maybe this can be a part of a separate PR, but would advise adding docstrings to all util methods. Otherwise LGTM.

libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
```diff
@@ -28,3 +30,51 @@ def standard_chat_model_params(self) -> dict:
...
+@pytest.mark.skip(reason="Needs guardrails setup to run.")
+def test_guardrails() -> None:
+    params = {
+        "region_name": "us-west-2",
+        "model": "anthropic.claude-3-sonnet-20240229-v1:0",
+        "temperature": 0,
+        "max_tokens": 100,
+        "stop": [],
+        "guardrail_config": {
+            "guardrailIdentifier": "e7esbceow153",
```
baskaryan, line 44: is this specific to a user or will it work with our creds as well?

PR #91, opened 2024-06-28
[WIP] Bedrock Agents Runnable

3coins:
# Description
This PR introduces a new Bedrock Agents Runnable that allows using [Bedrock Agents](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-returncontrol.html) with return of control functions as tools.

## Usage
Define tools
```python

@tool("AssetDetail::getAssetValue")
def getAssetValue(asset_holder_id: str = " ") -> str:
    """Get the asset value for an owner id"""
    return f"The total asset value for {asset_holder_id} is 100K"


@tool
def getMortgageRate(asset_holder_id: str = " ", asset_value: str = " ") -> str:
    """Get the mortgage rate based on asset value"""
    return (
        f"The mortgage rate for {asset_holder_id} "
        f"with asset value of {asset_value} is 8.87%"
    )
```

Create the agent and use with AgentExecutor
```python
from langchain.agents import AgentExecutor

agent = BedrockAgentsRunnable.create_agent(
    {"agent_id": "UKYYJIV1O1", "enable_trace": True}
)
tools = [getAssetValue, getMortgageRate]
agent_executor = AgentExecutor(agent=agent, tools=tools)  # type: ignore[arg-type]

output = agent_executor.invoke(
    {"input": "what is my mortgage rate for id AVC-1234"}
)

assert output["output"] == "The mortgage rate for id AVC-1234 is 8.87%"
```

## Notes
This change follows a strategy similar to the [OpenAIAssistantRunnable](https://github.com/langchain-ai/langchain/blob/800b0ff3b9e7fb5712445e8e4182107135f9f5f6/libs/langchain/langchain/agents/openai_assistant/base.py#L145) to create an agent. However, it's still missing the callback related code that should be plugged in before merging. This code also assumes that the agent and the corresponding actions have been setup and prepared to call before calling; this setup could be automated in the agent creation or validation function, but missing here.

### DIFF
--- a/libs/aws/langchain_aws/agents/base.py
+++ b/libs/aws/langchain_aws/agents/base.py
@@ -1,0 +1,152 @@
+from __future__ import annotations
+
+import json
+import uuid
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+import boto3
+import botocore
+from langchain_core.agents import AgentAction, AgentFinish
+from langchain_core.messages import AIMessage
+from langchain_core.pydantic_v1 import Field, root_validator
+from langchain_core.runnables import RunnableConfig, RunnableSerializable, ensure_config
+
+_DEFAULT_ACTION_GROUP_NAME = "DEFAULT_AG_"
+_TEST_AGENT_ALIAS_ID = "TSTALIASID"
+
+
+def get_bedrock_agents_client() -> None:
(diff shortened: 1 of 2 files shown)


3coins:
Closing in favor of #152

PR #104, opened 2024-07-05
Adding bedrock metadata filter and search type parameters

ravediamond:
### Description

This pull request enhances the functionality of the`AmazonKnowledgeBasesRetriever` class by adding optional parameters for the following features (already implemented in [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-agent-runtime/client/retrieve.html)):

- added `filter` parameter in `VectorSearchConfig` to filter on metadata tags.
- added `overrideSearchType` in `VectorSearchConfig` to override the default search type to be able to choose between 'Default', 'Hybrid' or 'Semantic'

Here an example answer before the change:
```
retriever = AmazonKnowledgeBasesRetriever(
    knowledge_base_id="KNOWLEDGE_BASE_ID,
    retrieval_config={
        "vectorSearchConfiguration": {
            "numberOfResults": 4
        }
    },
)
```

And now after the change:

```
retriever = AmazonKnowledgeBasesRetriever(
    knowledge_base_id="KNOWLEDGE_BASE_ID,
    retrieval_config={
        "vectorSearchConfiguration": {
            "numberOfResults": 4,
            "filter": {
                "equals": {"key": "type", "TAG1"}
            },
            "overrideSearchType": "HYBRID"
        }
    },
)
```

### Change
- added `SearchFilter` class to handle parameter filtering
- added `filter` parameter in `VectorSearchConfig`
- added `overrideSearchType` in `VectorSearchConfig` 

### Motivation
This will allow users to better use the full capacity of Bedrock Knowledge Base.

### Additional Notes
- This change maintains backwards compatibility with existing usage.
- The `nextToken` feature when with very long results has not been implemented.

### DIFF
--- a/libs/aws/langchain_aws/retrievers/bedrock.py
+++ b/libs/aws/langchain_aws/retrievers/bedrock.py
@@ -160,12 +160,11 @@ class AmazonKnowledgeBasesRetriever(BaseRetriever):
         query: str,
         *,
         run_manager: CallbackManagerForRetrieverRun,
-        next_token: Optional[str] = None,
     ) -> List[Document]:
         response = self.client.retrieve(
             retrievalQuery={"text": query.strip()},
             knowledgeBaseId=self.knowledge_base_id,
-            nextToken=next_token,
+            nextToken=None,
             retrievalConfiguration=self.retrieval_config.dict(),
         )
         results = response["retrievalResults"]
@@ -185,36 +184,3 @@ class AmazonKnowledgeBasesRetriever(BaseRetriever):
             )
 
         return self._filter_by_score_confidence(docs=documents)
-
-    def _get_relevant_documents_with_token(
-        self,
-        query: str,
-        *,
-        run_manager: CallbackManagerForRetrieverRun,
-        next_token: Optional[str] = None,
-    ) -> Tuple[List[Document], Optional[str]]:
-        response = self.client.retrieve(
-            retrievalQuery={"text": query.strip()},
-            knowledgeBaseId=self.knowledge_base_id,
-            nextToken=next_token or self.retrieval_config.nextToken,
-            retrievalConfiguration=self.retrieval_config.dict(),
-        )
-        new_next_token = response.get("nextToken", None)
-        results = response["retrievalResults"]
-        documents = []
-        for result in results:
-            content = result["content"]["text"]
-            result.pop("content")
-            if "score" not in result:
-                result["score"] = 0
-            if "metadata" in result:
-                result["source_metadata"] = result.pop("metadata")
-            documents.append(
-                Document(
-                    page_content=content,
-                    metadata=result,
-                )
-            )
-
-        filtered_documents = self._filter_by_score_confidence(docs=documents)
-        return (filtered_documents, new_next_token)
--- a/libs/aws/langchain_aws/retrievers/bedrock.py
+++ b/libs/aws/langchain_aws/retrievers/bedrock.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Literal, Optional, Tuple, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 import boto3
 from botocore.client import Config
--- a/libs/aws/langchain_aws/retrievers/bedrock.py
+++ b/libs/aws/langchain_aws/retrievers/bedrock.py
@@ -158,7 +158,6 @@ class AmazonKnowledgeBasesRetriever(BaseRetriever):
         response = self.client.retrieve(
             retrievalQuery={"text": query.strip()},
             knowledgeBaseId=self.knowledge_base_id,
-            nextToken=None,
             retrievalConfiguration=self.retrieval_config.dict(),
         )
         results = response["retrievalResults"]
--- a/libs/aws/langchain_aws/retrievers/bedrock.py
+++ b/libs/aws/langchain_aws/retrievers/bedrock.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, Optional, Union
 
 import boto3
 from botocore.client import Config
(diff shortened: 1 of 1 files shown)


ravediamond:
@3coins can you please have a look? I confirmed that this PR is working.
Thanks in advance.

libs/aws/langchain_aws/retrievers/bedrock.py
```diff
@@ -128,6 +170,7 @@ def _get_relevant_documents(
         response = self.client.retrieve(
             retrievalQuery={"text": query.strip()},
             knowledgeBaseId=self.knowledge_base_id,
+            nextToken=None,
```
3coins, line 173: Should this be handled to extract the next set of results, or parameterized from `kwargs` so that the users can send the token and get the next set of documents?

3coins (commented):
@ravediamond 
Thanks for making these changes. Added some suggestions to simplify the code and handle the nextToken.

libs/aws/langchain_aws/retrievers/bedrock.py
```diff
@@ -10,48 +10,90 @@
...
+        None, alias="notIn"
+    )
+    orAll: Optional[List["SearchFilter"]] = None
+    startsWith: Optional[
+        Dict[str, Union[Dict, List, int, float, str, bool, None]]
+    ] = None
+    stringContains: Optional[
+        Dict[str, Union[Dict, List, int, float, str, bool, None]]
+    ] = None
+
```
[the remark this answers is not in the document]
ravediamond: I will find a way to simplify and correct it.

libs/aws/langchain_aws/retrievers/bedrock.py
```diff
@@ -10,48 +10,90 @@
...
+    class Config:
+        allow_population_by_field_name = True
+
+
 class VectorSearchConfig(BaseModel, extra="allow"):  # type: ignore[call-arg]
     """Configuration for vector search."""
 
     numberOfResults: int = 4
+    filter: Optional[SearchFilter] = None
+    overrideSearchType: Optional[str] = None  # Can be 'HYBRID' or 'SEMANTIC'
```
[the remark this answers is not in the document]
ravediamond: Far better like this. I will integrate it.

libs/aws/langchain_aws/retrievers/bedrock.py
ravediamond: To implement the handling of `next_token`, I had to:

-  create another method called `_get_relevant_documents_with_token` that will return the list of `Documents` and the token as a tuple.
- This new method was created because the  method `_get_relevant_documents` from the class `BaseRetriever` from which this class inherits has a signature of `List[Document]`.
- Remove the `token` usage from `_get_relevant_documents` because of the duplication with the previous method.

ravediamond:
@3coins Thank you for the review. I correct my PR as per your recommendations. What do you think ?

libs/aws/langchain_aws/retrievers/bedrock.py
```diff
@@ -9,49 +9,82 @@
...
 from typing_extensions import Annotated
 
+FilterValue = Union[Dict[str, Any], List[Any], int, float, str, bool, None]
+
+
+class Criterion(BaseModel):
+    """A model to encapsulate a filter criterion with a key and its value."""
+
+    key: str
+    value: FilterValue
```
3coins, line 19: Criterion doesn't seem like the most intuitive or descriptive name for the filter. Can we stick with the Filter and FilterValue names. Rather than creating a model class, a type is sufficient here.

libs/aws/langchain_aws/retrievers/bedrock.py
```diff
@@ -123,11 +156,16 @@ def _filter_by_score_confidence(self, docs: List[Document]) -> List[Document]:
         return filtered_docs
 
     def _get_relevant_documents(
-        self, query: str, *, run_manager: CallbackManagerForRetrieverRun
+        self,
+        query: str,
+        *,
+        run_manager: CallbackManagerForRetrieverRun,
+        next_token: Optional[str] = None,
```
3coins, line 163: Let's not add this. Users can use the kwargs to pass this if needed.

```suggestion
```

libs/aws/langchain_aws/retrievers/bedrock.py
```diff
@@ -123,11 +156,16 @@ def _filter_by_score_confidence(self, docs: List[Document]) -> List[Document]:
...
     ) -> List[Document]:
         response = self.client.retrieve(
             retrievalQuery={"text": query.strip()},
             knowledgeBaseId=self.knowledge_base_id,
+            nextToken=next_token,
```
3coins, line 168: Assuming nextToken is required field, otherwise let's not add this.
```suggestion
            nextToken=None,
```

3coins (changes requested):

libs/aws/langchain_aws/retrievers/bedrock.py
ravediamond: Yes I removed it.

libs/aws/langchain_aws/retrievers/bedrock.py
ravediamond: I put this at None as I removed the argument nextToken.

libs/aws/langchain_aws/retrievers/bedrock.py
```diff
@@ -147,3 +185,36 @@ def _get_relevant_documents(
...
+                result["source_metadata"] = result.pop("metadata")
+            documents.append(
+                Document(
+                    page_content=content,
+                    metadata=result,
+                )
+            )
+
+        filtered_documents = self._filter_by_score_confidence(docs=documents)
+        return (filtered_documents, new_next_token)
```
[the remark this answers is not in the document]
ravediamond: Ok let's do that in another MR. I will put this as it was before.

libs/aws/langchain_aws/retrievers/bedrock.py
ravediamond: You are right, it is a little overkill. I simplified it with a simple `Filter`.

ravediamond:
Ok I corrected as per your reviews and I will create another PR for the nextToken when it is merged. 
Thanks in advance fir the review @3coins and also for having merged my other PR.

libs/aws/langchain_aws/retrievers/bedrock.py
```diff
@@ -123,11 +150,15 @@ def _filter_by_score_confidence(self, docs: List[Document]) -> List[Document]:
...
-        self, query: str, *, run_manager: CallbackManagerForRetrieverRun
+        self,
+        query: str,
+        *,
+        run_manager: CallbackManagerForRetrieverRun,
     ) -> List[Document]:
         response = self.client.retrieve(
             retrievalQuery={"text": query.strip()},
             knowledgeBaseId=self.knowledge_base_id,
+            nextToken=None,
```
3coins, line 161: This is redundant and not required. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-agent-runtime/client/retrieve.html

```suggestion
            nextToken=None,
```

ravediamond: You are right, I removed it.

3coins (approved):
@ravediamond 
Made a small update to exclude None values from filter, this was failing the integration test. 
Looks good now, thanks for making all the updates.

PR #115, opened 2024-07-17
Fixes ChatBedrock for llama3

### LINKED ISSUE #90: llama3-70b chat no longer functioning with Bedrock
With the most recent version it appears that ChatBedrock does not function when I call using a llama3 model using chain.ainvoke({}).

```python
\langchain_aws\llms\bedrock.py", line 867, in _aprepare_input_and_invoke_stream
    body = json.dumps(input_body)
                      ^^^^^^^^^^
UnboundLocalError: cannot access local variable 'input_body' where it is not associated with a value
```

It appears that the code does not set the input_body value if the model is not claude-3:
```python
if "claude-3" in self._get_model():
            if _tools_in_params(params):
                input_body = LLMInputOutputAdapter.prepare_input(
                    provider=provider,
                    model_kwargs=params,
                    prompt=prompt,
                    system=system,
                    messages=messages,
                    tools=params["tools"],
                )
            else:
                input_body = LLMInputOutputAdapter.prepare_input(
                    provider=provider,
                    prompt=prompt,
                    system=system,
                    messages=messages,
                    model_kwargs=params,
                )
body = json.dumps(input_body)
```
This is seemingly a result of the following commit, perhaps the else is at the wrong level?
https://github.com/langchain-ai/langchain-aws/commit/018ed897c79ed51008fef98ef4b2818e7306e044

3coins:
Fixes #90

### DIFF
--- a/libs/aws/langchain_aws/llms/bedrock.py
+++ b/libs/aws/langchain_aws/llms/bedrock.py
@@ -847,24 +847,23 @@ class BedrockBase(BaseLanguageModel, ABC):
             _model_kwargs["stream"] = True
 
         params = {**_model_kwargs, **kwargs}
-        if "claude-3" in self._get_model():
-            if _tools_in_params(params):
-                input_body = LLMInputOutputAdapter.prepare_input(
-                    provider=provider,
-                    model_kwargs=params,
-                    prompt=prompt,
-                    system=system,
-                    messages=messages,
-                    tools=params["tools"],
-                )
-            else:
-                input_body = LLMInputOutputAdapter.prepare_input(
-                    provider=provider,
-                    prompt=prompt,
-                    system=system,
-                    messages=messages,
-                    model_kwargs=params,
-                )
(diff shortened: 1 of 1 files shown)


ccurme (approved):

3coins:
@ccurme 
Thanks for approving this change. There seems to be a [related issue](https://github.com/langchain-ai/langchain/issues/24377) which is failing streaming use cases, would you be able to help with this? I can make the necessary changes, if this is an oversight in `langchain-aws`.

ccurme:
> @ccurme Thanks for approving this change. There seems to be a [related issue](https://github.com/langchain-ai/langchain/issues/24377) which is failing streaming use cases, would you be able to help with this? I can make the necessary changes, if this is an oversight in `langchain-aws`.

Believe this is a bug in langchain-aws that is fixed in the `ChatBedrockConverse` implementation
cc @baskaryan

PR #126, opened 2024-07-30
correctly raise ValueError for ChatBedrock models that don't support structured_output

andymli:
small addendum to an old PR: ttps://github.com/langchain-ai/langchain-aws/pull/70

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -387,7 +387,7 @@ class ChatBedrock(BaseChatModel, BedrockBase):
 
     system_prompt_with_tools: str = ""
     beta_use_converse_api: bool = False
-    """Use the new Bedrock ``converse`` API which provides a standardized interface to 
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -784,7 +784,7 @@ class AnswerWithJustification(BaseModel):
                 schema, include_raw=include_raw, **kwargs
             )
         if "claude-3" not in self._get_model():
-            ValueError(
+            raise ValueError(
                 f"Structured output is not supported for model {self._get_model()}"
             )
```
3coins, line 789: I would suggest to rather get rid of this block as there are other providers that support tool calling and structured output.

andymli: is there an up-to-date list of these providers given by AWS? the above branch condition was written by the original PR author. I think it's correct as from my manual testing, only `claude-3` models support structured output on Bedrock (for now).
I'm only providing a code correction so it correctly raises an error like it should have been doing, not any major breaking changes.

PR #133, opened 2024-08-02
de-beta converse

efriis:


3coins:
@efriis 
Should we also remove the `beta` qualifier from the ChatBedrock class and rename `beta_use_converse_api` to `use_converse_api`? Adding an alias for `beta_use_converse_api` should make this non-breaking for existing customers.

efriis:
> Should we also remove the beta qualifier from the ChatBedrock class

I think this doesn't have a `beta` decorator on it!

>and rename beta_use_converse_api to use_converse_api? Adding an alias for beta_use_converse_api should make this non-breaking for existing customers.

I think it's ok to keep the conversion logic as a beta feature, which we'd plan to remove if/when converse becomes the main entrypoint for the integration. I think this feature is most useful for people who want to see if their current integration works ok with converse!

PR #135, opened 2024-08-02
aws[patch]: merge messages in chat bedrock converse

### LINKED ISSUE #130: ChatBedrock with Converse enabled returns consecutive user messages
When leveraging `ChatBedrock` with `beta_use_converse_api` and Langgraph, I am getting the following error:

`ValidationException: An error occurred (ValidationException) when calling the Converse operation: A conversation must alternate between user and assistant roles. Make sure the conversation alternates between user and assistant roles and try again.`

The tool nodes are defined using `ToolNode` from `langgraph.prebuilt`. The setup is otherwise similar to the one at the bottom of this [example](https://github.com/aws-samples/amazon-bedrock-workshop/blob/main/06_OpenSource_examples/05_OpenSource_agents/03_langgraph_agents_of_agent.ipynb).

The setup runs without error with `beta_use_converse_api=False`. The only difference seems to be that consecutive user messages are merged.

baskaryan:
Fix #130 

Note this updates the behavior for all models, not just anthropic.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -32,6 +32,7 @@ from langchain_core.messages import (
     SystemMessage,
     ToolCall,
     ToolMessage,
+    merge_message_runs,
 )
 from langchain_core.messages.ai import AIMessageChunk, UsageMetadata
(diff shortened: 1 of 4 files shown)


baskaryan:
cc @3coins

PR #119, opened 2024-07-23
Fixes streaming for anthropic tool call.

3coins:
Fixes #80 
Replaces #81

PR #144, opened 2024-08-06
Fix: Call `on_llm_new_token` during streaming

hourliert:
# Description

This previous [PR](https://github.com/langchain-ai/langchain-aws/pull/119/files#diff-7b555324692cbefa47fc582ade684c4faa5d083023137e2839f37aff0401aa4eL832) removed the call to `run_manager.on_llm_new_token` and can break client relying on this callback.
This PR adds this back.

### DIFF
--- a/libs/aws/tests/integration_tests/chat_models/test_bedrock.py
+++ b/libs/aws/tests/integration_tests/chat_models/test_bedrock.py
@@ -68,16 +68,18 @@ def test_chat_bedrock_generate_with_token_usage(chat: ChatBedrock) -> None:
 @pytest.mark.scheduled
 def test_chat_bedrock_streaming() -> None:
     """Test that streaming correctly streams chunks."""
+    callback_handler = FakeCallbackHandler()
     chat = ChatBedrock(  # type: ignore[call-arg]
         model_id="anthropic.claude-v2"
     )
     message = HumanMessage(content="Hello")
-    stream = chat.stream([message])
+    stream = chat.stream([message], RunnableConfig(callbacks=[callback_handler]))
 
     full = next(stream)
     for chunk in stream:
         full += chunk  # type: ignore[assignment]
 
+    assert callback_handler.llm_streams > 0
     assert full.content
     assert full.response_metadata
     assert full.usage_metadata  # type: ignore[attr-defined]
@@ -309,7 +311,7 @@ def test_chat_bedrock_token_callbacks() -> None:
     """
     callback_handler = FakeCallbackHandlerWithTokenCounts()
     chat = ChatBedrock(  # type: ignore[call-arg]
-        model_id="anthropic.claude-v2", streaming=False, verbose=True
+        model_id="anthropic.claude-v2", streaming=True, verbose=True
     )
     message = HumanMessage(content="Hello")
     response = chat.invoke([message], RunnableConfig(callbacks=[callback_handler]))
--- a/libs/aws/langchain_aws/llms/bedrock.py
+++ b/libs/aws/langchain_aws/llms/bedrock.py
@@ -943,6 +943,21 @@ class BedrockBase(BaseLanguageModel, ABC):
         ):
             yield chunk
 
+            token = None
+
+            if isinstance(chunk, GenerationChunk):
+                token = chunk.text
+            elif isinstance(chunk, AIMessageChunk):
+                token = chunk.content
+
+            if run_manager is not None and asyncio.iscoroutinefunction(
+                run_manager.on_llm_new_token
+            ):
+                await run_manager.on_llm_new_token(token, chunk=chunk)
+            elif run_manager is not None:
+                run_manager.on_llm_new_token(token, chunk=chunk)  # type: ignore[unused-coroutine]
+
+
 
 class BedrockLLM(LLM, BedrockBase):
     """Bedrock models.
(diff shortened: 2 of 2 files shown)


libs/aws/tests/integration_tests/chat_models/test_bedrock.py
```diff
@@ -309,7 +311,7 @@ def test_chat_bedrock_token_callbacks() -> None:
     """
     callback_handler = FakeCallbackHandlerWithTokenCounts()
     chat = ChatBedrock(  # type: ignore[call-arg]
-        model_id="anthropic.claude-v2", streaming=False, verbose=True
```
hourliert, line 312: This test cases is for streaming (as per description). The `streaming=False` was toggled on this [PR](https://github.com/langchain-ai/langchain-aws/pull/119/files#diff-6d14d4a53f73d46e8fd45ed545e2a040c430b34bb79bc02c0843e0e33d3f2c6bR312), toggling back on.

pratik60:
@3coins Can you please review?

3coins:
@hourliert 
BaseChatModel already emits the `on_llm_new_token`. Adding these in the Bedrock LLMs will duplicate the calls to the callbacks.
https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/language_models/chat_models.py#L357

For instance, this test will pass without any new changes you are introducing in this PR.
```python
def test_chat_bedrock_streaming() -> None:
    """Test that streaming correctly streams chunks."""
    callback_handler = FakeCallbackHandler()
    chat = ChatBedrock(  # type: ignore[call-arg]
        model_id="anthropic.claude-v2"
    )
    message = HumanMessage(content="Hello")
    stream = chat.stream([message], RunnableConfig(callbacks=[callback_handler]))

    full = next(stream)
    for chunk in stream:
        full += chunk  # type: ignore[assignment]

    assert callback_handler.llm_streams > 0
    assert full.content
    assert full.response_metadata
    assert full.usage_metadata  # type: ignore[attr-defined]
```

Would help if you can include a sample code that can help reproduce the issue you are facing.

hourliert:
@3coins 👋 

We've tested with `langchain-aws==0.1.13` and `langchain-aws==0.1.15` and in the first version, the callback is called, not in the second. This is only impacting our code that is still using `LLMChain.run` (which is the majority of our codebase to this date).

Shared between both examples below:
```python
class LLMDatadogMetrics(BaseModel, BaseCallbackHandler):
    new_tokens = 0

    def on_llm_new_token(self, token: str, run_id: UUID, **kwargs: Any) -> Any:
        self.new_tokens += 1


@pytest.fixture
def sonnet_llm():
    llm = BedrockChat(
        model_id="anthropic.claude-3-5-sonnet-20240620-v1:0", streaming=True, callbacks=[LLMDatadogMetrics()]
    )
```

## 0.1.15

This **fails**, the majority of our code is like this:
```python
class TestBedrockChat:
    def test_performance_metrics_are_reported_on_success(self, sonnet_llm):
        prompt = PromptTemplate(template="Greet {name}", input_variables=["name"])
        chain = LLMChain(prompt=prompt, llm=sonnet_llm)

        chain.run(name="Jane")

        assert sonnet_llm.callbacks[0].new_tokens > 0
```

This **passes**: (ie. moving away from `LLMChain.run` and using a LCEL chain with `.stream()`):
```python
class TestBedrockChat:
    def test_performance_metrics_are_reported_on_success(self, sonnet_llm):
        prompt = PromptTemplate(template="Greet {name}", input_variables=["name"])
        chain = prompt | sonnet_llm

        for _ in chain.stream({"name": "Jane"}):
            pass

        assert sonnet_llm.callbacks[0].new_tokens > 0
```

## 0.1.13

**Both examples** pass.

hourliert:
FWIW, I tried to trace the code deep down into `langchain-core`, and when running through `LLMChain.run` we go through in order:

```python
langchain.chains.base.run
langchain.chains.base.invoke
langchain.chains.llm._call
langchain.chains.llm.generate
langchain_core.language_models.chat_models.generate_prompt
langchain_core.language_models.chat_models.generate
langchain_core.language_models.chat_models._generate_with_cache
langchain_aws.chat_models.bedrock._generate
```

The problem happens in `langchain_core.language_models.chat_models._generate_with_cache`:

When using `LLMChain.run(...)`, this [condition](https://github.com/langchain-ai/langchain/blob/4d28c70000e3845a6f162f322d7fafeb57713e38/libs/core/langchain_core/language_models/chat_models.py#L763) returns **false**, which prevents the `on_llm_new_token` callback to be [called](https://github.com/langchain-ai/langchain/blob/4d28c70000e3845a6f162f322d7fafeb57713e38/libs/core/langchain_core/language_models/chat_models.py#L784).

3coins:
@hourliert 
As per LangChain docs, `on_llm_new_token` should only be available when streaming is enabled. 
https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/callbacks/base.py#L67

hourliert:
@3coins
Streaming is enabled when I instantiate the llm:

```python
    llm = ChatBedrock(
        model_id="anthropic.claude-3-5-sonnet-20240620-v1:0", 
        streaming=True, # <-- HERE
        callbacks=[LLMDatadogMetrics()],
        # credentials...
    )
```

We take advantage of streaming in many places in our application, and I can guarantee that prior to `0.1.14`, it worked with `LLMChain.run` (with a `on_llm_new_token` callback passes to the `llm` that handles streaming each token).

----

I am not sure where this is going, we have a mitigation on our side now. So happy to close this PR. The interesting learning from me is that removing the `on_llm_new_token` from `langchain_aws` was a breaking change for us (it broke streaming for us) and it took us a while to understand why.

PR #146, opened 2024-08-07
Add ToolMessage's status to bedrock message for ChatBedrockConverse

thiagotps:
With this pull request, the `status` field of `ToolMessage` is now being passed to the `converse` api.


baskaryan:
we should double check that this doesn't cause any issues for non-anthropic models

3coins (approved):
@baskaryan 
Although, the status field seems to be only supported for Anthropic models as per Bedrock documentation, both success and error seems to work in my tests for Meta, Mistral and Cohere; these are the only other models that support tool calling currently.
https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolResultBlock.html

PR #26, opened 2024-04-25
fix guardrail attributes when invoking bedrock

brnaba-aws:
Fixed bedrock attributes when using guardrails. See example of usage:

```
from langchain_core.messages import HumanMessage
from langchain_aws.chat_models import ChatBedrock



llm_model_id='anthropic.claude-3-haiku-20240307-v1:0'

chat = ChatBedrock(
    model_id=llm_model_id,
    streaming=True,
    model_kwargs={"temperature": 0.1},
    guardrails={"guardrailIdentifier": "identifier", "guardrailVersion": "1", "trace": True},
)

messages = [
    HumanMessage(
        content="What are the financial advice you can give me?"
    )
]
for chunk in chat.stream(messages):
    print(chunk.content,  end="", flush=True)
```

PR #156, opened 2024-08-14
Fix ChatBedrock docs for guardrails

roberttstephens:
The guardrails attributes were changed in https://github.com/langchain-ai/langchain-aws/pull/26 . However, the docstring had the old parameter keys. It looks like this causes the example on https://api.python.langchain.com/en/latest/chat_models/langchain_aws.chat_models.bedrock.ChatBedrock.html#langchain_aws.chat_models.bedrock.ChatBedrock.guardrails to be wrong.

image


3coins (approved):
LGTM! Thanks for submitting the fix.

PR #168, opened 2024-08-26
Free boto3 dependancy from `<1.35.0` to `<2.0`

Will282:
Remove restrictive boto3 version requirement from `<1.35.0` to `<2.0`.

Boto3 follows semantic versioning ([source](https://docs.aws.amazon.com/sdkref/latest/guide/maint-policy.html#maint-versioning)) so there should be no changes to existing public interfaces with minor releases.

### DIFF
--- a/libs/aws/poetry.lock
+++ b/libs/aws/poetry.lock
@@ -2,98 +2,113 @@
 
 [[package]]
 name = "aiohappyeyeballs"
-version = "2.3.6"
+version = "2.4.0"
 description = "Happy Eyeballs for asyncio"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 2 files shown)


Maxi91f:
AFAIK they added a upper bound to [boto3 in release 0.1.2](https://github.com/langchain-ai/langchain-aws/releases/tag/v0.1.2) as this is a dependency bound defined on sagemaker-distribution ( #27 )

It seems that on the latest stable version of sagemaker-distribution they still [use a strict bound](https://github.com/aws/sagemaker-distribution/blob/517eea2cb8b0ae7dcd5b02f55d36e2c228150566/build_artifacts/v1/v1.9/v1.9.1/cpu.env.in#L30).

But I can't see if sagemaker-distribution is still related to this project.

I'm posting here to follow this discussion. I have also dependency version conflicts with boto3 on our project.

3coins:
Fixed by #201

PR #197, opened 2024-09-16
feat(langchain-aws): Add Support to pass in AWS credentials to bedrock models directly

langchain-infra:


baskaryan:
main thing to decide:
- are we ok adding env vars that are automatically read from
- do we want to call these AWS_ACCESS_...or BEDROCK_AWS_ACCESS_...

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -307,8 +308,49 @@ class Joke(BaseModel):
...
+    )
+    """AWS access key id. 
+    
+    If provided, aws_secret_access_key must also be provided.
+    If not specified, the default credential profile or, if on an EC2 instance,
+    credentials from IMDS will be used.
+    See: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html
+    
+    Will be read from env var 'AWS_ACCESS_KEY_ID' if it is set and env var 
+    aws_access_key_id init arg is not passed in.
```
3coins, line 326: Can we simplify these doc statements. Maybe something like:

If not provided, will be read from 'AWS_ACCESS_KEY_ID' environment variable.

3coins (approved):
Minor comment about the docs, looks good otherwise.

3coins:
> main thing to decide:
> 
> * are we ok adding env vars that are automatically read from
    Yes, we might not have to do this explicitly, as boto3 does this on it's own. See here: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables

> * do we want to call these AWS_ACCESS_...or BEDROCK_AWS_ACCESS_...
    Let's, keep the AWS_* naming, as it is an already defined convention and understood by users.

PR #201, opened 2024-09-18
Removing version ceiling for boto3.

3coins:
Removes version ceiling currently set at v1.35 for boto3.


ravediamond:
Thank you for this MR, I so much needed it for the bedrock cross-region inference endpoints.

Issue #202, opened 2024-09-19
Kendra queries are failing with langchain update


supreetkt:
After a recent upgrade to langchain-aws (from `v0.1.12` to `v0.2.0`), I'm noticing my kendra retrieval fail due to pydantic errors:

```
Validation errors for RetrieveResult\nResultItems.0.DocumentAttributes.0.Value.DateValue
Field required [type=missing, input_value={'StringValue': 'http://d...NotificationHowTo.html'}, input_type=dict]
For further information visit https://errors.pydantic.dev/2.9/v/missing
ResultItems.0.DocumentAttributes.0.Value.LongValue\n  Field required [type=missing, input_value={'StringValue': 'http://d...NotificationHowTo.html'}, input_type=dict]
For further information visit https://errors.pydantic.dev/2.9/v/missing\nResultItems.0.DocumentAttributes.0.Value.StringListValue\n  Field required [type=missing, input_value={'StringValue': 'http://d...NotificationHowTo.html'}, input_type=dict]
For further information visit https://errors.pydantic.dev/2.9/v/missing\nResultItems.1.DocumentAttributes.0.Value.DateValue\n  Field required [type=missing, input_value={'StringValue': 'http://d...NotificationHowTo.html'}, input_type=dict]\n    For further information visit https://errors.pydantic.dev/2.9/v/missing
```

Essentially, the retrieved results from Kendra are not fitting the pydantic schema set out for `DocumentAttributeValue`:
```
class DocumentAttributeValue(BaseModel, extra="allow"):  # type: ignore[call-arg]
    """Value of a document attribute."""

    DateValue: Optional[str]
    """The date expressed as an ISO 8601 string."""
    LongValue: Optional[int]
    """The long value."""
    StringListValue: Optional[List[str]]
    """The string list value."""
    StringValue: Optional[str]
    """The string value."""
```
Setting `DateValue`, `LongValue`, `StringListValue` and `StringValue` "optional" values are no longer working optionally. A default value of None needs to be set up on these above like below which would fix these issues:

```
class DocumentAttributeValue(BaseModel, extra="allow"):  # type: ignore[call-arg]
    """Value of a document attribute."""

    DateValue: Optional[str] = None
    """The date expressed as an ISO 8601 string."""
    LongValue: Optional[int] = None
    """The long value."""
    StringListValue: Optional[List[str]] = None
    """The string list value."""
    StringValue: Optional[str] = None
    """The string value."""
```

[...]

PR #203, opened 2024-09-19
aws[patch]: set defaults on DocumentAttributeValue

ccurme:
Resolves https://github.com/langchain-ai/langchain-aws/issues/202


supreetkt:
Thanks for a quick PR. This looks good to me. @3coins

ccurme:
> Thanks for a quick PR. I'm also assuming that you should add it for other optional fields which is not done yet. For example, `ResultItem`, etc. @ccurme

Thanks, should be all set.

supreetkt:
Great! Looks good to me @3coins

PR #216, opened 2024-09-25
Move yield of metrics chunk after generation chunk

### LINKED ISSUE #215: Streaming responses for Mistral models are getting truncated
**Description**
When using any of the Mistral models in streaming mode, the last token will get truncated if a stop sequence is added.

**Reproducible Sample**

```python
from langchain_core.messages import HumanMessage, SystemMessage

messages = [
    SystemMessage(
        content="Respond in no more than 5-8 words."
    ),
    HumanMessage(
        content="Hello, how are you?"
    )
]

llm = ChatBedrock(
    model_id="mistral.mistral-small-2402-v1:0",
    streaming=True,
    model_kwargs={
        "stop": ["test"] #needed to trigger the bug; this causes Bedrock service to add last token alongside the stop_reason
    },
)

#print streamed chunks
for chunk in llm.stream(input=messages):
    print(chunk.content, end="", flush=True)
```
  
**Example Truncated Response (Input = _"Hello, how are you?"_)**
_I'm an AI, I don't have feelin_

ihmaws:
## Move yield of metrics chunk after generation chunk
  - when using mistral and streaming is enabled,the final chunk includes a stop_reason. There is nothing to say this final chunk doesn't also include some generated text. The existing implementation would result in that final chunk never getting sent back
  - this update moves the yield of the metrics chunk after the generation chunk
  - also included a change to include invocation metrics for cohere models

Closes #215

### DIFF
--- a/libs/aws/langchain_aws/llms/bedrock.py
+++ b/libs/aws/langchain_aws/llms/bedrock.py
@@ -369,7 +369,17 @@ class LLMInputOutputAdapter:
             ):
                 return
 
-            elif (
+            generation_chunk = _stream_response_to_generation_chunk(
+                chunk_obj,
+                provider=provider,
+                output_key=output_key,
+                messages_api=messages_api,
+                coerce_content_to_string=coerce_content_to_string,
+            )
+            if generation_chunk:
+                yield generation_chunk
+
+            if (
                 provider == "mistral"
                 and chunk_obj.get(output_key, [{}])[0].get("stop_reason", "") == "stop"
             ):
@@ -383,18 +393,6 @@ class LLMInputOutputAdapter:
             elif messages_api and (chunk_obj.get("type") == "message_stop"):
                 yield _get_invocation_metrics_chunk(chunk_obj)
                 return
-
-            generation_chunk = _stream_response_to_generation_chunk(
-                chunk_obj,
-                provider=provider,
-                output_key=output_key,
-                messages_api=messages_api,
-                coerce_content_to_string=coerce_content_to_string,
-            )
-            if generation_chunk:
-                yield generation_chunk
-            else:
-                continue
 
     @classmethod
     async def aprepare_output_stream(
--- a/libs/aws/tests/unit_tests/llms/test_bedrock.py
+++ b/libs/aws/tests/unit_tests/llms/test_bedrock.py
@@ -270,6 +270,11 @@ MOCK_STREAMING_RESPONSE = [
     {"chunk": {"bytes": b'{"text": " you"}'}},
 ]
 
+MOCK_STREAMING_RESPONSE_MISTRAL = [
+    {"chunk": {"bytes": b'{"outputs": [{"text": "Thank","stop_reason": null}]}'}},
+    {"chunk": {"bytes": b'{"outputs": [{"text": "you.","stop_reason": "stop"}]}'}},
+]
+
 
 async def async_gen_mock_streaming_response() -> AsyncGenerator[Dict, None]:
     for item in MOCK_STREAMING_RESPONSE:
@@ -328,6 +333,12 @@ def mistral_response():
         },
     )
 
+    return response
+
+
+@pytest.fixture
+def mistral_streaming_response():
+    response = dict(body=MOCK_STREAMING_RESPONSE_MISTRAL)
     return response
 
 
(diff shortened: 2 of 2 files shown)


libs/aws/langchain_aws/llms/bedrock.py
```diff
@@ -367,9 +367,20 @@ def prepare_output_stream(
             if provider == "cohere" and (
                 chunk_obj["is_finished"] or chunk_obj[output_key] == "<EOS_TOKEN>"
             ):
+                yield _get_invocation_metrics_chunk(chunk_obj)
```
ihmaws, line 370: This is an optional change, if you believe there is risk in adding it, happy to remove it. It is not the core focus of the review.

I couldn't find a reason why metrics weren't included for cohere.

SanjayTiwaryMLAI:
@3coins, Need urgent help here as there is issue with the production deployment GenAI Chatbot for Government e market place. (https://github.com/ihmaws)has done deep dive on the issue and suggested few resolution. Can you look and please accept the PR.
 
Regards
Sanjay Tiwary

libs/aws/langchain_aws/llms/bedrock.py
```diff
@@ -367,9 +367,20 @@ def prepare_output_stream(
...
-            elif (
+            generation_chunk = _stream_response_to_generation_chunk(
+                chunk_obj,
+                provider=provider,
+                output_key=output_key,
+                messages_api=messages_api,
+                coerce_content_to_string=coerce_content_to_string,
+            )
+            if generation_chunk:
+                yield generation_chunk
```
3coins, line 381: This ends up adding an additional chunk to the response with an empty text. For example:
```
# Last text chunk with content
content='.' additional_kwargs={} response_metadata={'stop_reason': None} id='run-98ddf143-b333-47de-b8e5-b0dd55832cf6'

# empty content chunk processed for case when metrics are present
content='' additional_kwargs={} response_metadata={'stop_reason': 'stop', 'amazon-bedrock-invocationMetrics': {'inputTokenCount': 33, 'outputTokenCount': 10, 'invocationLatency': 230, 'firstByteLatency': 139}} id='run-98ddf143-b333-47de-b8e5-b0dd55832cf6'

# This one is processed with the _get_invocation_metrics_chunk
content='' additional_kwargs={} response_metadata={} id='run-98ddf143-b333-47de-b8e5-b0dd55832cf6' usage_metadata={'input_tokens': 33, 'output_tokens': 10, 'total_tokens': 43}
``` 
Ideally, we should be able to extract both the content and metrics in one go inside the `_stream_response_to_generation_chunk` method, not sure why we ended up with 2 different statements. I am not sure if Bedrock or any of the models guarantee the metrics as a last chunk; if that ends up not the case, this logic will break because of the return statements. However, this change seems harmless, maybe we can tackle the other issue in a separate PR.

3coins (commented):
@ihmaws 
Thanks for submitting this change. One minor comment about removing the block for cohere, looks good otherwise.

libs/aws/langchain_aws/llms/bedrock.py
ihmaws: Done. Removed in latest push.

3coins:
@ihmaws 
One of integration tests is failing, can you verify if this passes for you, otherwise, update it.
```
======================================================= FAILURES ========================================================
__________________________________________________ test_bedrock_invoke __________________________________________________

chat = ChatBedrock(client=<botocore.client.BedrockRuntime object at 0x121fccd90>, region_name='us-west-2', model_id='anthropic.claude-3-sonnet-20240229-v1:0', model_kwargs={'temperature': 0})

    @pytest.mark.scheduled
    def test_bedrock_invoke(chat: ChatBedrock) -> None:
        """Test invoke tokens from ChatBedrock."""
        result = chat.invoke("I'm Pickle Rick", config=dict(tags=["foo"]))
        assert isinstance(result.content, str)
        assert "usage" in result.additional_kwargs
>       assert result.additional_kwargs["usage"]["prompt_tokens"] == 12
E       assert 13 == 12

tests/integration_tests/chat_models/test_bedrock.py:240: AssertionError
```

Use this to verify:
```
poetry run pytest tests/integration_tests/chat_models/test_bedrock.py -k "test_bedrock_invoke"
```

ihmaws:
> @ihmaws One of integration tests is failing, can you verify if this passes for you, otherwise, update it.
> 

Interesting, I'm not seeing that test failure. Let me try and have a look. Can you also run the test one more time and confirm you've got the latest code?

3coins:
> Interesting, I'm not seeing that test failure. Let me try and have a look. Can you also run the test one more time and confirm you've got the latest code?

Let me try pulling in latest.

3coins (approved):

PR #218, opened 2024-09-30
Semantic cache for InMemory Vector store 

LakshmiPeri23:
Added cache.py which provide Semantic Caching functionality to store question and answers in MemoryDB. Inlucded an example SemanticCache.ipynb jupyter notebook for test


libs/aws/langchain_aws/cache.py
```diff
@@ -0,0 +1,275 @@
...
+Cache directly competes with Memory. See documentation for Pros and Cons.
+
+**Class hierarchy:**
+
+.. code-block::
+
+    BaseCache --> <name>Cache  # Examples: I
+"""
+
+from __future__ import annotations
```
3coins, line 21: Would this code function without this statement?

libs/aws/langchain_aws/cache.py
```diff
@@ -0,0 +1,275 @@
...
+    # TODO - implement a TTL policy in MemoryDB
+
+    DEFAULT_SCHEMA = {
+        "content_key": "prompt",
+        "text": [
+            {"name": "prompt"},
+            {"name": "return_val"},
+            {"name": "llm_string"},
+        ],
+    }  # This bracket closes the dictionary correctly
```
3coins, line 159: Please remove arbitrary comments like these.

libs/aws/langchain_aws/cache.py
```diff
@@ -0,0 +1,275 @@
...
+
+    def __init__(
+        self, redis_url: str, embedding: Embeddings, score_threshold: float = 0.2
+    ):
+        """Initialize by passing in the `init` GPTCache func
```
3coins, line 164: Is this accurate?

samples/inmemory/Semanticcache.ipynb
```diff
@@ -0,0 +1,270 @@
...
+    "                embedding=embeddings)\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0ac943fd-6ffe-4c99-ac5c-e9fbf93b7e44",
+   "metadata": {},
+   "source": [
+    "#### As a question to the LLM and Re-run the same block to see the improvemnt in response time. "
```
3coins, line 195:
```suggestion
    "#### Submit a query to the LLM and re-run the same cell to see the improvement in response time. "
```

samples/inmemory/Semanticcache.ipynb
```diff
@@ -0,0 +1,270 @@
...
+    "rc.ping()\n",
+    "#rc.flushall()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "1f7fe013-f713-4797-ae38-fe7a0f3e32b3",
+   "metadata": {},
+   "source": [
+    "## Ask a question without a cache"
```
3coins, line 150:
```suggestion
    "## Submit a query without a cache"
```

samples/inmemory/Semanticcache.ipynb
```diff
@@ -0,0 +1,270 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "41f56fd0-efc1-4edb-9a99-d41de0954949",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "%env MEMORYDB_HOST=your_cluster_endpoint\n",
+    "%env MEMORYDB_PORT=6379"
```
3coins, line 21: Suggestion to remove setting env variables in the notebook as the main reason to use env variables are so that they are set by the host, and they are not visible directly in the notebook. We can add a note that these env variables are expected to run this notebook. 
An alternative would be to just use regular variables to store these values.

3coins (commented):
@LakshmiPeri23 
Thanks for submitting this change. Should this be part of `caches` module, rather than root?

LakshmiPeri23:
> ange. Should this be part of `caches` module,

I saw even tn elangchain code cache is in root and not moved to a seperate module so left the same format.

libs/aws/langchain_aws/cache.py
LakshmiPeri23: yes we are not implementing standard cache, only semantic caching.

LakshmiPeri23:
i will work on addressing remaining feedback and submit the changes

3coins (approved):

PR #226, opened 2024-10-03
aws[patch]: support ChatBedrock(max_tokens, temperature)

baskaryan:


efriis (approved):
does the standard params standard unit test still fail? Could we un-xfail it after this?

baskaryan:
> does the standard params standard unit test still fail? Could we un-xfail it after this?

which test

PR #241, opened 2024-10-15
Fixes token logging in callbacks when streaming=True is used.

### LINKED ISSUE #217: Setting streaming to True doesn't seems to make any difference
Hi,

If I try to run following code:

```python
import boto3
from typing import Any
from langchain.callbacks.base import BaseCallbackHandler


class CallbackHandler(BaseCallbackHandler):
    def on_llm_new_token(self, token: str, **kwargs: Any) -> Any:
        print(token, end="|")


session = boto3.Session(profile_name='sandbox')
bedrock_client = session.client(service_name='bedrock-runtime', region_name='us-east-1')
model_id = "anthropic.claude-3-5-sonnet-20240620-v1:0"
model = ChatBedrock(model_id=model_id, client=bedrock_client, streaming=True)
model.invoke("Hello, how are you?", config={"callbacks": [CallbackHandler()]})

# output:
# AIMessage(content="Hello! As an AI language model, I don't have feelings, but I'm functioning well and ready to assist you. How can I help you today?", additional_kwargs={'usage': {'prompt_tokens': 13, 'completion_tokens': 35, 'total_tokens': 48}, 'stop_reason': 'end_turn', 'model_id': 'anthropic.claude-3-5-sonnet-20240620-v1:0'}, response_metadata={'usage': {'prompt_tokens': 13, 'completion_tokens': 35, 'total_tokens': 48}, 'stop_reason': 'end_turn', 'model_id': 'anthropic.claude-3-5-sonnet-20240620-v1:0'}, id='run-0e73f036-39a4-402f-8e1a-438d1b9bd03b-0', usage_metadata={'input_tokens': 13, 'output_tokens': 35, 'total_tokens': 48})
```

The `on_llm_new_token` won't be called, even though I set `streaming` to True.

But when i run this:

```python
import boto3
from typing import Any
from langchain.callbacks.base import BaseCallbackHandler


class CallbackHandler(BaseCallbackHandler):
    def on_llm_new_token(self, token: str, **kwargs: Any) -> Any:
        print(token, end="|")


session = boto3.Session(profile_name='sandbox')
bedrock_client = session.client(service_name='bedrock-runtime', region_name='us-east-1')
model_id = "anthropic.claude-3-5-sonnet-20240620-v1:0"
model = ChatBedrock(model_id=model_id, client=bedrock_client, streaming=True)
for event in model.stream("Hello, how are you?", config={"callbacks": [CallbackHandler()]}):
    pass

# output: 
# |Hello|! As| an AI| language| model, I don|'t have| feelings|,| but I'm| functioning| well| an|d ready| to assist| you.| How can| I help| you today?|||
```

It works.

My question is why is that the case? I'm using langchain 3.0.0, before that, calling `invoke` with `streaming`equal to `True` worked. Now it looks like I need to call `stream` even if I set `streaming` parameter.

It's confusing. Can anyone explain why is that the case?

### LINKED ISSUE #240: Streaming with ChatBedrock: on_llm_new_token doesn't seem to be called
Since I've done a minor version upgrade from `v0.1.x` to `v0.2.y`, I'm having some trouble with streaming responses. Non-streaming works as expected. I'm not yet able to pinpoint the root cause, but as I'm trying to debug this, the sample code provided below isn't working as expected.

```
import boto3
from langchain.callbacks.base import BaseCallbackHandler
from langchain_core.prompts import ChatPromptTemplate

streaming = True
session = boto3.session.Session()
bedrock_client = session.client("bedrock-runtime", region_name="us-east-1")

class MyCustomHandler(BaseCallbackHandler):
    def on_llm_new_token(self, token: str, **kwargs) -> None:
        print(f"My custom handler, token: {token}")

prompt = ChatPromptTemplate.from_messages(["Tell me a joke about {animal}"])

model = ChatBedrock(
    client=bedrock_client, 
    model_id="anthropic.claude-3-haiku-20240307-v1:0",
    streaming = streaming, 
    callbacks=[MyCustomHandler()]
)

chain = prompt | model

response = chain.invoke({"animal": "bears"})
print(response)
```

For context, using the custom callback implementation provided [here](https://python.langchain.com/docs/how_to/custom_callbacks/), I'm trying out a sample code which should print out the `on_llm_new_token` invocations as the tokens are streamed. The response should be similar to the linked documentation:

```
My custom handler, token: Here
My custom handler, token: 's
My custom handler, token:  a
My custom handler, token:  bear
...
...


content="Here's a bear joke for you:\n\nWhy don't bears wear socks?\nBecause they have bear feet!\n\nHow's that? I tried to come up with a simple, silly joke about bears. Puns involving bears' physical characteristics are a common type of bear-related humor. Let me know if you'd like to hear another bear joke!" additional_kwargs={} response_metadata={'stop_reason': 'end_turn', 'stop_sequence': None} id='xxx' usage_metadata={'input_tokens': 13, 'output_tokens': 77, 'total_tokens': 90}
```

However, I seem to only get the final response in the following code and no intermediate `on_llm_new_token` invocations. I tried the code with `ChatAnthropic` as in the documentation and it works as expected/shown there. I also looked through ChatAnthropic and [SageMakerEndpoint](https://github.com/langchain-ai/langchain-aws/blob/adcfc348c235635e910b9b9bb5c661d3829b3955/libs/aws/langchain_aws/llms/sagemaker_endpoint.py#L356) implementations. They both invoke `on_llm_new_token` but `ChatBedrock` doesn't. 

I'm not sure if this is the complete reason why streaming has stopped working since the upgrade, but it seems like something that should be fixed.

3coins:
Fixes #240
Fixes #217 

### Code to verify
```python
from langchain.callbacks.base import BaseCallbackHandler
from langchain_core.prompts import ChatPromptTemplate

streaming = True

class MyCustomHandler(BaseCallbackHandler):
    def on_llm_new_token(self, token: str, **kwargs) -> None:
        print(f"My custom handler, token: {token}")

prompt = ChatPromptTemplate.from_messages(["Tell me a joke about {animal} in a few words."])

model = ChatBedrock(
    model_id="anthropic.claude-3-haiku-20240307-v1:0",
    streaming = streaming, 
    callbacks=[MyCustomHandler()]
)

chain = prompt | model

response = chain.invoke({"animal": "bears"})
```

### Output
```
My custom handler, token: 
My custom handler, token: Bear
My custom handler, token: -
My custom handler, token: ly funny
My custom handler, token: .
My custom handler, token: 
My custom handler, token: 
```

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -477,14 +477,19 @@ class ChatBedrock(BaseChatModel, BedrockBase):
             **kwargs,
         ):
             if isinstance(chunk, AIMessageChunk):
-                yield ChatGenerationChunk(message=chunk)
+                generation_chunk = ChatGenerationChunk(message=chunk)
+                if run_manager:
+                    run_manager.on_llm_new_token(
+                        generation_chunk.text, chunk=generation_chunk
+                    )
+                yield generation_chunk
             else:
                 delta = chunk.text
                 if generation_info := chunk.generation_info:
                     usage_metadata = generation_info.pop("usage_metadata", None)
                 else:
                     usage_metadata = None
-                yield ChatGenerationChunk(
+                generation_chunk = ChatGenerationChunk(
                     message=AIMessageChunk(
                         content=delta,
                         response_metadata=chunk.generation_info,
@@ -493,6 +498,11 @@ class ChatBedrock(BaseChatModel, BedrockBase):
                     if chunk.generation_info is not None
                     else AIMessageChunk(content=delta)
                 )
+                if run_manager:
+                    run_manager.on_llm_new_token(
+                        generation_chunk.text, chunk=generation_chunk
+                    )
+                yield generation_chunk
 
     def _generate(
         self,


3coins:
@wuodar @supreetkt
Can you folks verify. Thanks.

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -477,14 +477,19 @@ def _stream(
             **kwargs,
         ):
             if isinstance(chunk, AIMessageChunk):
-                yield ChatGenerationChunk(message=chunk)
+                generation_chunk = ChatGenerationChunk(message=chunk)
+                if run_manager:
```
supreetkt, line 481: Adding a check to test for empty/None values will fix the 2 empty responses in the end:
```
 if run_manager and generation_chunk.text:
```

supreetkt (changes requested):

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -493,6 +498,11 @@ def _stream(
                     if chunk.generation_info is not None
                     else AIMessageChunk(content=delta)
                 )
+                if run_manager:
+                    run_manager.on_llm_new_token(
```
supreetkt, line 502: And the same check should be added here as well.

supreetkt:
Btw, this check also seems to be missing from SagemakerEndpoint class. So there might be empty tokens there as well.

3coins:
@supreetkt 
Thanks for those suggestions. Upon checking some other partner implementations, there doesn't seem to be any filtering of messages in their code based on the content. There are valid cases in which other values in the callbacks might be useful, for example the response_metadata. 

I will look separately into the responses to see if we are yielding extra chunks that shouldn't be exposed. As far as this PR is concerned, I believe this fixes the raised issue. Let me know if you have any other suggestions.

3coins:
@supreetkt 

> I checked one partner implementation and they seemed to be checking only to ensure that the output is a string. Here is a sample empty generated chunk

Can you provide a link to that implementation. 

> I can just handle the same logic on my end within our custom implementation of the on_llm_new_token too :)

Yes, I think this can be handled easily by the callbacks based on their use case.

> I had one more concern: I tried using ChatBedrock with beta_use_converse_api set to True and it doesn't stream responses. I think its missing the usage of on_llm_new_token as well.

I will tackle this in a separate PR, but in general `streaming` attribute is legacy and not required at-all, only there for backward compatibility, since streaming is natively supported by LCEL and Runnables, and callbacks approach should be discouraged IMO.

supreetkt:
Here is the [link](https://github.com/langchain-ai/langchain/blob/a38e9033605ddada70f03b321f94c8e01202bc27/libs/partners/anthropic/langchain_anthropic/chat_models.py#L717)

supreetkt:
> @supreetkt
> 
> > I checked one partner implementation and they seemed to be checking only to ensure that the output is a string. Here is a sample empty generated chunk
> 
> Can you provide a link to that implementation.
> 
> > I can just handle the same logic on my end within our custom implementation of the on_llm_new_token too :)
> 
> Yes, I think this can be handled easily by the callbacks based on their use case.
> 
> > I had one more concern: I tried using ChatBedrock with beta_use_converse_api set to True and it doesn't stream responses. I think its missing the usage of on_llm_new_token as well.
> 
> I will tackle this in a separate PR, but in general `streaming` attribute is legacy and not required at-all, only there for backwards compatibility, since streaming is natively supported by LCEL and Runnables, and callbacks approach should be discouraged IMO.

Here is the [link](https://github.com/langchain-ai/langchain/blob/a38e9033605ddada70f03b321f94c8e01202bc27/libs/partners/anthropic/langchain_anthropic/chat_models.py#L717)

Also, if you're tackling the converse API issue in a separate PR, this LGTM.

PR #242, opened 2024-10-16
Fixes support for cross region inference

### LINKED ISSUE #239: ChatBedrockConverse#stream not streaming response for model ids with cross region inference when bind_tools is used
To reproduce, the following code will result in the same behavior as `invoke`; if comment out .bind_tools line, the response would be properly streamed.

```python

@tool(response_format="content_and_artifact")
def simple_calculator(a: int, b: int):
    """Use this tool to calcuate the sum of two integers.

    Args:
        a (int): The first integer.
        b (int): The second integer.

    Returns:
        int: The sum of the two integers.
    """
    return a + b

llm = ChatBedrockConverse(
    model="us.anthropic.claude-3-sonnet-20240229-v1:0",
    temperature=0,
    top_p=1,
    max_tokens=4096,
    region_name="us-west-2"
).bind_tools(tools=[simple_calculator])

a = llm.stream(
    input=[
        ("human", "Hello"),
    ],
)

full = next(a)

for x in a:
    print(x)
    full += x

print(full)
```

3coins:
Fixes #239 

1. Fixes streaming for cross-region inference models ids. 
2. Adds support for Cohere models in ChatBedrockConverse.
3. Also, adds standard integration tests for Cohere.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -400,15 +400,18 @@ class ChatBedrockConverse(BaseChatModel):
     @model_validator(mode="before")
     @classmethod
     def set_disable_streaming(cls, values: Dict) -> Any:
-        values["provider"] = (
-            values.get("provider")
-            or (values.get("model_id", values["model"])).split(".")[0]
+        model_id = values.get("model_id", values.get("model"))
+        model_parts = model_id.split(".")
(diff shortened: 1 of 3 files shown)


3coins:
@efriis 
Can you plz make sure CI account has access to the `cohere.command-r-plus-v1:0` model.

rsgrewal-aws:
@3coins why do we need the provider values and why do we need to parse the model id ? can we simply use the model id as is ?

PR #187, opened 2024-09-08
feat(community): Add support for Bedrock cross-region inference models

keremnalbant:
Merged for `LangChainJS`: https://github.com/langchain-ai/langchainjs/pull/6682
Closed PR as the `Bedrock` is deprecated in `langchain-core`: https://github.com/langchain-ai/langchain/pull/26038

Hey, I recently I switched using to Bedrock Inference models.

With Inference models, Amazon introduced a new prefix which is a region code. If the `modelId` previously was `anthropic.claude-3-5-sonnet-20240620-v1:0` now, with the change it's `eu.anthropic.claude-3-5-sonnet-20240620-v1:0`

This PR adds:
- Support for each AWS Bedrock supported regions.
- Support streaming for inference models in Converse API

Fixes #186

PR #246, opened 2024-10-20
adding support for cross region models

rsgrewal-aws:
This PR is rebased from the approved PR https://github.com/langchain-ai/langchain-aws/pull/187

this includes the fixes and is re-based to the main branch


rsgrewal-aws:
@3coins  could you please review this PR -- this PR has been taken from https://github.com/langchain-ai/langchain-aws/pull/187 which has been approved and rebased to the main branch. we have a few customers asking for this

rsgrewal-aws:
@3coins  why do we need the provider values and why do we need to parse the model id ? can we simply use the model id as is ?

3coins:
@rsgrewal-aws 

> for converse api we have to assume streaming is supported

As per the documentation, only Anthropic and Cohere support streaming with tools at the moment.

> one of the keys in this can be provider, second can be streaming

What do you mean by this?

PR #254, opened 2024-10-24
Fix bug with Agents for the enable_trace events

jdbaker01:
The trace_log was either always empty or only showed the last one because the events come through in succession and the last one was being returned only. Changed the trace log variable to a list and append new events then at the end of of the loop serialize the array to a string. 

Added new unit tests for our different output types and to check the existence of the trace log. 

Fixes bug introduced in #244.


rsgrewal-aws:
@3coins  could you please review this urgent PR

3coins:
@jdbaker01 
Thanks for this fix. Code looks good, a few minor things. 
1. Can you fix the lint error. Running `make format` and `make lint` should fix them.
    ```shell
    tests/unit_tests/agents/test_bedrock_agents.py:49: error: X | Y syntax for unions requires Python 3.10  [syntax]
    tests/unit_tests/agents/test_bedrock_agents.py:57: error: X | Y syntax for unions requires Python 3.10  [syntax]
    ```
2. langchain-aws and langchain repos use pytest as the test runner, there is extra setup `main` and `run_tests` that are redundant if you convert these tests to pytest format with just functions. Use fixtures instead of the setup function. Let me know if you need help with this.

3coins (approved):
LGTM!

PR #274, opened 2024-11-06
Added integration tests for meta

3coins:
Added integration tests for `ChatBedrockConverse` for Meta provider. 

2 tests are primarily failing here. 
1. `test_tool_calling`, which is consistently returning a quoted string input value instead of an expected integer e.g., {input: '3'} instead of {input: 3}, thus failing the test. This needs further investigation to determine whether this is a bug with the Bedrock service or an expected behavior with Llama models when a numeric input is required. 
2. `test_tool_message_histories_list_content`, which is failing for several other providers as well, as Bedrock doesn't seem to allow conversation block and tool use blocks in the same turn.
        ```python
        botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the Converse operation: messages.1.content: Conversation blocks and tool use blocks cannot be provided in the same turn.
        ```

### DIFF
--- a/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
@@ -70,6 +70,28 @@ class TestBedrockCohereStandard(ChatModelIntegrationTests):
         pass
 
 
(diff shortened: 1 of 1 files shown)


3coins:
cc @cab938 @Forma-Lin

PR #278, opened 2024-11-15
fix(bedrock):  Correct Cross-Region Inference Identifiers

watany-dev:
Inference endpoints in the Asia-Pacific region now correctly identify model IDs starting with the `apac` prefix. This update ensures that region-specific identifiers are parsed and processed accurately for cross-region inference.

ex.
https://dev.classmethod.jp/articles/amazon-bedrock-cross-region-inference-apac/


paikwiki:
I also encountered this issue. When using the “apac” region prefix in its current state, the following error occurs:

```sh
NotImplementedError: Provider apac model does not support chat.
env/lib/python3.11/site-packages/langchain_aws/chat_models/bedrock.py:362: NotImplementedError
```

The cause of this error is that the “apac” region prefix is not recognized and is mistakenly interpreted as a provider name. This seems to be an issue that needs to be fixed quickly.

Somers1:
A workaround until this gets merged is to specify the provider explicitly 
e.g.
`ChatBedrock(model_id='apac.anthropic.claude-3-5-sonnet-20240620-v1:0', provider='anthropic')`

michaelnchin (approved):
Thanks @watany-dev !

PR #282, opened 2024-11-19
aws[patch]: Remove numpy dependency (previously patch to support numpy2)

### LINKED ISSUE #147: Support numpy v2
langchain-aws currently pinned to `^1`

ghinch:
Increasing version of numpy to include 2.x.x for the langchain-aws package. Fixes #147


3coins:
@ghinch 
Thanks for submitting this change. langchain-community and langchain are both still using < 2 for numpy. https://github.com/langchain-ai/langchain/blob/master/libs/community/pyproject.toml#L51
https://github.com/langchain-ai/langchain/blob/master/libs/langchain/pyproject.toml#L49

Are there any implications of allowing v2? What would be the default numpy version installed with this change?

ghinch:
Thanks for following up @3coins 

This PR follows on from the update to `langchain-core`which was [released as part of 0.3.19 ](https://github.com/langchain-ai/langchain/releases/tag/langchain-core%3D%3D0.3.19). I've followed the pattern there which was to simply set the version of numpy to "less than 3". So the support would be there for any 2.x.x version. I believe the default behavior in `langchain-core` is for it to install the _lowest_ supported version (1.26.2 for core).

I have also opened similar PRs for `langchain` https://github.com/langchain-ai/langchain/pull/28183 and `langchain-community` https://github.com/langchain-ai/langchain/pull/28184.

I am not aware of any changes in the 2.x.x versions of `numpy` which are breaking for any of the Langchain modules (and the CI seems to be passing for all PRs I opened), but I'm happy if you would like to wait for the above PRs to be merged first.

3coins:
@ghinch 
Since, numpy is already present as dependency in langchain packages, we can remove it as a direct dependency in langchain-aws; I don't think it is directly used anywhere within this package. Run `poetry lock` to update the poetry lock file, after you remove it in pyproject.toml. I have tested this change, and I don't see any failures. Thanks again for raising this PR.

ghinch:
Thanks @3coins, makes sense. 
I've run `poetry remove numpy` and committed this change

3coins (approved):
LGTM!

Issue #234, opened 2024-10-10
Fix ChatBedrock to enable "test_stop_sequence" integration test


3coins:
Add `stop` attribute to ChatBedrock and ChatBedrockConverse classes to fix the standard "test_stop_sequence" integration tests.

```shell
tests/integration_tests/chat_models/test_standard.py::TestBedrockStandard::test_stop_sequence FAILED
tests/integration_tests/chat_models/test_standard.py::TestBedrockUseConverseStandard::test_stop_sequence FAILED

```

Forma-Lin:
Hello @3coins, is anyone working on this issue right now? If no, may I take it?

3coins:
@Forma-Lin 
Plz feel free to work on this. Let me know if I can help with anything to get you started. Assigned the issue to you.

Forma-Lin:
![screenshot](https://github.com/user-attachments/assets/be7e03c1-dd18-4a47-8d07-52e714efaa24)
@3coins 
Hi, I setted up the project following the CONTRIBUTING.md, and run integration test. But the results seems wired. I got lots of tests failed and I cannot find the standard "test_stop_sequence" integration tests in the output. Is there something else I should set up before running integration tests?

Forma-Lin:
@3coins 
My teammate have submitted a PR, and want to kindly invite you to review it.
The PR address is: https://github.com/langchain-ai/langchain-aws/pull/290

michaelnchin:
`stop_sequences` were added to ChatBedrock in #290 - resolving.

PR #290, opened 2024-11-28
Added stop sequence parameter to ChatBedrock model

zhaoJoseph:
This is a fix for https://github.com/langchain-ai/langchain-aws/issues/234. Test still passes while noting it is not implemented.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -36,7 +36,7 @@ from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResu
 from langchain_core.runnables import Runnable, RunnableMap, RunnablePassthrough
 from langchain_core.tools import BaseTool
 from langchain_core.utils.pydantic import TypeBaseModel, is_basemodel_subclass
-from pydantic import BaseModel, ConfigDict
+from pydantic import BaseModel, ConfigDict, Field
 
 from langchain_aws.chat_models.bedrock_converse import ChatBedrockConverse
 from langchain_aws.function_calling import (
@@ -392,6 +392,12 @@ class ChatBedrock(BaseChatModel, BedrockBase):
     """Use the new Bedrock ``converse`` API which provides a standardized interface to 
     all Bedrock models. Support still in beta. See ChatBedrockConverse docs for more."""
 
+    stop_sequences: Optional[List[str]] = Field(default=None, alias='stop')
+    """Stop sequence inference parameter from new Bedrock ``converse`` API providing 
+    a sequence of characters that causes a model to stop generating a response. See 
+    https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_InferenceConfiguration.html 
+    for more."""
+    
     @property
     def _llm_type(self) -> str:
         """Return type of chat model."""


michaelnchin (changes requested):
@zhaoJoseph, thank you for taking a look at this.

We need to also be providing the value of `stop_sequences` to the model to make the parameter functional.

Can you add additional code to ensure that the value of `stop_sequences` is passed with `inferenceConfig` in the [body of the invoke_model request](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html)?

The implementation in `ChatBedrockConverse` class can be used as a reference:
https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L632

zhaoJoseph:
Hi @michaelnchin, thanks for reviewing my PR. I have taken a look at ``ChatBedrockConverse`` class and added an implementation for ``stop_sequences`` in ``ChatBedrock``. Also modified the test case for  ``stop_sequences`` to reflect this change.

libs/aws/tests/integration_tests/chat_models/test_standard.py
```diff
@@ -44,6 +39,7 @@ def standard_chat_model_params(self) -> dict:
         return {
             "temperature": 0,
             "max_tokens": 100,
+            "stop_sequences": [],
```
michaelnchin, line 42: We shouldn't include `stop_sequences` here, as it's not a standard ChatModel parameter: https://python.langchain.com/docs/concepts/chat_models/#standard-parameters

libs/aws/tests/integration_tests/chat_models/test_standard.py
```diff
@@ -22,10 +21,6 @@ def chat_model_params(self) -> dict:
     def standard_chat_model_params(self) -> dict:
         return {"temperature": 0, "max_tokens": 100}
 
-    @pytest.mark.xfail(reason="Not implemented.")
-    def test_stop_sequence(self, model: BaseChatModel) -> None:
-        super().test_stop_sequence(model)
```
michaelnchin, line 27: The `test_stop_sequence` integ tests should be kept, I verified that they're now working with your fix. Just replace the `@pytest.mark.xfail(reason="Not implemented.")` decorator with ` @pytest.mark.scheduled`

libs/aws/tests/integration_tests/chat_models/test_standard.py
```diff
@@ -52,7 +48,3 @@ def standard_chat_model_params(self) -> dict:
     @property
     def supports_image_inputs(self) -> bool:
         return True
-
-    @pytest.mark.xfail(reason="Not implemented.")
-    def test_stop_sequence(self, model: BaseChatModel) -> None:
-        super().test_stop_sequence(model)
```
michaelnchin, line 58: Same as above, keep the test and replace the `xfail` decorator.

michaelnchin (changes requested):
Implementation LGTM! Few comments on the tests.

zhaoJoseph:
Thank you again for the quick response! I was considering removing the test case from `test_standard.py` because when running the integration tests I got an overrided failure `FAILED tests/integration_tests/chat_models/test_standard.py::TestBedrockStandard::test_no_overrides_DO_NOT_OVERRIDE - AssertionError: Standard tests overridden without @pytest.mark.xfail(reason="..."): ['test_stop_sequence']
FAILED tests/integration_tests/chat_models/test_standard.py::TestBedrockUseConverseStandard::test_no_overrides_DO_NOT_OVERRIDE - AssertionError: Standard tests overridden without @pytest.mark.xfail(reason="..."): ['test_stop_sequence']`
Should we ignore this result?

michaelnchin:
That makes sense @zhaoJoseph - revised my suggestions, let's stick with removing the `xfail` tests.

3coins (approved):
LGTM! Thanks for submitting this change.

PR #300, opened 2024-12-05
Fix model_id validation in ChatBedrock

michaelnchin:
Fixed a minor regression in v0.2.9. Runs of `ChatBedrock` with missing or NoneType `model_id` get caught in the new `beta_use_converse_api` parameter check, which throws an ambiguous TypeError. 

This change restores the original behavior where ValidationError is thrown for absent `model_id`.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -410,7 +410,7 @@ class ChatBedrock(BaseChatModel, BedrockBase):
     @model_validator(mode="before")
     @classmethod
     def set_beta_use_converse_api(cls, values: Dict) -> Any:
-        model_id = values.get("model_id", values.get("model"))
+        model_id = values.get("model_id", values.get("model")) or ""
 
         if "beta_use_converse_api" not in values:
             values["beta_use_converse_api"] = "nova" in model_id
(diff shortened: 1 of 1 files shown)


libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -410,7 +410,7 @@ def get_lc_namespace(cls) -> List[str]:
     @model_validator(mode="before")
     @classmethod
     def set_beta_use_converse_api(cls, values: Dict) -> Any:
-        model_id = values.get("model_id", values.get("model"))
+        model_id = values.get("model_id", values.get("model")) or ""
```
3coins, line 413: The error that you see is due to a null model_id on line 416. Rather than modifying the input, can we just add a check for model_id on line 415?

michaelnchin: Good point - updated.

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -412,7 +412,7 @@ def get_lc_namespace(cls) -> List[str]:
     def set_beta_use_converse_api(cls, values: Dict) -> Any:
         model_id = values.get("model_id", values.get("model"))
 
-        if "beta_use_converse_api" not in values:
+        if model_id is not None and "beta_use_converse_api" not in values:
```
3coins, line 415:
```suggestion
        if model_id and "beta_use_converse_api" not in values:
```

michaelnchin: This was my initial consideration, but I noticed that ChatBedrock treats empty strings as valid input for `model_id`, instead of throwing an error like it does for `None`. Likely no difference in the end result, but this check was made the same for consistency.

PR #306, opened 2024-12-17
Add Neptune chains

michaelnchin:
Initial port of [`NeptuneOpenCypherQAChain`](https://python.langchain.com/api_reference/community/chains/langchain_community.chains.graph_qa.neptune_cypher.NeptuneOpenCypherQAChain.html#langchain_community.chains.graph_qa.neptune_cypher.NeptuneOpenCypherQAChain) and [`NeptuneSparqlQAChain`](https://python.langchain.com/api_reference/community/chains/langchain_community.chains.graph_qa.neptune_sparql.NeptuneSparqlQAChain.html) out of the `langchain-community` package. 

This also adds the base `langchain` package as a new dependency.

To import, run:
```
from langchain_aws.chains import (
    NeptuneOpenCypherQAChain
    NeptuneSparqlQAChain
)
```


3coins:
@efriis 
Is there any concern on adding `langchain` as a direct dependency here?

3coins (changes requested):
@michaelnchin 
Thanks for submitting this change. Adding a direct dependency on `langchain` will not only increase the size of the package, but also increase the dependency footprint of `langchain-aws`. We should pivot to updating the code to start using LCEL chains, so there is no direct dependency on `langchain`. Here is an example where LangChain moved their StuffDocumentsChain to a factory method that returns an LCEL chain, we should follow the same strategy.
https://github.com/langchain-ai/langchain/blob/1378ddfa5fa901179a26e0e2a91cbea52de625cf/libs/langchain/langchain/chains/combine_documents/stuff.py#L24

michaelnchin:
Closing in favor of #322.

PR #311, opened 2024-12-19
Improve performance of NeptuneAnalyticsGraph

michaelnchin:
Improved the graph schema retrieval performance of `NeptuneAnalyticsGraph` by using the [pg_schema](https://docs.aws.amazon.com/neptune-analytics/latest/userguide/custom-algorithms-property-graph-schema.html) algorithm.

Before:
Screenshot 2024-12-18 at 8 59 02 PM
After:
Screenshot 2024-12-18 at 8 59 33 PM

### DIFF
--- a/libs/aws/langchain_aws/graphs/neptune_graph.py
+++ b/libs/aws/langchain_aws/graphs/neptune_graph.py
@@ -1,9 +1,6 @@
 import json
-import logging
 from abc import ABC, abstractmethod
 from typing import Any, Dict, List, Optional, Tuple, Union
-
-logger = logging.getLogger(__name__)
 
 
 def _format_triples(triples: List[dict]) -> List[str]:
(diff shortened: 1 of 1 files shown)


3coins (commented):
@michaelnchin 
Thanks for this performance improvement 🚀. Code looks good, but would be great if one of @krlawrence or @bechbd can review. 

One small question, would users need an option within [NeptuneOpenCypherQAChain](https://python.langchain.com/api_reference/community/chains/langchain_community.chains.graph_qa.neptune_cypher.NeptuneOpenCypherQAChain.html#langchain_community.chains.graph_qa.neptune_cypher.NeptuneOpenCypherQAChain) and [NeptuneSparqlQAChain](https://python.langchain.com/api_reference/community/chains/langchain_community.chains.graph_qa.neptune_sparql.NeptuneSparqlQAChain.html) to allow them to switch to a False option in any scenario?

michaelnchin:
Yes, I think users should be able to control this from the chain level also - will add separate options there.

bechbd:
Along the same note, is there any reason to keep around the old method for an NA graph?  From the end user perspective this should still give the same (or better) result.

michaelnchin:
Discussed offline with @bechbd. While the `pg_schema` algorithm is a brand new feature, all Neptune Analytics graph instances are automatically upgraded to the newest version, and should therefore support it.

Since there are no concerns around backwards compatibility, we will revise this PR to always use `pg_schema` for graph schema retrieval.

3coins (approved):

krlawrence:
I took a quick and had no major comments. Thanks for making this improvement.

PR #315, opened 2024-12-23
Add support for latency optimized performanceConfig

cipher982:
AWS Bedrock recently introduced support for latency optimization that greatly speeds up tokens/second response times for certain models being rolled out. This PR adds the ability to pass the config through langchain to make use of the faster model calls.

AWS Docs:
- https://docs.aws.amazon.com/bedrock/latest/userguide/latency-optimized-inference.html

Example usage compared to boto3
```python
# BOTO3
client = boto3.client("bedrock-runtime")
response = client.converse(
    modelId=model_id,
    messages=[{"role": "user", "content": [{"text": query}]}],
    performanceConfig={"latency": "optimized"|"standard"},
)

# LANGCHAIN
client = ChatBedrockConverse(
    model_id=model_id,
    performance_config={"latency": "optimized"|"standard"},
)
response = client.invoke(query)
```

Benchmarks I tried locally after the change. Langchain performance closely matches the same configuration in raw boto3 calls (with the exception of some overhead it appears)

benchmark_results

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -393,6 +393,16 @@ class ChatBedrockConverse(BaseChatModel):
     ('auto') if a 'nova' model is used, empty otherwise.
     """
 
+    performance_config: Optional[Dict[str, Any]] = Field(
+        default=None,
+        description="""Performance configuration settings for latency optimization.
+        
+        Example:
+            performance_config={'latency': 'optimized'}
+        If not provided, defaults to standard latency.
+        """,
+    )
+
     model_config = ConfigDict(
         extra="forbid",
         populate_by_name=True,
@@ -623,6 +633,7 @@ class ChatBedrockConverse(BaseChatModel):
         additionalModelRequestFields: Optional[dict] = None,
         additionalModelResponseFieldPaths: Optional[List[str]] = None,
         guardrailConfig: Optional[dict] = None,
+        performanceConfig: Optional[Dict[str, Any]] = None,
     ) -> Dict[str, Any]:
         if not inferenceConfig:
             inferenceConfig = {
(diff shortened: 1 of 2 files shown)


stephenswetonic:
I would love to see this. The speedup from latency optimized is amazing!

joshuasanchezdfx5:
I'm working on a project right now where this speed-up will be amazing.
I'm hoping this will be merged soon, don't make us wait too long. 😉

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -393,6 +393,16 @@ class Joke(BaseModel):
     ('auto') if a 'nova' model is used, empty otherwise.
     """
 
+    performance_config: Optional[Dict[str, Any]] = Field(
```
patilvishal0597, line 396: [Per Bedrock](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_PerformanceConfiguration.html), supported values are `standard`|`optimized`.

Appropriate would be essentially an `StrEnum`

3coins: I think what you are suggesting is a `Optional[Mapping[str, Literal["standard", "optimized"]]]`. That might be too specific, and it's not clear if Bedrock would have any other keys in the `performanceConfig` in future. Can you confirm that, otherwise, I think we can go with Mapping[str, Any] which keeps the option open for other values in future.

patilvishal0597: Yes, let's have Mapping[str, Any] for now for performanceConfig.

cipher982: Just pushed a change to mapping so its a bit more flexible

3coins:
@cipher982 
Looks good! One minor request to add this to the `ChatBedrock._as_converse` as well,  so it can be passed through `model_kwargs` for users of ChatBedrock.

cipher982:
> @cipher982 Looks good! One minor request to add this to the `ChatBedrock._as_converse` as well, so it can be passed through `model_kwargs` for users of ChatBedrock.
> 
> https://github.com/langchain-ai/langchain-aws/blob/ac7ec07c56c33306d2f3b688e3c663fb43cefe2b/libs/aws/langchain_aws/chat_models/bedrock.py#L834

ok! just pushed that update

PR #323, opened 2025-01-11
Raise ValueError if .with_structured_output is called with an unsupported model

GlennViroux:


michaelnchin (approved):
Looks good from testing, thanks @GlennViroux !

michaelnchin (approved):

PR #324, opened 2025-01-13
Add AWS Bedrock Callback

wolfgangcolsman:
This pull request makes the following changes:

- add llm_output to ChatResult in ChatBedrockConverse
- add BedrockTokenUsageCallbackHandler
- add get_bedrock_callback()

The implementation of cache create and cache read is not tested yet as prompt caching is not available to all AWS customers at this time point.

For more information on prompt caching, see 
https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-call.html

### DIFF
--- a/libs/aws/langchain_aws/callbacks/bedrock_callback.py
+++ b/libs/aws/langchain_aws/callbacks/bedrock_callback.py
@@ -1,4 +1,3 @@
-import logging
 import threading
 from typing import Any, Dict, List, Union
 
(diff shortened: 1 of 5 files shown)


michaelnchin (commented):
@wolfgangcolsman, thank you for your submission. 

For now, we're looking to avoid migrating the Bedrock token counting callback class to this package, due to maintenance costs associated with the hard-coded model names and prices. 

However, the changes do look reasonable to me. I would suggest opening a separate pull request in the main LangChain repository to modify the original [Bedrock callback class](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/callbacks/bedrock_anthropic_callback.py) for `langchain-community`.

PR #340, opened 2025-01-28
Adding support for a new runnable for InlineAgents

### LINKED ISSUE #309: Bedrock Inline Agents and Code Interpreter
Curious if there's any LangChain integration capability with these features yet? Here's an example with boto3:

```python
import boto3
client = boto3.client('bedrock-agent-runtime', region_name='us-west-2')

response = client.invoke_inline_agent(
    actionGroups=[
        {
            'actionGroupName': 'CodeInterpreterAction',
            'parentActionGroupSignature': 'AMAZON.CodeInterpreter'
        },
    ],
    enableTrace=True,
    endSession=False,
    foundationModel='anthropic.claude-3-sonnet-20240229-v1:0',
    inputText='Calculate the square root of 169 added to the sum of the fibonacci sequence up to 321.',
    instruction='You are a helpful assistant who can perform complex calculations and return the outputs from running Python code.',
    sessionId='session-1'
)

eventstream = response.get('completion')

for event in eventstream:
    print(event)
```

divekarshubham:
## Description

This PR introduces a new Bedrock Inline-Agents Runnable that allows using [InvokeInlineAgent API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent-runtime_InvokeInlineAgent.html) with return of control functions as tools and configurable agentic properties.

## Usage
See [Jupyter notebook](https://github.com/divekarshubham/langchain-aws/blob/1d963b79621b4d4596d97fab8a1753b6968865cf/samples/agents/inline_agents_roc.ipynb) for usage with `AgentExecutor`

Closes #309


jdbaker01 (changes requested):
Good start, I really like how inline agents fits the interfaces better than the create agent/invoke agent pattern. Left a comment on the interface itself. We should adhere to the messages input/output contract similar to how the chat models do it.

libs/aws/langchain_aws/agents/inline_chat.py
```diff
@@ -0,0 +1,239 @@
...
+)
+
+from .utils import (
+    _get_action_group_and_function_names,
+    _tool_to_function,
+    parse_agent_response,
+    get_boto_session,
+)
+
+class BedrockInlineAgentsChatModel(BaseChatModel):
```
jdbaker01, line 26: Let's not have both BedrockInlineAgentRunnable and BedrockInlineAgentsChatModel. Remove the state based BedrockInlineAgentRunnable implementation and rename BedrockInlineAgentsChatModel to BedrockInlineAgentRunnable. I like how the new implementation supports the standardized message interface and will integrate with LangGraph better. 

Please update the inline_agent_langgraph notebook to use this implementation instead of the current state based approach. I would like use to completely remove "should_continue" just like here in this sample they have no should_continue: https://langchain-ai.github.io/langgraph/how-tos/create-react-agent/#usage

jdbaker01 (changes requested):
I like this better than the state based approach. Left a comment to remove the older executor based approach.

libs/aws/langchain_aws/agents/inline_chat.py
divekarshubham: Thanks, merged the two interfaces and added a langgraph integration notebook

libs/aws/langchain_aws/agents/base.py
```diff
@@ -609,3 +311,219 @@ def _parse_intermediate_steps(
                     return session_id, session_state
 
         return None, None
+
+class BedrockInlineAgentsRunnable(BaseChatModel):
```
3coins, line 315: What was the motivation behind using the `ChatModel` as base rather than a Runnable? 

ChatModel have specific mechanisms, for example `bind_tools` to work with tools, that this implementation will bypass, so this might not be the right interface for the agents.

libs/aws/langchain_aws/agents/utils.py
```diff
@@ -0,0 +1,287 @@
...
+
+    Returns
+        Either a BedrockAgentAction or a BedrockAgentFinish
+    """
+    response_text = ""
+    event_stream = response["completion"]
+    session_id = response["sessionId"]
+    trace_log_elements = []
+    files = []
+    for event in event_stream:
```
nateglims, line 62: Should we have some sort of signal for unknown event types so consumers can more easily identify service model changes?

libs/aws/langchain_aws/agents/utils.py
```diff
@@ -0,0 +1,287 @@
...
+            BedrockAgentAction(
+                tool=tool,
+                tool_input=parameters_json,
+                log=response_text,
+                session_id=session_id,
+                trace_log=trace_log,
+            )
+        ]
+    except Exception as ex:
+        raise Exception("Parse exception encountered {}".format(repr(ex)))
```
nateglims, line 120: Is there an exception hierarchy that we can add onto here? `Exception` is broad.

libs/aws/langchain_aws/agents/utils.py
divekarshubham: Good point, will update this separately.

libs/aws/langchain_aws/agents/utils.py
divekarshubham: Unsure of this one, keeping it broad to encompass all validation + serviceUnavailable exceptions.

libs/aws/langchain_aws/agents/utils.py
divekarshubham: Currently this is not a type within the response, will investigate more and update this in a later PR

3coins:
@divekarshubham 
Can you rebase from main and submit again?

libs/aws/langchain_aws/agents/utils.py
3coins: Agree with @nateglims comment here, catching specific exceptions and then providing a simplified error message for the user will be better, but don't find it blocking for this PR.

nit: Can you use the form `raise Exception(...) from ex`, so the original exception is chained.

libs/aws/langchain_aws/agents/utils.py
```diff
@@ -0,0 +1,289 @@
...
+
+    if idle_session_ttl_in_seconds:
+        create_agent_request["idleSessionTTLInSeconds"] = idle_session_ttl_in_seconds
+
+    create_agent_response = bedrock_client.create_agent(**create_agent_request)
+    request_id = create_agent_response.get("ResponseMetadata", {}).get("RequestId", "")
+    logger.info(f"Create bedrock agent call successful with request id: {request_id}")
+    agent_id = create_agent_response["agent"]["agentId"]
+    create_agent_start_time = time.time()
+    while time.time() - create_agent_start_time < 10:
```
3coins, line 170: Does the agent creation have a 10s absolute timeout? We should figure out a better way to handle this, for example provide an option to control that wait time before throwing the error.

libs/aws/langchain_aws/agents/utils.py
```diff
@@ -0,0 +1,289 @@
...
+            .get("agent", {})
+            .get("agentStatus", {})
+        )
+        if agent_creation_status == "NOT_PREPARED":
+            return agent_id
+        else:
+            time.sleep(2)
+
+    logger.error(f"Failed to create bedrock agent {agent_id}")
+    raise Exception(f"Failed to create bedrock agent {agent_id}")
```
3coins, line 182: Should we use TimeoutError instead of Exception here? Would there be other reasons agent creation can fail here, for example if the role is not correct, does the API provide info for those failures?

libs/aws/langchain_aws/agents/base.py
```diff
@@ -632,3 +323,295 @@ def _parse_intermediate_steps(
                     return session_id, session_state
 
         return None, None
+
+
+class BedrockInlineAgentsRunnable(RunnableSerializable[List[BaseMessage], BaseMessage]):
+    """Invoke Bedrock Inline Agent as a runnable chat model."""
```
3coins, line 329:
```suggestion
    """Invoke Bedrock Inline Agent as a Runnable."""
```

samples/agents/inline_agent_langraph.ipynb
```diff
@@ -0,0 +1,372 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Bedrock Inline Agent Chat Model with LangGraph\n",
+    "\n",
+    "This notebook demonstrates how to use BedrockInlineAgentsChatModel with LangGraph for a mortgage rate assistant that maintains chat history and context.\n",
```
3coins, line 9: Should this refer to `BedrockInlineAgentsRunnable`?

samples/agents/inline_agent_langraph.ipynb
```diff
@@ -0,0 +1,372 @@
...
+    "\n",
+    "tools = [get_asset_value, get_mortgage_rate]\n",
+    "tool_map = {tool.name: tool for tool in tools}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Step 2: Create the BedrockInlineAgentsChatModel"
```
3coins, line 62: Should this refer to BedrockInlineAgentsRunnable?

3coins (changes requested):
@divekarshubham 
Looks good overall! Make sure that in docs and sample notebooks correct class name (BedrockInlineAgentsRunnable) is used.

I left some other comments for improving the code and exception handling; this is not blocking for the PR, but would like to have those updates soon after the initial release.

libs/aws/langchain_aws/agents/utils.py
divekarshubham: Added some more specific exceptions and chaining.

libs/aws/langchain_aws/agents/utils.py
divekarshubham: Ack, will address in future PRs

libs/aws/langchain_aws/agents/utils.py
divekarshubham: Good call. Added broad exception handling to `create_agent` call and updated this to TimeoutError.

3coins:
@divekarshubham 
Code looks good! Before we can merge, can you fix the lint and test errors.

3coins (approved):

PR #341, opened 2025-01-30
Fix Bedrock token count and IDs for Anthropic models

### LINKED ISSUE #314: AttributeError: 'Anthropic' object has no attribute 'count_tokens'
Getting this error when using anthropic models from bedrock.py module of langchain. 

The error comes from this module.

Screenshot 2024-12-23 at 1 35 50 PM

Looking at the anthropic code, the api for count_tokens has been changed and it needs to be updated in the utils.py for bedrock.
import anthropic

client = anthropic.Anthropic()

response = client.messages.count_tokens(
    model="claude-3-5-sonnet-20241022",
    system="You are a scientist",
    messages=[{
        "role": "user",
        "content": "Hello, Claude"
    }],
)

print(response.json())

michaelnchin:
Fixes #314

The `get_num_tokens()` and `get_token_ids()` methods of `BedrockLLM` and `ChatBedrock` currently fail when used with Anthropic models and `anthropic>=0.39.0` installed. 

In such scenarios, this PR fixes `get_num_tokens()` and `get_token_ids()` by falling back to the base class implementations [in `BaseLanguageModel`](https://python.langchain.com/api_reference/core/language_models/langchain_core.language_models.base.BaseLanguageModel.html#langchain_core.language_models.base.BaseLanguageModel.get_num_tokens) and [in `BaseChatModel`](https://python.langchain.com/api_reference/core/language_models/langchain_core.language_models.chat_models.BaseChatModel.html#langchain_core.language_models.chat_models.BaseChatModel.get_num_tokens), which use the HuggingFace [GPT2 Tokenizer](https://huggingface.co/docs/transformers/en/model_doc/gpt2#transformers.GPT2TokenizerFast). 

If `anthropic<=0.38.0` (and other requirements) are present instead, the Anthropic SDK token methods will continue to be used as normal.

**Note about tokenizer accuracy:** 

The GPT2 and Anthropic SDK tokenizers (see [here](https://github.com/anthropics/anthropic-sdk-python/pull/726/files#diff-9595fafc42ceb1044adb6f4f2a93774f58704ee22b6f9d40ad9d0a336c118d39L540-L542)) both may produce inaccurate token estimates for Claude 3 and 3.5 models. To obtain a more accurate estimate, Anthropic recommends using the new [Count Message Tokens API](https://docs.anthropic.com/en/api/messages-count-tokens). For example:
```
import os
import anthropic

os.environ["ANTHROPIC_API_KEY"] = "<your-api-key>"
anthropic.Anthropic().messages.count_tokens(
    model="claude-3-5-sonnet-20241022",
    messages=[
        {"role": "user", "content": "Hello, world"}
    ]
)
```

As another alternative, you can implement your own token counter method, and pass this using [`custom_get_token_ids`](https://python.langchain.com/api_reference/core/language_models/langchain_core.language_models.base.BaseLanguageModel.html#langchain_core.language_models.base.BaseLanguageModel.custom_get_token_ids) when initializing the model. This will override both the GPT2 and Anthropic tokenizers.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -637,7 +637,7 @@ class ChatBedrock(BaseChatModel, BedrockBase):
             else:
                 warnings.warn(
                     f"Falling back to default token method due to conflicts with the Anthropic API: {bad_deps}"
-                    f"\n\nFor Anthropic SDK versions > 0.38.0, it is recommended to provide the chat model class with a"
+                    f"\n\nFor Anthropic SDK versions > 0.38.0, it is recommended to provide the model class with a"
                     f" custom_get_token_ids method that implements a more accurate tokenizer for Anthropic. "
                     f"For get_num_tokens, as another alternative, you can implement your own token counter method "
                     f"using the ChatAnthropic or AnthropicLLM classes."
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -622,26 +622,30 @@ class ChatBedrock(BaseChatModel, BedrockBase):
         return final_output
 
     def get_num_tokens(self, text: str) -> int:
-        if self._model_is_anthropic and check_anthropic_tokens_dependencies():
-            return get_num_tokens_anthropic(text)
-        else:
-            if self._model_is_anthropic:
+        if self._model_is_anthropic:
+            bad_deps = check_anthropic_tokens_dependencies()
+            if not bad_deps:
+                return get_num_tokens_anthropic(text)
+            else:
                 logger.debug(
-                    "Falling back to default token counting due to incompatible or missing Anthropic dependencies. "
-                    "To fix this, ensure that you have anthropic<=0.38.0, httpx<=0.27.2, and Python<=3.12 installed."
+                    "Falling back to default token counting due to incompatible/missing Anthropic dependencies:"
                 )
-            return super().get_num_tokens(text)
+                for x in bad_deps:
+                    logger.debug(x)
+        return super().get_num_tokens(text)
 
     def get_token_ids(self, text: str) -> List[int]:
-        if self._model_is_anthropic and check_anthropic_tokens_dependencies():
-            return get_token_ids_anthropic(text)
-        else:
-            if self._model_is_anthropic:
+        if self._model_is_anthropic:
+            bad_deps = check_anthropic_tokens_dependencies()
+            if not bad_deps:
+                return get_token_ids_anthropic(text)
+            else:
                 logger.debug(
-                    "Falling back to default token ids retrieval due to incompatible or missing Anthropic dependencies."
-                    " To fix this, ensure that you have anthropic<=0.38.0, httpx<=0.27.2, and Python<=3.12 installed."
+                    "Falling back to default token ids retrieval due to incompatible/missing Anthropic dependencies:"
                 )
-            return super().get_token_ids(text)
+                for x in bad_deps:
+                    logger.debug(x)
+        return super().get_token_ids(text)
 
     def set_system_prompt_with_tools(self, xml_tools_system_prompt: str) -> None:
         """Workaround to bind. Sets the system prompt with tools"""
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -1,3 +1,4 @@
+import logging
 import re
 from collections import defaultdict
 from operator import itemgetter
@@ -50,9 +51,12 @@ from langchain_aws.llms.bedrock import (
     _combine_generation_info_for_llm_result,
 )
 from langchain_aws.utils import (
+    check_anthropic_tokens_dependencies,
     get_num_tokens_anthropic,
     get_token_ids_anthropic,
 )
+
+logger = logging.getLogger(__name__)
 
 
 def _convert_one_message_to_text_llama(message: BaseMessage) -> str:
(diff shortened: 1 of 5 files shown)


libs/aws/langchain_aws/utils.py
```diff
@@ -1,21 +1,51 @@
...
-            "This is needed in order to accurately tokenize the text "
-            "for anthropic models. Please install it with `pip install anthropic`."
-        )
+        bad_httpx = "none installed"
+
+    if bad_anthropic:
+        bad_deps.append(f"anthropic<=0.38.0 required, found {bad_anthropic}.")
+    if bad_httpx:
+        bad_deps.append(f"httpx<=0.27.2 required, found {bad_httpx}.")
+
```
3coins, line 43: Curious about these other checks here, would only checking the anthropic version not suffice?

libs/aws/langchain_aws/utils.py
```diff
@@ -1,21 +1,51 @@
...
+from packaging import version
+
 
 def enforce_stop_tokens(text: str, stop: List[str]) -> str:
     """Cut off the text as soon as any stop words occur."""
     return re.split("|".join(stop), text, maxsplit=1)[0]
 
 
-def _get_anthropic_client() -> Any:
+def check_anthropic_tokens_dependencies() -> List[str]:
```
3coins, line 13: nit: Would anthropic_tokens_supported be a better name here?

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -619,15 +623,29 @@ def _combine_llm_outputs(self, llm_outputs: List[Optional[dict]]) -> dict:
...
-            return super().get_num_tokens(text)
+            bad_deps = check_anthropic_tokens_dependencies()
+            if not bad_deps:
+                return get_num_tokens_anthropic(text)
+            else:
+                logger.debug(
+                    "Falling back to default token counting due to incompatible/missing Anthropic dependencies:"
+                )
+                for x in bad_deps:
+                    logger.debug(x)
```
3coins, line 634: Would this cause duplicate logger messages, as `super().get_num_tokens(text)` ends up calling the `get_token_ids` method. Also, adding a logger message would just fall under the radar for most users, unless they have the debug logging switched on, should we rather use a warning?

```python
import warnings

if self._model_is_anthropic and not self.custom_get_token_ids:
    warnings.warn(
        "Falling back to default token counting due to incompatible anthropic count_tokens API. "
        "For anthropic versions > 0.38.0, it is recommended to provide a custom_get_token_ids "
        "method to the chat model class that implements the appropriate tokenizer for Anthropic."
    )
```

3coins (commented):
@michaelnchin 
Thanks for submitting these changes. A question about the version checks, and a suggestion to use of warnings rather than logger statements.

libs/aws/langchain_aws/utils.py
michaelnchin: > bad_deps.append(f"httpx<=0.27.2 required, found {bad_httpx}.")

The `langsmith` chain dependency of `langchain-aws` installs the latest `httpx` version. 
```
Collecting httpx<1,>=0.23.0 (from langsmith<0.4,>=0.1.125->langchain-core<0.4.0,>=0.3.27->langchain-aws==0.2.11)
  Using cached httpx-0.28.1-py3-none-any.whl.metadata (7.1 kB)
```

However, this is incompatible with `anthropic==0.38.0`, which breaks on anything higher than `httpx==0.27.2`. Unfortunately, Anthropic 0.38.0 itself does not cap `httpx` to the version it requires, and subsequently may break after it picks up the version installed by LangSmith SDK:
```
Collecting httpx<1,>=0.23.0 (from anthropic==0.38.0)
  Using cached httpx-0.28.1-py3-none-any.whl.metadata (7.1 kB)
```

So, users must install `httpx<=0.27.2` to get the Anthropic tokens methods working.

michaelnchin: >  bad_deps.append(f"Python 3.12 or earlier required, found {'.'.join(map(str, python_version[:3]))})")

`anthropic==0.38.0` used to be incompatible with Python 3.13 due to issues with a chain dependency (`pyo3-ffi`): https://github.com/anthropics/anthropic-sdk-python/issues/718

But `pyo3-ffi` was recently updated, and the 3.13 installation appears to be fixed now - will remove this check as it is no longer necessary.

libs/aws/langchain_aws/utils.py
michaelnchin: Updated this 👍

libs/aws/langchain_aws/chat_models/bedrock.py
michaelnchin: Thanks for the catch - removed the warning from `get_num_tokens`, and updated the message as suggested.

libs/aws/langchain_aws/utils.py
michaelnchin: Yes, they would hit an exception if using the latest `httpx` version, as Anthropic 0.38.0 uses a `proxies` parameter that was deprecated in httpx [v0.28.0](https://github.com/encode/httpx/releases/tag/0.28.0):
```
TypeError: Client.__init__() got an unexpected keyword argument 'proxies'
```

Anthropic did later add a fix in v0.40.0 to become compatible with `httpx>0.27.2`:
https://github.com/anthropics/anthropic-sdk-python/pull/765
https://github.com/anthropics/anthropic-sdk-python/releases/tag/v0.40.0

However, this wasn't backported to Anthropic 0.38.x.

libs/aws/langchain_aws/utils.py
```diff
@@ -1,21 +1,39 @@
...
+        httpx_import_msg = "httpx<=0.27.2 is required."
+        try:
+            import httpx
+        except ImportError:
+            raise ImportError(httpx_import_msg)
+        httpx_version = version.parse(httpx.__version__)
+        if httpx_version > version.parse("0.27.2"):
+            raise ImportError(httpx_import_msg)
+        else:
+            return True
```
3coins, line 32: Can simplify this by getting rid of the 2 else as we are either raising or return early.
```suggestion
    if version.parse(anthropic.__version__) > version.parse("0.38.0"):
        return False

    try:
        import httpx
        if httpx_version > version.parse("0.27.2"):
            raise ImportError()
    except ImportError:
        raise ImportError("httpx<=0.27.2 is required.")

    return True
```

PR #345, opened 2025-02-01
[Fix] added missed client parameter in ChatBedrock

Amitgb14:
Update the ChatBedrock class to accept the user initiated Bedrock client.

### DIFF
--- a/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
@@ -1,7 +1,6 @@
 """Standard LangChain interface tests"""
 
 from typing import Literal, Type
-from unittest.mock import Mock
 import pytest
 from langchain_core.language_models import BaseChatModel
 from langchain_core.messages import HumanMessage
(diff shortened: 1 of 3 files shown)


libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
```diff
@@ -74,6 +74,33 @@ def test_tool_message_histories_list_content(self, model: BaseChatModel) -> None
         super().test_tool_message_histories_list_content(model)
 
 
+class TestBedrockNovaStandardWithClient(ChatModelIntegrationTests):
```
michaelnchin, line 77: It looks like this tests a specific use case of the client parameter directly from `ChatBedrockConverse`, and does not involve the new code path added to `ChatBedrock`.

Should we consider moving this test out to a separate PR?

michaelnchin (changes requested):
Thank you for the contribution, @Amitgb14. Have one question about the added tests.

libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
Amitgb14: @michaelnchin revert the integration tests case changes.

michaelnchin (approved):
Thanks @Amitgb14 !

PR #346, opened 2025-02-03
chore: update github actions monthly with dependabot

jlosito:
This change will use dependabot to check for updates to any of the GitHub actions this project uses and submit pull requests with version bumps.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      action-dependencies:
+        patterns:
+          - "*"


michaelnchin (changes requested):
Thank you for the contribution, @jlosito. I'm not certain of our general stance regarding automated changes, but this seems like a useful change to keep us informed about available updates.

I have one concern - by default, Dependabot opens individual pull requests for every new dependency update it detects. This could potentially end up spamming the repository with new PRs every time the monthly check occurs. 

To prevent this, can we make use of the [Grouped Version Updates](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--) feature? This would allow Dependabot to roll multiple changes into the same PR to keep things cleaner, and prevent possible conflicts when co-dependent packages each have pending updates.

For example:
```
version: 2
updates:
 - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "monthly"
    groups:
      actions-dependencies:
        patterns:
          - "*"
```

We may also want to enhance this to suggest updates for the main project.
```
- package-ecosystem: "pip"
    directory: "/"
    schedule:
      interval: "monthly"
    groups:
      prod-dependencies:
        dependency-type: "production"
      dev-dependencies:
        dependency-type: "development"
```

3coins:
@efriis 
Any concern on adding `dependabot` alerts to the repo here? Noticed that `langchain` had attempted this at some point, but discarded now.
https://github.com/langchain-ai/langchain/actions/workflows/dependabot/dependabot-updates

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
```
3coins, line 6: Should we change to `weekly`? Also, this seem to only include updates to Github Actions, should we also add `pip`, so it can cover python packages?

efriis:
I believe it might be better to configure in the `security` tab at the top without doing in a github action. I believe this posts security issues live. This is how we do them in the langchain monorepo

But this likely works too, and no issues with doing in a workflow!

jlosito:
@3coins I had opened a similar change for the core lib. Right now it's in draft mode though because I haven't verified that all of the tests pass.

https://github.com/langchain-ai/langchain/pull/29554

3coins:
@jlosito 
I concur with suggestions from @michaelnchin. Would you be able to make these updates. The `pip` workflow seems a bit more involved, so we need some deep dive into that, but we can start with `github-actions` at the moment.

```
version: 2
updates:
 - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    groups:
      actions-dependencies:
        patterns:
          - "*"
```

3coins:
> @3coins I had opened a similar change for the core lib. Right now it's in draft mode though because I haven't verified that all of the tests pass.
> 
> [langchain-ai/langchain#29554](https://github.com/langchain-ai/langchain/pull/29554)

Someone from LangChain team would be able to review that, but based on info from @efriis, it seems like the `langchain` monorepo is handling this from the settings panel.

efriis:
closed the monorepo one - yes to be clear we access them by enabling [these settings (linking the langchain-aws one for convenience)](https://github.com/langchain-ai/langchain-aws/security)

jlosito:
@efriis this change is slightly different than the security related pull requests that dependabot opens. Dependabot will open security related pull requests using the method you described, but this change is more of a cron job that constantly tries to update packages regardless whether there is a security update or not.

jlosito:
@3coins did you want me to include the python updates within this pull request or did you want to close this out completely?

3coins:
@jlosito 
Closing this issue, completed the setup with #426.

PR #347, opened 2025-02-04
[Fix] Adding default description in tools

### LINKED ISSUE #336: tools require a filled description
Generally langchain tools default to having an empty string as a description. Some (maybe all) bedrock models require a filled description, which affects the ease of use of `with_structured_output(method="function_calling")` and `bind_tools` with `ChatBedrock`

In particular, this impacts model interoperability, where people are trying their tools with different models, and the bedrock versions simply fail.

Some ideas:
- populate the description with the same as the tool name by default in bedrock to avoid the error
- raise a more descriptive error to help people populate docstrings on their tool function/basemodel/etc

I'm probably more in favor of auto populating for models this impacts

cc @hinthornw

patilvishal0597:
Adding tool name as default tool descriptions if description is an empty string or not passed

Fixes #336

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -1030,6 +1030,9 @@ def _format_tools(
             spec = convert_to_openai_tool(tool)["function"]
             spec["inputSchema"] = {"json": spec.pop("parameters")}
             formatted_tools.append({"toolSpec": spec})
+        
+        tool_spec = formatted_tools[-1]["toolSpec"]
+        tool_spec["description"] = tool_spec.get("description") or tool_spec["name"]
     return formatted_tools
 
 
--- a/libs/aws/langchain_aws/function_calling.py
+++ b/libs/aws/langchain_aws/function_calling.py
@@ -211,15 +211,13 @@ def convert_to_anthropic_tool(
     if isinstance(tool, dict) and all(
         k in tool for k in ("name", "description", "input_schema")
     ):
-        if not tool["description"]:
-            tool["description"] = tool["name"]
+        tool["description"] = tool["description"] or tool["name"]
         return AnthropicTool(tool)  # type: ignore
     else:
         formatted = convert_to_openai_tool(tool)["function"]
-        description = formatted["description"] if formatted["description"] \
-            else formatted["name"]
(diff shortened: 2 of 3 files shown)


libs/aws/langchain_aws/function_calling.py
```diff
@@ -211,9 +211,13 @@ def convert_to_anthropic_tool(
     if isinstance(tool, dict) and all(
         k in tool for k in ("name", "description", "input_schema")
     ):
+        if not tool["description"]:
+            tool["description"] = tool["name"]
```
michaelnchin, line 215: Nit: can simplify to
```
tool["description"] = tool["description"] or tool["name"]
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -1030,6 +1030,10 @@ def _format_tools(
             spec = convert_to_openai_tool(tool)["function"]
             spec["inputSchema"] = {"json": spec.pop("parameters")}
             formatted_tools.append({"toolSpec": spec})
+        
+        tool_spec = formatted_tools[-1]["toolSpec"]
+        if "description" not in tool_spec or not tool_spec["description"]:
+            tool_spec["description"] = tool_spec["name"]
```
michaelnchin, line 1036: Nit: use `dict.get()`
```
tool_spec["description"] = tool_spec.get("description") or tool_spec["name"]
```

libs/aws/langchain_aws/function_calling.py
```diff
@@ -211,9 +211,13 @@ def convert_to_anthropic_tool(
...
         return AnthropicTool(tool)  # type: ignore
     else:
         formatted = convert_to_openai_tool(tool)["function"]
+        if "description" not in formatted or not formatted["description"]:
+            formatted["description"] = formatted["name"]
```
michaelnchin, line 220: Nit: use `dict.get()`
```
formatted["description"] = formatted.get("description") or formatted["name"]
```

michaelnchin (changes requested):
Thanks @patilvishal0597 - tested and looks good with several models that were previously failing.

michaelnchin (approved):
LGTM!

michaelnchin:
Additionally, verified that this fix works when passing multiple tools of mixed descriptions.

Screenshot 2025-02-11 at 7 31 55 PM

PR #350, opened 2025-02-06
Enhance `embed_documents` to make use of Cohere's ability to request multiple embeddings at once

jimfingal:
The [Cohere embedding provider](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html) supports asking for multiple embeddings at the same time.

The current implementation of `embed_documents` always iterates across the submitted texts and attempts to embed them one-by-one.

This PR updates this `embed_documents` logic to check whether we are using the cohere provider, and if so request all of the embeddings at once.

### DIFF
--- a/libs/aws/langchain_aws/embeddings/bedrock.py
+++ b/libs/aws/langchain_aws/embeddings/bedrock.py
@@ -78,6 +78,11 @@ class BedrockEmbeddings(BaseModel, Embeddings):
         protected_namespaces=(),
     )
 
+    @property
+    def provider(self) -> str:
+        """Provider of the model."""
+        return self.model_id.split(".")[0]
+
     @model_validator(mode="after")
     def validate_environment(self) -> Self:
         """Validate that AWS credentials to and python package exists in environment."""
@@ -121,20 +126,38 @@ class BedrockEmbeddings(BaseModel, Embeddings):
         return self
 
     def _embedding_func(self, text: str) -> List[float]:
-        """Call out to Bedrock embedding endpoint."""
+        """Call out to Bedrock embedding endpoint with a single text."""
         # replace newlines, which can negatively affect performance.
         text = text.replace(os.linesep, " ")
 
         # format input body for provider
-        provider = self.model_id.split(".")[0]
         input_body: Dict[str, Any] = {}
-        if provider == "cohere":
+        if self.provider == "cohere":
             input_body["input_type"] = "search_document"
             input_body["texts"] = [text]
         else:
             # includes common provider == "amazon"
             input_body["inputText"] = text
 
+        response_body = self._invoke_model(input_body)
+
+        if self.provider == "cohere":
+            return response_body.get("embeddings")[0]
+        else:
+            return response_body.get("embedding")
+
+    def _cohere_multi_embedding(self, texts: List[str]) -> List[float]:
+        """Call out to Cohere Bedrock embedding endpoint with multiple inputs."""
+        # replace newlines, which can negatively affect performance.
+        texts = [text.replace(os.linesep, " ") for text in texts]
+
+        input_body: Dict[str, Any] = {}
+        input_body["input_type"] = "search_document"
+        input_body["texts"] = texts
+
+        return self._invoke_model(input_body).get("embeddings")
+
+    def _invoke_model(self, input_body: Dict[str, Any] = {}) -> Dict[str, Any]:
         if self.model_kwargs:
             input_body = {**input_body, **self.model_kwargs}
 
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/embeddings/bedrock.py
```diff
@@ -121,20 +126,36 @@ def validate_environment(self) -> Self:
...
+            return response_body.get("embedding")
+
+    def _cohere_multi_embedding(self, texts: List[str]) -> List[float]:
+        """Call out to Cohere Bedrock embedding endpoint with multiple inputs."""
+        # replace newlines, which can negatively affect performance.
+        texts = [text.replace(os.linesep, " ") for text in texts]
 
+        input_body: Dict[str, Any] = {}
+        input_body["input_type"] = "search_document"
+        input_body["texts"] = texts
```
3coins, line 154:
```suggestion
        input_body: Dict[str, Any] = {
            "input_type": "search_document",
            "texts": texts
        }
```

libs/aws/langchain_aws/embeddings/bedrock.py
```diff
@@ -78,6 +78,11 @@ class BedrockEmbeddings(BaseModel, Embeddings):
         protected_namespaces=(),
     )
 
+    @property
+    def provider(self) -> str:
+        """Provider of the model."""
+        return self.model_id.split(".")[0]
```
3coins, line 84: 👍 Makes sense to move to a common method.

libs/aws/langchain_aws/embeddings/bedrock.py
jimfingal: Good callout -- made this simpler / terser with: [fb3e3c5](https://github.com/langchain-ai/langchain-aws/pull/350/commits/fb3e3c597ed4dc26392172c5c8fbb318261aa025)

jimfingal:
@3coins This would be the case for any cohere provider -- the API interface is different for [Cohere](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html) vs. [Amazon embeddings](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-titan-embed-text.html), with Amazon just accepting one input and Cohere expecting an array of texts.

romenlee:
It breaks our application where we embed 200 documents at one time and receive an error:
```
An error occurred (ValidationException) when calling the InvokeModel operation: Malformed input request: #/texts: expected maximum item count: 128, found: 217, please reformat your input and try again.
```

if there are 100 documents then:
```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: Invalid parameter combination. Please check and try again.
```

it's not a minor change.

jimfingal:
> It breaks our application where we embed 200 documents at one time and receive an error:
> 
> ```
> An error occurred (ValidationException) when calling the InvokeModel operation: Malformed input request: #/texts: expected maximum item count: 128, found: 217, please reformat your input and try again.
> ```
> 
> if there are 100 documents then:
> 
> ```
> botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: Invalid parameter combination. Please check and try again.
> ```
> 
> Only for 50 it's Ok.
> 
> it's not a minor change.

Interesting -- looks like [the embedding API ](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html) wants at most 96 texts, or 2048 characters. That will require more advanced batching logic, I can take a shot at that.

romenlee:
@jimfingal 
Thank you.

You can try chunk texts in `_cohere_multi_embedding` method. Something like:
```
MAX_COHERE_BATCH_SIZE = 96
for i in range(0, len(texts), MAX_BATCH_SIZE):
    batch_texts= split_docs[i : i + MAX_BATCH_SIZE]
    self._invoke_model(
            input_body={
                "input_type": "search_document",
                "texts": batch_texts,....
```

jimfingal:
> @jimfingal Thank you.
> 
> You can try chunk texts in `_cohere_multi_embedding` method. Something like:
> 
> ```
> MAX_COHERE_BATCH_SIZE = 96
> for i in range(0, len(texts), MAX_COHERE_BATCH_SIZE):
>     batch_texts= texts[i : i + MAX_COHERE_BATCH_SIZE]
>     self._invoke_model(
>             input_body={
>                 "input_type": "search_document",
>                 "texts": batch_texts,....
> ```

@romenlee Made an attempted fix with: https://github.com/langchain-ai/langchain-aws/pull/376 -- the max character count gives an extra bit of fun complexity.

Issue #354, opened 2025-02-09
Update Streaming Support Validator for AWS Bedrock Models (e.g. Meta, AI21, Mistral)


martinwangjian:
#### Description

Our AWS Bedrock integration currently misidentifies the streaming support behavior for some models. According to the [AWS Bedrock Models Documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html), several providers—such as Anthropic, Cohere, AI21 Labs, Meta, and Mistral—support streaming mode. For Amazon models, streaming is enabled only for specific model IDs that contain keywords like `"nova-lite"`, `"nova-micro"`, `"nova-pro"`, `"titan-text-express"`, `"titan-text-lite"`, or `"titan-text-premier"`.

However, we have observed that a Meta model (e.g. `"us.meta.llama3-3-70b-instruct-v1:0"`) is being flagged as having streaming disabled. This behaviour is inconsistent with the AWS documentation, where Meta models are expected to support streaming.

---

#### Steps to Reproduce
AWS Region: `us-east-1`
```python
from langchain_aws import ChatBedrock
from langchain_core.messages import HumanMessage

# Instantiate a ChatBedrock model (use a model ID that should support streaming)
chat = ChatBedrock(model_id="us.meta.llama3-3-70b-instruct-v1:0")  # You can also test with "meta.llama3-3-70b-instruct-v1:0" without cross region inference profile

# Create a sample human message
message = HumanMessage(content="Hello")

# Initiate streaming for the message
stream = chat.stream([message])

# Retrieve and print the first response from the stream
response = next(stream)
print(response)
```

####  Expected Behaviour
Streaming Enabled: Models from Anthropic, Cohere, AI21 Labs, Meta, and Mistral should have streaming enabled (i.e. disable_streaming is False).
Amazon Models: For Amazon models, streaming should be enabled only if the model ID contains one of the approved keywords.

#### Actual Behaviour
`ERROR	ValidationException: This model doesn't support tool use in streaming mode.`
The current behaviour incorrectly disables streaming for models like "us.meta.llama3-3-70b-instruct-v1:0", even though the AWS documentation confirms that such models support streaming.

PR #355, opened 2025-02-09
Update Streaming Validator for AWS Bedrock Models

martinwangjian:
This pull request updates the streaming validator logic in our AWS Bedrock integration to align with the latest AWS documentation. According to the [AWS Bedrock Models Documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html), models from Anthropic, Cohere, AI21 Labs, Meta, and Mistral support streaming mode, while only specific Amazon models (those with IDs containing keywords like "nova-lite", "nova-micro", "nova-pro", "titan-text-express", "titan-text-lite", or "titan-text-premier") support streaming.

#### Changes Made

- **Streaming Validator Update:**  
  Modified the `set_disable_streaming` validator in `ChatBedrockConverse` to:
  - Recognize Meta, AI21 Labs, and Mistral models as having streaming enabled.
  - For Amazon models, enable streaming only when the model ID contains one of the approved keywords.

- **Integration Tests Update:**  
  - Enhanced the parameterized test for streaming mode (`test_set_disable_streaming`) to include additional cases, including the new test case for `"us.meta.llama3-3-70b-instruct-v1:0"`.
  - Ensured that models supporting streaming correctly have `disable_streaming` set to `False`, while models without streaming support set this flag to `"tool_calling"`.

#### Benefits

- **Alignment with AWS Official Documentation:**  
  This update ensures that our integration behaves as expected based on the official [AWS Bedrock Models Documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html).

- **Improved User Experience:**  
  Users leveraging AWS Bedrock models will now have the correct streaming behavior, avoiding unexpected validation errors.

#### Related Issue

- This PR addresses the concerns raised in [Issue #354](https://github.com/langchain-ai/langchain-aws/issues/354).

Please review the changes and provide any feedback or additional suggestions.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -420,19 +420,44 @@ class ChatBedrockConverse(BaseChatModel):
     def set_disable_streaming(cls, values: Dict) -> Any:
         model_id = values.get("model_id", values.get("model"))
         model_parts = model_id.split(".")
-        values["provider"] = values.get("provider") or (
+
+        # As of 2025-02-09: based on the AWS documentation: https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html
+        # Extract provider from the model_id (e.g., "amazon", "anthropic", "ai21", "meta", "mistral")
+        provider = values.get("provider") or (
             model_parts[-2] if len(model_parts) > 1 else model_parts[0]
         )
+        values["provider"] = provider
 
-        # As of 12/03/24:
-        # Anthropic, Cohere and Amazon Nova models support streamed tool calling
+        # Determine if the model supports streaming based on updated documentation:
+        streaming_supported = False
+
+        # Providers that always support streaming
+        if provider in ["anthropic", "cohere", "ai21", "meta", "mistral"]:
+            streaming_supported = True
+
+        # For Amazon, only specific models support streaming:
+        # e.g. amazon.nova-lite, amazon.nova-micro, amazon.nova-pro,
+        # amazon.titan-text-express, amazon.titan-text-lite, amazon.titan-text-premier
+        elif provider == "amazon":
+            amazon_streaming_keywords = [
+                "nova-lite",
+                "nova-micro",
+                "nova-pro",
+                "titan-text-express",
+                "titan-text-lite",
+                "titan-text-premier",
+            ]
+            if any(keyword in model_id for keyword in amazon_streaming_keywords):
+                streaming_supported = True
+
+        # Set the disable_streaming flag accordingly
+        # If streaming is supported, we want it enabled (i.e. disable_streaming == False)
+        # Otherwise, set it to "tool_calling" (which you can handle downstream as a validation error or alternative mode)
         if "disable_streaming" not in values:
             values["disable_streaming"] = (
-                False
-                if values["provider"] in ["anthropic", "cohere"]
-                or (values["provider"] == "amazon" and "nova" in model_id)
-                else "tool_calling"
+                False if streaming_supported else "tool_calling"
             )
+
         return values
 
     @model_validator(mode="after")
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -420,45 +420,58 @@ class ChatBedrockConverse(BaseChatModel):
     def set_disable_streaming(cls, values: Dict) -> Any:
         model_id = values.get("model_id", values.get("model"))
         model_parts = model_id.split(".")
-
-        # As of 2025-02-09: based on the AWS documentation: https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html
+        
         # Extract provider from the model_id (e.g., "amazon", "anthropic", "ai21", "meta", "mistral")
-        provider = values.get("provider") or (
-            model_parts[-2] if len(model_parts) > 1 else model_parts[0]
-        )
+        provider = values.get("provider") or (model_parts[-2] if len(model_parts) > 1 else model_parts[0])
         values["provider"] = provider
 
-        # Determine if the model supports streaming based on updated documentation:
-        streaming_supported = False
+        model_id_lower = model_id.lower()
 
-        # Providers that always support streaming
-        if provider in ["anthropic", "cohere", "ai21", "meta", "mistral"]:
-            streaming_supported = True
+        # Determine if the model supports plain-text streaming (ConverseStream)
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -420,19 +420,44 @@ class Joke(BaseModel):
...
             model_parts[-2] if len(model_parts) > 1 else model_parts[0]
         )
+        values["provider"] = provider
+
+        # Determine if the model supports streaming based on updated documentation:
+        streaming_supported = False
+
+        # Providers that always support streaming
+        if provider in ["anthropic", "cohere", "ai21", "meta", "mistral"]:
+            streaming_supported = True
```
3coins, line 436: From the linked issue, it seems like this PR is trying to fix tool calling in streaming mode. Is that correct?
In that case, do all these models support streaming tool calls with Bedrock Converse? Looking at this list from documentation, it doesn't seem like that's the case.

martinwangjian: thanks for review, if I understand well, the purpose of set_disable_streaming function is not limited to handling tool calls, it governs the overall streaming behavior of the model, in our case is about usage of llama3-3 with streaming mode, you can find my python code to reproduce in https://github.com/langchain-ai/langchain-aws/issues/354

from my point of view, the function determines whether streaming is enabled based on two factors:

1. General Streaming Support:
 It checks if the model supports streaming for plain conversational output (using the ConverseStream capability from the AWS documentation https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-supported-models-features.html for Converse).

2. Tool-Calling Context:
If a tool is being used (i.e. if the tools argument is provided), disable streaming (by returning "tool_calling") because streaming tool calls are not supported—even if the model supports plain-text streaming( column "Streaming tool use" of https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-supported-models-features.html)

Is that correct? if yes, i will update my PR

3coins: Thanks for providing that explanation. It makes sense to enable streaming for these providers, but `disable_streaming` is a bit subtle, and I want to make sure that setting `disable_streaming` as False for some of these providers (for example meta), would not have any impact to tool calling or usage with `with_structured_output` in streaming mode. If it does, we might want to set the correct value "tool_calling" for those providers. 

Here is some more info on this from LangChain package. The `BaseChatModel` uses [_should_stream](https://github.com/langchain-ai/langchain/blob/9cd20080fc259a84ed326fa492edffc2ca063d22/libs/core/langchain_core/language_models/chat_models.py#L338-L342) to determine behavior in [stream](https://github.com/langchain-ai/langchain/blob/9cd20080fc259a84ed326fa492edffc2ca063d22/libs/core/langchain_core/language_models/chat_models.py#L360-L364) and switches to `invoke` in cases where `disable_streaming` is False or "tool_calling" (only in tool calls). We do want to enable streaming but only in non tool calling modes and setting `disable_streaming` to False for meta models will use streaming, and will end up as rejected by the Bedrock API. 

There are additional updates in [ChatBedrockConverse](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L607-L624), that uses `disable_streaming` to switch output parsers, which should be correctly handled as this pertains to tool calling.

@michaelnchin will do some validations and report back, I think this is an important fix, and we want to merge it before the next release.

michaelnchin (changes requested):
Thanks @martinwangjian ! Following up on @3coins 's comments, let's take these changes a little further, and add additional checks/handling for setting the value of [`disable_streaming`](https://python.langchain.com/api_reference/core/language_models/langchain_core.language_models.chat_models.BaseChatModel.html#langchain_core.language_models.chat_models.BaseChatModel.disable_streaming), based whether the model supports tool calls with streaming.

The updated code should look something like this:
```python
@model_validator(mode="before")
    @classmethod
    def set_disable_streaming(cls, values: Dict) -> Any:
        model_parts = model_id.split(".")

        # As of 2025-02-09: based on the AWS documentation: https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html
        # Extract provider from the model_id (e.g., "amazon", "anthropic", "ai21", "meta", "mistral")
        provider = values.get("provider") or (
            model_parts[-2] if len(model_parts) > 1 else model_parts[0]
        )
        values["provider"] = provider

        # Determine if the model supports streaming based on updated documentation:
        model_id_lower = model_id.lower()

        if (
                # AI21 Jamba 1.5 models
                (provider == "ai21" and "jamba-1-5" in model_id_lower) or
                # Some Amazon Nova models
                (provider == "amazon" and any(x in model_id_lower for x in [
                    "nova-lite", "nova-micro", "nova-pro"
                ])) or
                # Anthropic Claude 3 and newer models
                (provider == "anthropic" and "claude-3" in model_id_lower) or
                # Cohere Command R models
                (provider == "cohere" and "command-r" in model_id_lower)
        ):
            streaming_support = True
        elif (
                # AI21 Jamba-Instruct model
                (provider == "ai21" and "jamba-instruct" in model_id_lower) or
                # Amazon Titan Text models
                (provider == "amazon" and "titan-text" in model_id_lower) or
                # Anthropic older Claude models (Claude 2, Claude 2.1, Claude Instant)
                (provider == "anthropic" and any(x in model_id_lower for x in [
                    "claude-v2", "claude-instant"
                ])) or
                # Cohere Command (non-R) models
                (provider == "cohere" and "command" in model_id_lower and "command-r" not in model_id_lower) or
                # All Meta Llama models
                (provider == "meta") or
                # All Mistral models
                (provider == "mistral")
        ):
            streaming_support = "no_tools"
        else:
            streaming_support = False

        # Set the disable_streaming flag accordingly
        # If streaming is supported, we want it enabled (i.e. disable_streaming == False)
        # Otherwise, set it to "tool_calling" (which you can handle downstream as a validation error or alternative mode)
        if "disable_streaming" not in values:
            if not streaming_support:
                values["disable_streaming"] = True
            elif streaming_support == "no_tools":
                values["disable_streaming"] = "tool_calling"
            else:
                values["disable_streaming"] = False

        return values
```

This will allow

martinwangjian:
> michaelnchin

Hi @michaelnchin, thanks for the feedback. I agree with your suggestions and have updated the PR accordingly. The updated validator now correctly distinguishes plain-text streaming (using ConverseStream) from tool-calling (setting "tool_calling"). Please review when you have a moment. Thanks!

michaelnchin (commented):
LGTM, thanks @martinwangjian !

michaelnchin (approved):

PR #356, opened 2025-02-10
Added support for Prompt Caching for Anthropic when using InvokeModel API

patcheng:
Using langchain's example for creating SystemMessage and HumanMessage for prompt caching:
https://js.langchain.com/docs/integrations/chat/anthropic/#prompt-caching

Looks like SystemMessage's content could be non-string, added support for that.

Using AWS Bedrock's InvokeModel API prompt caching example on what message is supposed to structure: 
https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#large-context-caching-example

Doesn't look like `SystemMessage` supports prompt caching, so ignore the `cache_control`.

But preserve that for `HumanMessage`

Added tests for the change


michaelnchin:
Integration tests pass: https://github.com/langchain-ai/langchain-aws/actions/runs/13864108781/job/38799222179

michaelnchin (approved):
Thanks @patcheng !

nitinkalra-bst:
@patcheng @michaelnchin 

what do you mean by this "Doesn't look like SystemMessage supports prompt caching, so ignore the cache_control."

Anthropic docs clearly add         "cache_control": {"type": "ephemeral"} to system prompt as well.

https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#large-context-caching-example

Issue #245, opened 2024-10-20
Lack of logging in the langchain_aws package


cab938:
The `langchain_aws` package, in particular the `ChatBedrockConverse` class, has no standard python logging in it. This means visibility into issues is pretty dramatically reduced, and one has to either log at the boto3 level or across into AWS cloud resources. It would be useful to add more common python logging to the various classes in the `langchain_aws` package.

As an example real use case, I want to see the body of any request and responses sent to boto so I can understand how the language model might be interpreting some of the data. I expected I would see them if I did something like:

```logging.getLogger('langchain_aws').setLevel(logging.DEBUG)```

However, that doesn't do anything because the module doesn't log, and the best I could do was:

```logging.getLogger('boto3').setLevel(logging.DEBUG)```

This shows me request headers (not body, which is a problem), and responses (including body, yay!)

3coins:
@cab938 
LangChain provides some built-in utilities to log application data, that doesn't require adding logging statements directly to the integration classes.

1. [LangSmith](https://www.langchain.com/langsmith) is the best tool for reviewing/monitor your langchain application. Here is the setup [guide](https://docs.smith.langchain.com/).
![image](https://github.com/user-attachments/assets/f986d376-d414-4832-82a2-6e4726ef72f9)


2. Use the `set_debug` flag. Here is an example:
  ```python
  from langchain_core.globals import set_debug
  set_debug(True)
  
  llm = ChatBedrockConverse(model="anthropic.claude-3-haiku-20240307-v1:0")
  llm.invoke("What is the capital of France?")
  
  # Output
  [llm/start] [llm:ChatBedrockConverse] Entering LLM run with input:
  {
    "prompts": [
      "Human: What is the capital of France?"
    ]
  }
  [llm/end] [llm:ChatBedrockConverse] [261ms] Exiting LLM run with output:
  {
    "generations": [
      [
        {
          "text": "The capital of France is Paris.",
          "generation_info": null,
          "type": "ChatGeneration",
          "message": {
            "lc": 1,
            "type": "constructor",
            "id": [
              "langchain",
              "schema",
              "messages",
              "AIMessage"
            ],
            "kwargs": {
              "content": "The capital of France is Paris.",
              "response_metadata": {
                "ResponseMetadata": {
                  "RequestId": "e411108a-0918-4d28-8030-a9b16f7c8ea1",
                  "HTTPStatusCode": 200,
                  "HTTPHeaders": {
[...]

PR #358, opened 2025-02-10
removed unecessary debug statements

aaronbriel:
The bedrock chat model has several logging statements that seem to be left over from someone debugging the code. Some of them result in the entire prompt being logged to the console, cluttering logging of applications leveraging the module. This PR removes said logging statements.


3coins:
@aaronbriel 
Agree that some of these are redundant (info), but the debug ones are something that were specifically called out in #245. These should only be logged if you have the env variable `LANGCHAIN_AWS_DEBUG=1` or otherwise have the logging set to debug.

aaronbriel:
@3coins I restored the debug statements as requested.

PR #362, opened 2025-02-14
Add CodeInterpreter flag to create_agent

jdbaker01:
1. Add CodeInterpreter to the create_agent method for the existing agent executor implementation. 
2. Add the ability to return files in the output.
3. Add example notebook showing how to use the code interpreter and the files output.

Related to Issue 309 - Bedrock Inline Agents and Code Interpreter
https://github.com/langchain-ai/langchain-aws/issues/309


3coins (approved):
LGTM!

PR #364, opened 2025-02-17
Use `logger.exception` to log errors inside exception handlers and avoid logging from the root logger

edgarrmondragon:
## Rationale

- Not using the root logger lets the end user better control logging for specific packages and modules, e.g. by configuring `langchain_aws.llms.bedrock` to log at the `ERROR` level, but `langchain_aws.llms.bedrock` at the `DEBUG` level.
- Using `logger.exception` adds exception info to the [log record](https://docs.python.org/3/library/logging.html#logrecord-attributes). This is particularly useful if you want to use a library like [python-json-logger](https://pypi.org/project/python-json-logger/) or [structlog](https://github.com/hynek/structlog), where the exception info could be nicely formatted into an array/object.


3coins (approved):
LGTM!

PR #376, opened 2025-02-27
fix: text and character limits on Cohere embedding API

jimfingal:
In a [previous PR](https://github.com/langchain-ai/langchain-aws/pull/350), I added support for retrieving more than one embedding at once from the Cohere embedding API.

@romenlee [pointed out](https://github.com/langchain-ai/langchain-aws/pull/350#issuecomment-2688770323) that this causes problems with a large number of texts. It looks like the [Cohere embedding API](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html) has a limit of either 96 texts, or 2048 characters.

This PR implements more intelligent batching that respects these limits -- it batches texts in chunks that are at most 96 texts or 2048 characters, submits them to the embedding API, and combines the results.

There is a complexity vs. efficiency tradeoff call to make here for the maintainers. This code is much more complex than the original naive code, pre #350 -- but it is much more efficient with network calls so will execute faster. An alternative if we don't want to take on this complexity would be to just revert #350, which would restore functionality for embedding large texts at the cost of giving up performance improvements.

### DIFF
--- a/libs/aws/langchain_aws/embeddings/bedrock.py
+++ b/libs/aws/langchain_aws/embeddings/bedrock.py
@@ -2,7 +2,7 @@ import asyncio
 import json
 import logging
 import os
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Generator, List, Optional
 
 import numpy as np
 from langchain_core.embeddings import Embeddings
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/embeddings/bedrock.py
```diff
@@ -262,3 +269,39 @@ async def aembed_documents(self, texts: List[str]) -> List[List[float]]:
         result = await asyncio.gather(*[self.aembed_query(text) for text in texts])
 
         return list(result)
+
+
+def _batch_cohere_embedding_texts(texts: List[str]) -> Generator[List[str], None, None]:
+    """Batches a set of texts into chunks that are acceptable for the Cohere embedding API:
+    chunks of at most 96 items, or 2048 characters."""
+
+    # Cohere embeddings want a maximum of 96 items and 2048 characters
```
3coins, line 278: Is this documented anywhere? Can we add that link here in the comments.

romenlee: > [Cohere embedding API](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html) has a limit of either 96 texts, or 2048 characters.

jimfingal, line 278:
```suggestion
    # Cohere embeddings want a maximum of 96 items and 2048 characters
    # See: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html
```

jimfingal: @3coins added in [60b6205](https://github.com/langchain-ai/langchain-aws/pull/376/commits/60b6205f3f471815bb7312ba1767c0dfd8b3348f)

3coins (approved):
@jimfingal 
Great job and thanks for fixing this. Looks good!

PR #380, opened 2025-02-28
aws[patch]: allow structured output when thinking is enabled for Claude 3.7

ccurme:
Structured output will currently always raise a ValidationException when Claude 3.7 Sonnet's `thinking` is enabled, because we rely on forced tool use for structured output and this feature is not supported when `thinking` is enabled.

Here we:
- Emit a warning if `with_structured_output` is called when `thinking` is enabled.
- Raise `OutputParserException` if no tool calls are generated.

This is arguably preferable to raising an error in all cases.

```python
from pydantic import BaseModel


class Person(BaseModel):
    name: str
    age: int


llm = ChatBedrockConverse(
    model="us.anthropic.claude-3-7-sonnet-20250219-v1:0",
    max_tokens=5000,
    additional_model_request_fields={
        "thinking": {"type": "enabled", "budget_tokens": 2000}
    },
)
structured_llm = llm.with_structured_output(Person)  # <-- this generates a warning
```

```python
structured_llm.invoke("Alice is 30.")  # <-- works
```

```python
structured_llm.invoke("Hello!")  # <-- raises OutputParserException
```

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -679,10 +679,13 @@ class ChatBedrockConverse(BaseChatModel):
             tool_choice = "any"
         else:
             tool_choice = None
-        if tool_choice:
-            llm = self.bind_tools([schema], tool_choice=tool_choice)
-        else:
+        if tool_choice is None and "claude-3-7-sonnet" in self.model_id:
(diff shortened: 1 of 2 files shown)


michaelnchin (approved):
LGTM, thanks @ccurme

EllangoK:
Question if possible, is this true for just Converse or also normal ChatBedrock

PR #382, opened 2025-03-02
extras checks after boto3 bedrock-runtime client invoke

xming-cn:
In the ChatBedrockConverse class, additional checks have been added after invoking the bedrock-runtime boto3 client to ensure that errors are not delayed in cases where the response is successful but invalid. [langchain#29729](https://github.com/langchain-ai/langchain/issues/29729)

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -574,6 +574,8 @@ class ChatBedrockConverse(BaseChatModel):
             messages=bedrock_messages, system=system, **params
         )
         logger.debug(f"Response from Bedrock: {response}")
+        if 'output' not in response:
+            raise ValueError(f"AWS returns an invalid response, key 'output' is not in reponse.")
         response_message = _parse_response(response)
         return ChatResult(generations=[ChatGeneration(message=response_message)])
 
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -575,7 +575,15 @@ class ChatBedrockConverse(BaseChatModel):
         )
         logger.debug(f"Response from Bedrock: {response}")
         if 'output' not in response:
-            raise ValueError(f"AWS returns an invalid response, key 'output' is not in reponse.")
+            if self.endpoint_url:
+                raise ValueError(
+                    "The AWS response is invalid as it does not contain the expected 'output' key. "
+                    "Please verify that the correct bedrock-runtime endpoint is being used."
+                )
+            else:
+                raise ValueError(
+                    "The AWS response is invalid as it does not contain the expected 'output' key. "
+                )
         response_message = _parse_response(response)
         return ChatResult(generations=[ChatGeneration(message=response_message)])
 


3coins:
@xming-cn 
Not sure if this is a necessary check. The issue you have linked and the sample code, seems to be using the `bedrock` rather than `bedrock-runtime` endpoint that `converse` works with.

xming-cn:
你的进度真的很快，马上就要结束了，祝你好运bWdlYWVlXzE1X2ZzdGlhaG4=HEX5845 is big

stuaxo:
Is there any way to surface the ClientError from boto ?

It's trying to tell us we are misconfigured but we hide that and raise something generic.

stuaxo:
I guess ValueError will do, though having the ClientError (maybe in our own wrapper) would be better.

Misconfiguration is that uncommon, and software like Django has exceptions for it when the user does something like Misconfigure the database.

3coins:
> Misconfiguration is that uncommon, and software like Django has exceptions for it when the user does something like Misconfigure the database.

This would be valuable if we could provide some info to the user, stating that they used a wrong endpoint so they could correct it. The current message doesn't provide anything useful info other than that the response was invalid. Does the service API provide this info that could be surfaced to the user?

xming-cn:
Unfortunately, there isn't any obvious way to effectively determine the service from which a response originated

```python
>>> INCORRECT_ENDPOINT = 'https://bedrock.us-east-1.amazonaws.com'
>>> messages = [{"role": "ai", "content": ({"text": "hello"}, )}]
>>> bedrock_runtime = boto3.client('bedrock-runtime', endpoint_url=INCORRECT_ENDPOINT)
>>> bedrock_runtime.converse(modelId="a", messages=messages)
{
    'ResponseMetadata': {
        'RequestId': '38b36959-780b-474e-bb56-10c0c7e6dc4c', 
        'HTTPStatusCode': 200, 
        'HTTPHeaders': {
            'date': 'Fri, 14 Mar 2025 05:27:03 GMT', 
            'content-type': 'application/json', 
            'content-length': '90', 
            'connection': 'keep-alive', 
            'x-amzn-requestid': '38b36959-780b-474e-bb56-10c0c7e6dc4c'
        }, 
        'RetryAttempts': 0
    }
}
```

stuaxo:
This is a good start, I think we can tweak the message a little - and maybe avoid using and `if` since we are not on the happy path.

Here's what we know / don't know at this point -
- We know it's not a converse response, because it doesn't have an output, there's a doc for that: https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html#API_runtime_Converse_ResponseSyntax
- We know it *may* be because of a misconfiguration, but not which one (we can suggest though).


Looking at what we have `_parse_resonse` is expecting a dict with 'output', in theory this could be sent for any reason - so this is a good spot to raise a ValueError that it's not a Converse Response.

The call site this PR is at calls `_parse_response` with the result of the client call, so this is a good point wrap `_parse_response` in a exception handler, catch the ValueError from it and add our hints.

Then in both of these we can have wording using what we know/don't know.

Would these changes make sense on the PR ?

```python
def _parse_response(response: Dict[str, Any]) -> AIMessage:
    try:
        output = response.pop("output")
    except KeyError:
        raise ValueError("No 'output' key, this is not a converse-response.  https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html#API_runtime_Converse_ResponseSyntax")
    lc_content = _bedrock_to_lc(output["message"]["content"])
    tool_calls = _extract_tool_calls(lc_content)
    usage = UsageMetadata(_camel_to_snake_keys(response.pop("usage")))  # type: ignore[misc]
    return AIMessage(
        content=_str_if_single_text_block(lc_content),  # type: ignore[arg-type]
        usage_metadata=usage,
        response_metadata=_extract_response_metadata(response),
        tool_calls=tool_calls,
    )

...

try
    response_message = _parse_response(response)
except ValueError as e:
    ValueError(f"Unexpected endpoint response. The response format does not match Bedrock Converse API. This may indicate a misconfiguration (e.g., using 'bedrock' instead of 'bedrock-runtime', incorrect region, or endpoint URL). Original error: {str(e)}")
```

3coins:
Should we rather add a validator for the `endpoint_url` so we check for this at initialization? This should help with catching the misconfiguration early rather than waiting for the bad response to occur.

```python
    @field_validator('endpoint_url')
    @classmethod
    def validate_endpoint(cls, v: Optional[str]) -> Optional[str]:  
        if v is not None and not v.starts_with("bedrock-runtime"):
          raise ValueError(
              "Incorrect endpoint_url provided. Use the correct endpoint that starts with bedrock-runtime. "
              "See https://docs.aws.amazon.com/general/latest/gr/bedrock.html for a list of valid"
              "endpoint_url values."
          )
        return v

```

xming-cn:
> Should we rather add a validator for the `endpoint_url` so we check for this at initialization? This should help with catching the misconfiguration early rather than waiting for the bad response to occur.
> 
> ```python
>     @field_validator('endpoint_url')
>     @classmethod
>     def validate_endpoint(cls, v: Optional[str]) -> Optional[str]:  
>         if v is not None and not v.startswith("bedrock-runtime"):
>           raise ValueError(
>               "Incorrect endpoint_url provided. Use the correct endpoint that starts with bedrock-runtime. "
>               "See https://docs.aws.amazon.com/general/latest/gr/bedrock.html for a list of valid"
>               "endpoint_url values."
>           )
>         return v
> ```

@3coins This approach might have some limitations worth considering. While checking the "bedrock-runtime" prefix could be helpful in many cases, it may not fully guarantee endpoint validity without actual requests. For instance, VPC endpoint URLs like vpce-xxxxxxx.bedrock-runtime.us-east-1.vpce.amazonaws.com or custom DNS configurations through Route53 private zones might deviate from this pattern. This means there could be valid scenarios where the url validation alone wouldn’t reliably confirm the endpoint’s validity.

stuaxo:
It could lock out compatible bedrock implementations.

Still, it's a good url to mention in an error.

3coins:
> This approach might have some limitations worth considering. While checking the "bedrock-runtime" prefix could be helpful in many cases, it may not fully guarantee endpoint validity without actual requests. For instance, VPC endpoint URLs like vpce-xxxxxxx.bedrock-runtime.us-east-1.vpce.amazonaws.com or custom DNS configurations through Route53 private hosted zones might deviate from this pattern. This means there could be valid scenarios where the url validation alone wouldn’t reliably confirm the endpoint’s validity

Yeah, that makes sense. I would suggest keeping single error, where you are checking for the `output` key, and provide a generic message.

```python
if 'output' not in response:
  raise ValueError(
    "Not 'output' key found in the response from the Bedrock Converse API. This usually "
    "happens due to misconfiguration of endpoint or region, ensure that you are using valid "
    "values for endpoint_url (starting with bedrock-runtime) listed here: "
    "https://docs.aws.amazon.com/general/latest/gr/bedrock.html"
  )
```

I would also suggest adding the documentation link to the endpoint_url doc comments. The current doc comments don't seem that useful, we should emphasize here that for most users this value should be not set, as the API takes care of calling the right endpoint based on the default region or the region supplied by the user. 
https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L406-L407

stuaxo:
I'll submit a PR, I think it's OK to go via Exception handling since in the happy path this will "output" will always be there.

stuaxo:
Hope this doesn't step on any toes, I did this as a PR as it's maybe quicker than just commenting.
(I actually don't care who's name is on the commit, but the functionality would be good).

https://github.com/langchain-ai/langchain-aws/pull/430

> Yeah, that makes sense. I would suggest keeping single error, where you are checking for the output key, and provide a generic message.

- [x] Done.

This uses an Exception. which since this is the un-happy path, is probably correct here.

I took the error message from your suggestion with a small tweak:

> raise ValueError(
            "No 'output' key found in the response from the Bedrock Converse API.  This usually "
            "happens due to misconfiguration of endpoint or region, ensure that you are using valid "
            "values for endpoint_url (on AWS this starts with bedrock-runtime), see: "
            "https://docs.aws.amazon.com/general/latest/gr/bedrock.html"
        )

This is a little hint thast bedrock-runtime is the AWS default (this keeps the text correct if in future other services provide Bedrock Converse compatibility).

3coins:
Closing as #430 supersedes this PR.

PR #390, opened 2025-03-05
Add user agent extra field to header for Bedrock Agent runtime invocations

jdbaker01:
Add to the config field for the user_agent_extra field.
Add unit tests to cover no config, partial config, all fields in config.


jdbaker01:
Simplifying the InlineAgent use with LangGraph by adding ToolMessage support.

libs/aws/langchain_aws/agents/base.py
```diff
@@ -506,6 +532,22 @@ def _convert_messages_to_text(self, messages: List[BaseMessage]) -> str:
...
+                isinstance(last_message, AIMessage)
+                and last_message.additional_kwargs is not None
+                and "roc_log" in last_message.additional_kwargs
+            ):
+                return json.loads(
+                    last_message.additional_kwargs.get("roc_log", "{}")
+                ).get("returnControl", {})
+            msg_index -= 1
+
+        return None
```
3coins, line 549: Would suggest avoiding index based iteration and separate conditions. 

```python
for message in reversed(messages) -> Optional[Dict[str, Any]]:
    if not isinstance(message, AIMessage):
        continue
            
    roc_log = message.additional_kwargs.get("roc_log")
    if not roc_log:
        continue
            
    try:
        return json.loads(roc_log).get("returnControl", {})
    except json.JSONDecodeError:
        # log error
        continue

    return None
```

libs/aws/langchain_aws/agents/utils.py
```diff
@@ -20,29 +20,62 @@
 
 logger = logging.getLogger(__name__)
 
+__version__ = "0.1.0"
```
3coins, line 23: `__version__` has a special meaning in python, which signifies the version of the module loaded. That's not the intent here. Should we use `bedrock_agents_version` here? Might also be useful to add a note to update this value for future updates in Bedrock.

libs/aws/langchain_aws/agents/utils.py
```diff
@@ -20,29 +20,62 @@
...
+SDK_USER_AGENT = f"LangChainAWS#Agents#{__version__}"
+
+# Set default client parameters
+DEFAULT_CONFIG = Config(
+    connect_timeout=120,
+    read_timeout=120,
+    retries={"max_attempts": 0},
+)
```
3coins, line 31: nit: Can simplify this a bit, see related suggestion at the bottom.

```python
# Default values to check and set
DEFAULT_CONFIG_VALUES = {
    "connect_timeout": 120,
    "read_timeout": 120,
    "retries": {"max_attempts": 0}
}
```

libs/aws/langchain_aws/agents/utils.py
```diff
@@ -20,29 +20,62 @@
...
+        if getattr(config, "read_timeout", None) is None:
+            config.read_timeout = DEFAULT_CONFIG.read_timeout
+        if getattr(config, "retries", None) is None:
+            config.retries = DEFAULT_CONFIG.retries
+    else:
+        config = DEFAULT_CONFIG
+
+    existing_user_agent = getattr(config, "user_agent_extra", "") or ""
+    new_user_agent = f"{existing_user_agent} md/sdk_user_agent/{SDK_USER_AGENT}".strip()
+    config.user_agent_extra = new_user_agent
```
3coins, line 70: nit: Can simplify this a bit (would scale if add new default values), see related suggestion above.

```python
config = config or Config(**DEFAULT_CONFIG_VALUES)

# Set default values if not present in custom config
for key, default_value in default_values.items():
    if getattr(config, key, None) is None:
        setattr(config, key, default_value)

# Update user agent
existing_user_agent = getattr(config, "user_agent_extra", "") or ""
config.user_agent_extra = f"{existing_user_agent} md/sdk_user_agent/{SDK_USER_AGENT}".strip()

```

3coins (commented):
@jdbaker01 
Thanks for submitting this change. Left some suggestions to simplify some of the code, looks good otherwise.

3coins (approved):
Thanks for all the updates. Looks good!

PR #398, opened 2025-03-13
Fix issue with tool_response block ordering after message merge

berkybear:
as discussed in the following thread, the newly added _merge_messages function does not respect the proper ordering for tool_result blocks following tool_use and results in the following exception:
"Did not find 1 tool_result block(s) at the beginning of this message"

This change fixes the ordering to be in the form that claude expects.
link to discussion: https://github.com/langchain-ai/langgraph/discussions/2161


berkybear:
Looks like there is a similar attempt at a fix in this pr that has already merged. Will close this pr. Looking forward to a release.

PR #405, opened 2025-03-15
Fix ChatBedrock streaming with DeepSeek-R1 distilled models

michaelnchin:
This PR updates ChatBedrock's DeepSeek-R1 stream response handling, compensating for differences between the base model and distilled model formats.

For example, a base model stream chunk looks like:
```
{
  "choices": [
    {
      "text": "hello",
      "stop_reason": null
    }
  ]
}
```

The Llama and Qwen distills both follow the format:
```
{
  "choices": [
    {
      "index": 0,
      "text": "hello",
      "logprobs": null,
      "finish_reason": ""
    }
  ],
  "model": "/opt/ml/model",
  "system_fingerprint": "3.0.1-native",
  "object": "text_completion",
  "id": "",
  "created": 1742011074
}
```

### DIFF
--- a/libs/aws/tests/integration_tests/chat_models/test_bedrock.py
+++ b/libs/aws/tests/integration_tests/chat_models/test_bedrock.py
@@ -137,7 +137,7 @@ def test_chat_bedrock_streaming_deepseek_r1() -> None:
     assert response.usage_metadata
 
 
-@pytest.mark.xfail(reason="Provisioned instance unavailable.")
+@pytest.mark.skip("Needs provisioned instance setup.")
 def test_chat_bedrock_streaming_deepseek_r1_distill_llama() -> None:
     chat = ChatBedrock(  # type: ignore[call-arg]
         provider="deepseek",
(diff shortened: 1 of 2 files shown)


libs/aws/tests/integration_tests/chat_models/test_bedrock.py
```diff
@@ -120,6 +120,59 @@ def test_chat_bedrock_streaming_llama3() -> None:
...
+    for chunk in chat.stream([message]):
+        response += chunk # type: ignore[assignment]
+
+    assert response.content
+    assert response.response_metadata
+    assert response.usage_metadata
+
+
+@pytest.mark.xfail(reason="Provisioned instance unavailable.")
+def test_chat_bedrock_streaming_deepseek_r1_distill_qwen() -> None:
```
3coins, line 159: Should we rather mark these as `@pytest.mark.skip`, since these will be executed in CI. Not sure, if we will ever have a provisioned instance to enable these in the CI. Let me know what you think.

michaelnchin: Agreed - updated to `skip`

3coins (approved):

Issue #381, opened 2025-02-28
BedrockEmbeddings Cohere Models does not support input_type


flipthedog:
https://github.com/langchain-ai/langchain-aws/blob/cd6728a71626f64c366d96f0c0c2b517c05c8e2f/libs/aws/langchain_aws/embeddings/bedrock.py#L130

This function does not properly support the different input types for Cohere models in Bedrock. This leads to serious performance degradation as it always uses "search_document" input_type. "search_query" needs to be implemented in order to get the full functionality of a similarity search here. 

When embedding the document, search_document should be used, but when querying the "search_query" input type is essential to perform a similarity search.

More info here: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html

Happy to discuss

3coins:
@flipthedog 
Thanks for bringing this up. Seems like the `input_type` is not something that can be applied to other models. Do you have a proposed solution to accommodate this option?

flipthedog:
I think this might be a pretty big issue as anyone who is using the BedrockEmbeddings langchain-aws function and the cohere model will have pretty bad search results. At least with my testing. Link to docs on input_type: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed.html

I took a look through the code and the following has to be changed:

The embed_query function needs additional logic like other functions with if `self.provider == "cohere"`. And then add an input_type input variable for all of the next functions called where `input_type="search_query"`.

https://github.com/langchain-ai/langchain-aws/blob/cd6728a71626f64c366d96f0c0c2b517c05c8e2f/libs/aws/langchain_aws/embeddings/bedrock.py#L224

up until:

https://github.com/langchain-ai/langchain-aws/blob/cd6728a71626f64c366d96f0c0c2b517c05c8e2f/libs/aws/langchain_aws/embeddings/bedrock.py#L130



It looks like your current implementation always defaults to `input_type=search_document`. The question is whether we want to keep that and just override for a retrieval have `input_type` be based on whether you called `embed_query()` or `embed_document()`

3coins:
@flipthedog 
Do you want to take a first pass and submit a PR with those changes?

flipthedog:
Yep, I'll take a look

michaelnchin:
Thank you for the contribution, @flipthedog !

PR #406, opened 2025-03-15
added input_type for cohere embedding models

flipthedog:
To fix bug with Cohere embedding models input_type for embedding a query rather than a document. #381 

Small change to add input_type to _embedding_func(), still defaults input_type to "search_document" to ensure that multi embedding and other functions are unchanged.

### DIFF
--- a/libs/aws/langchain_aws/embeddings/bedrock.py
+++ b/libs/aws/langchain_aws/embeddings/bedrock.py
@@ -127,7 +127,7 @@ class BedrockEmbeddings(BaseModel, Embeddings):
 
         return self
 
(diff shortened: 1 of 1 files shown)


libs/aws/langchain_aws/embeddings/bedrock.py
```diff
@@ -127,15 +127,18 @@ def validate_environment(self) -> Self:
 
         return self
 
-    def _embedding_func(self, text: str) -> List[float]:
+    def _embedding_func(self, text: str, input_type="search_document") -> List[float]:
```
3coins, line 130:
```suggestion
    def _embedding_func(self, text: str, input_type: str = "search_document") -> List[float]:
```

Issue #395, opened 2025-03-11
DeepSeek R1 streaming support


Ryuya1995:
Hi Team

I tried to use ChatBedrockConverse with the new released deepseek R1 model (model id: us.deepseek.r1-v1:0). I can get the full response after 10~20 seconds waiting but not the streaming response. With the same code I can get streaming response for Claude Sonnet 3.7. 

Wondering if it's the limitation on bedrock side or the implementation on ChatBedrockConverse.

Thank you

tahouse:
Added a fix for ChatBedrockConverse, working in my local tests https://github.com/langchain-ai/langchain-aws/pull/407

PR #400, opened 2025-03-13
Support DeepSeek-R1 streaming with ChatBedrock

michaelnchin:
Related: #395

This PR updates `LLMInputOutputAdapter` to correctly handle streaming responses from DeepSeek-R1 when using `ChatBedrock`.

PR #407, opened 2025-03-16
Add DeepSeek-R1 'no_tools' streaming support for ChatBedrockConverse

tahouse:
Fixes https://github.com/langchain-ai/langchain-aws/issues/395

DeepSeeks supports ConverseStream but not Streaming Tool Use: https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-supported-models-features.html

Similar to https://github.com/langchain-ai/langchain-aws/pull/400 but for the ChatBedrockConverse. Added a test to ensure no_tools streaming is working. 

Tests pass, and manually testing appears to work as well.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -492,7 +492,9 @@ class ChatBedrockConverse(BaseChatModel):
             # All Meta Llama models
             (provider == "meta") or
             # All Mistral models
-            (provider == "mistral")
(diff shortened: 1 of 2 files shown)


tahouse:
@3coins added a separate but related PR https://github.com/langchain-ai/langchain-aws/pull/408 for enabling DeepSeek conversations. Can combine if preferred.

3coins (approved):
@tahouse 
Looks good!

However, this list is getting out of hand, and in future we should look into moving these rules to a dict. Will open a separate issue with some suggestions.

PR #417, opened 2025-03-18
Fix max_tokens for DeepSeek-R1 base

### LINKED ISSUE #414: ChatBedrock wrong max_tokens parameter conversion - not compatible with AWS boto3 Invoke API
Hi, 

I want to initialize a ChatBedrock instance as follows:
```python
ChatBedrock(model_id='us.deepseek.r1-v1:0',provider='deepseek', region_name='us-east-1', max_tokens=5000, temperature=0)
``` 

However I got the following error:
`botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: Validation Error`

When I debug the issue, following line in the bedrock.py refines the `max_token` parameter as `max_new_tokens` but that is wrong. You can find the actual parameter name in here : `https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-deepseek.html` which is directly `max_tokens`.

https://github.com/langchain-ai/langchain-aws/blob/1b61a6f5e901272eeb997731ec98b64044544a60/libs/aws/langchain_aws/llms/bedrock.py#L370

Removing that conversion will resolve the problem.

michaelnchin:
Fixes #414.

Bedrock Invoke API accepts both `max_new_tokens` and `max_tokens` for DeepSeek-R1-Distill-Qwen and DeepSeek-R1-Distill-Llama models. However, the DeepSeek-R1 base model is less flexible, only accepting `max_tokens`.

For better compatibility, this PR updates BedrockBase to use `max_tokens` for all DeepSeek-R1 models.

### DIFF
--- a/libs/aws/langchain_aws/llms/bedrock.py
+++ b/libs/aws/langchain_aws/llms/bedrock.py
@@ -367,7 +367,7 @@ class LLMInputOutputAdapter:
                 elif provider == "mistral":
                     input_body["max_tokens"] = max_tokens
                 elif provider == "deepseek":
-                    input_body["max_new_tokens"] = max_tokens
+                    input_body["max_tokens"] = max_tokens
                 else:
                     # TODO: Add AI21 support, param depends on specific model.
                     pass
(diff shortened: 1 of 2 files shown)


3coins (approved):
Looks good! Thanks.

Issue #422, opened 2025-03-20
support expanded options for tool choice in ChatBedrockConverse


ccurme:
Following https://aws.amazon.com/about-aws/whats-new/2025/03/amazon-nova-expands-tool-converse-api/

PR #424, opened 2025-03-20
aws[patch]: support tool_choice in Nova models

ccurme:
Addresses #422.

https://aws.amazon.com/about-aws/whats-new/2025/03/amazon-nova-expands-tool-converse-api/


michaelnchin (approved):
Looks good - thanks @ccurme !

PR #426, opened 2025-03-20
Added config for dependabot grouped updates.

3coins:
Continuing on the work in #346. I have enabled dependabot alerts in the settings.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+ - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      actions-dependencies:
+        patterns:
+          - "*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+updates:
+ - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
```
michaelnchin, line 6: Should we change to this to weekly as mentioned [here](https://github.com/langchain-ai/langchain-aws/pull/346#discussion_r1941819410), to better cover our typical release cadence? 

Bi-weekly would be optimal - this will be possible in the near future, as DependaBot is adding support for cron jobs soon: https://github.com/dependabot/dependabot-core/issues/6339#issuecomment-2628628530

michaelnchin (changes requested):

michaelnchin (approved):
LGTM, thanks @3coins !

PR #427, opened 2025-03-20
Updated log statements.

### LINKED ISSUE #421: BedrockChat overlogging
I've noticed that ChatBedrock responses are logged at Info level, without an easy way to configure that. I'm quite surprised that it's happening by default because many companies have policies that logs can't contain PII, which can come back from LLM (if it was sent to LLM).

I understand that it's created as a reaction to https://github.com/langchain-ai/langchain-aws/issues/245, but it seems excessive to me that all responses are logged in plain text, without ability to configure it.

Additionally, langchain supports verbose mode (https://python.langchain.com/docs/how_to/debugging/#set_verbosetrue).

Any chance you can consider dropping responses to debug level or make it configurable (and probably turned off by default)?

3coins:
Fixes #421

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -737,7 +737,7 @@ class ChatBedrock(BaseChatModel, BedrockBase):
             )
         else:
             usage_metadata = None
-        logger.info(f"The message received from Bedrock: {completion}")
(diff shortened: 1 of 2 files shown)


3coins:
@mario-at-intercom
These are the only statements (info) I found that are logging response payloads. Also, `langchain-aws` doesn't set the log level for the project, it's probably set at the root logger either in environment or application. Let me know if this change solves your logging issue.

PR #430, opened 2025-03-21
Raise a ValueError if the bedrock response with data in an unexpected…

stuaxocabinetoffice:
Raise a ValueError if the bedrock response with data in an unexpected format - this can indicate misconfiguration on the users part.

Many thanks to @3coins and @xming-cn for finding doing most of the work on this - finding where this should go and giving feedback on the wording, + finding the right AWS doc


stuaxo:
@3coins cheers, I have committed your change.

3coins (approved):
Looks good!

stuaxo:
Thanks @3coins hopefully this will save someone some hassle in the future, it may even be me if I have a break between doing langchain projects.

PR #431, opened 2025-03-22
Add base model ID field to ChatBedrock models

### LINKED ISSUE #228: Error - custom imported model on Bedrock
- Lanchain V 0.3.2
- Lanchain AWS v 0.2.2

We are using a fine-tuned version of Llama 3.1-instruct, uploaded to Bedrock. Since we are using an ARN model ID (which does not contain any information about the specific Foundation Model used), we encountered an issue.

In the code `chat_models/bedrock.py` at line 349, there is an if statement evaluating the model string to choose between Llama2 and Llama3 for prompt conversion.

In our case, we need to use `convert_messages_to_prompt_llama3`, but the logic falls into the else statement, which uses `convert_messages_to_prompt_llama`.

Is there any solution to ensure the correct conversion function is used?

Thank you!

### LINKED ISSUE #253: Support for Custom and Imported models in ChatBedrock/ChatBedrockConverse
## Problem
The current Bedrock API has 5 different types of model ids that it supports as documented [here](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html). While some of these have a defined pattern for the provider/model and can be used to extract these values, others don't have the necessary components to derive these values.

Both `ChatBedrockConverse` and `ChatBedrock` classes currently rely on the `provider` and `model` values to determine some of the logic required to process provider and model specific code. Our current approach of parsing the model id and determining the provider and model though works for foundation models and cross-region model ids, it does not work well for custom and imported models, that might not have the necessary components to derive the provider and model values. Here are examples of the code used in the 2 implementations.

https://github.com/langchain-ai/langchain-aws/blob/95794e1c91bc5431a2fec80b21cc7a95d5480421/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L402-L416

https://github.com/langchain-ai/langchain-aws/blob/95794e1c91bc5431a2fec80b21cc7a95d5480421/libs/aws/langchain_aws/llms/bedrock.py#L712-L733

### Short Term
1. Current implementation do have a `provider` field available that users can use to pass in the provider value. This does need a slight change in the `ChatBedrockConverse` where we should raise an exception if either a custom or imported model id or arn is passed by the user and the provider value is missing.
2. We also need a `base_model` or a `base_model_id` field that users can pass in to identify the base model used to create the custom or imported model. 

### Long Term
Longer term, the model info should be built into the Bedrock API, so that when new types of models or providers are introduced, we don't have to update logic in chat model implementations to accommodate them. Having this within the Bedrock API would also be helpful in other places not just LangChain implementation. This API should provide a comprehensive set of capabilities for the model not just the provider, or base model. For example, streaming, tool support, tool streaming, multi-modal support etc and should work for all kinds of models not just foundation models.

While the current Bedrock API `get-foundation-model` provides some info about the foundation models, it is not clear if it would work for custom and imported models. It also doesn't give any information about whether the model supports tool calls, streaming tool calls or other features. Here is an example of running the API call for the Anthropic haiku model.

```shell
% aws bedrock get-foundation-model --model-identifier anthropic.claude-3-haiku-20240307-v1:0
{
    "modelDetails": {
        "modelArn": "arn:aws:bedrock:us-west-2::foundation-model/anthropic.claude-3-haiku-20240307-v1:0",
        "modelId": "anthropic.claude-3-haiku-20240307-v1:0",
        "modelName": "Claude 3 Haiku",
        "providerName": "Anthropic",
        "inputModalities": [
            "TEXT",
            "IMAGE"
        ],
        "outputModalities": [
            "TEXT"
        ],
        "responseStreamingSupported": true,
        "customizationsSupported": [],
        "inferenceTypesSupported": [
            "ON_DEMAND"
        ],
        "modelLifecycle": {
            "status": "ACTIVE"
        }
    }
}

```

**Note**: Bedrock does seem to have a `get-custom-model` API as well, but I don't have a custom model or imported model setup to verify these APIs. There is also a `list-foundation-models` and `list-custom-models` APIs, that provide the same info for all models present in a user's account, however these seem to be a heavy-handed approach than the get APIs that provide info for specific models.

### LINKED ISSUE #394: Chat Template Selection in AWS Bedrock for Custom Model Import Meta
When importing a custom model into AWS Bedrock, the system detects the chat template based on the model ID. However, the issue arises because the model ID is an ARN and does not explicitly contain the model name.

By default, AWS Bedrock assigns the Llama 2 chat template for Meta models, but this becomes problematic when using Llama 3 or other custom models. There is currently no flexibility to specify the correct chat template during model import.

michaelnchin:
Fixes #228 
Fixes #253 
Fixes #394 

Added an optional `base_model_id`(or `base_model`) parameter to BedrockBase and ChatBedrockConverse. 

This provides an alternate means of identifying the base model if `model_id` is ambiguous (i.e. a custom or provisioned model ARN). In particular, this is useful for applying specific Bedrock request tuning (ex. prompt templating in ChatBedrock, and streaming support levels in ChatBedrockConverse) for different models under the same provider.


3coins (approved):
LGTM!

PR #437, opened 2025-03-26
Excludes tool calling spec from camelcase conversion.

### LINKED ISSUE #409: ChatBedrockConverse bind_tools doesn't work as expected
Hi 

I recently switched from ChatBedrock to ChatBedrockConverse for testing Claude 3.7 features in an agent system. I noticed that after switching to ChatBedrockConverse, tool use failure rate increased a lot. After dive deep, seems like ChatBedrockConverse is using a different way to bind_tools. 

For example, when I bind a list_directory tool to ChatBedrock and ask LLM to describe the input schema for it. I will get:

{
  "name": "list_directory",
  "parameters": {
    "properties": {
      "**dir_path**": {
...

But when I do the same for ChatBedrockConverse, I will get:

{
  "name": "list_directory",
  "parameters": {
    "properties": {
      "**dirPath**": {
...

The main difference is "dir_path" (correct one) vs "dirPath" (wrong one). It makes ChatBedrockConverse using wrong parameters in many cases therefore increasing the failure rates for tool use.

It seems that bind_tools in ChatBedrockConverse is converting tools to OpenAI functions but ChatBedrock is converting tools to Anthropic tools. Could you please check if there is a better way to bind tools in ChatBedrockConverse?

I am also wondering what's the plan for choosing ChatBedrock vs ChatBedrockConverse? Some features like Claude 3.7 reasoning are having different implementations on those two and requires the user to learn the different config manners. 

Thanks

3coins:
Fixes #409

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -592,7 +592,7 @@ class ChatBedrockConverse(BaseChatModel):
         logger.debug(f"input message to bedrock: {bedrock_messages}")
         logger.debug(f"System message to bedrock: {system}")
         params = self._converse_params(
-            stop=stop, **_snake_to_camel_keys(kwargs, excluded_keys={"inputSchema"})
+            stop=stop, **_snake_to_camel_keys(kwargs, excluded_keys={"inputSchema", "function"})
         )
         logger.debug(f"Input params: {params}")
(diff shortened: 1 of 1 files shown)


michaelnchin (approved):
LGTM, thanks @3coins

ccurme:
This appears to have broken integration tests on `main` for tool calling, see example here (I just ran one test to demonstrate): https://github.com/langchain-ai/langchain-aws/actions/runs/14113050001

PR #446, opened 2025-04-01
Populate prompt-caching related UsageMetadata field.

patcheng:
UsageMetadata was not populated the extra fields are being returned.

Tested ChatBedrock to Haiku, and ChatBedrockConverse to Haiku


michaelnchin (changes requested):
Thanks @patcheng! I tested both modified models with caching, outputs look good. 

Related ask, can you also bump the `boto3` dependency floor [here](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/pyproject.toml#L16) to [v1.37.24](https://github.com/boto/boto3/blob/develop/CHANGELOG.rst#L15)? This will ensure that `cachePoint` will not result in an error when passed to ChatBedrockConverse.

michaelnchin (approved):
Thanks @patcheng !

PR #452, opened 2025-04-07
Drops tool choices for nova models in non-us regions.

### LINKED ISSUE #435: Structured output and tool calling not working for Nova
Testing both structured output and tool calling provide the same error: An error occurred (ModelErrorException) when calling the Converse operation: The system encountered an unexpected error during processing. Try your request again. This doesn't occur when using normal generation, and it works without using Langchain.

Code to replicate:

```python
from langchain_core.messages import SystemMessage, HumanMessage
from pydantic import Field, BaseModel
from models import OutputGuardrail

client = client(
    service_name='bedrock-runtime',
    region_name='eu-west-1',

)
additionalModelRequestFields={"inferenceConfig": {"topK": 1}}

nova_micro = ChatBedrockConverse(
    model="eu.amazon.nova-micro-v1:0",
    client=client,
    region_name="eu-west-1",
    temperature=1,
    top_p=1,
    additional_model_request_fields=additionalModelRequestFields,
)
class TestModel(BaseModel):
    """A test model to check the tool use conversion."""
    name: str = Field(description="The name of the user.")
    age: int = Field(description="The age of the user.")

model = nova_micro.bind_tools([TestModel], tool_choice="any")
messages = [SystemMessage(content="You are a chatbot assistant."), HumanMessage(content="Hi")]
response = model.invoke(messages)
print(response)```



Update: I've been testing this functionality with the Converse API directly, and believe it's a problem with that, not the LangChain implementation. Whenever I try to use 'any' or named tool parameter, it gives the error above.

3coins:
Fixes #435

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -567,7 +567,7 @@ class ChatBedrockConverse(BaseChatModel):
                     self.supports_tool_choice_values = ("auto", "any", "tool")
             elif "mistral-large" in self._get_base_model():
                 self.supports_tool_choice_values = ("auto", "any")
-            elif "nova" in self._get_base_model():
+            elif "nova" in self._get_base_model() and self._get_base_model().startswith("us"):
                 self.supports_tool_choice_values = ("auto", "any", "tool")
             else:
                 self.supports_tool_choice_values = ()


michaelnchin (approved):
LGTM

3coins:
Closing this in favor of passing in the `supports_tool_choice_values` value directly. The concern is that this change adds another dimension to the tool choice logic, and will limit any tool choice value updates for users. Will like to rather rely on the Bedrock's documentation (or releases) to guide us through these changes.

PR #460, opened 2025-04-17
Add support of Prompt Caching for tools

sharonxiaohanli:
1. add prompt caching for tools to bedrock converse api
2. unit tests and integration tests have been added 
3. local run on unit tests and integration tests:

poetry run pytest tests/integration_tests/chat_models/test_bedrock_converse.py::test_tool_use_with_cache_point     
...
tests/integration_tests/chat_models/test_bedrock_converse.py .                                                             [100%]

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -479,13 +479,19 @@ class ChatBedrockConverse(BaseChatModel):
         # (e.g., "amazon", "anthropic", "ai21", "meta", "mistral")
         if "provider" not in values:
             if model_id.startswith("arn"):
-                raise ValueError("Model provider should be supplied when passing a model ARN as model_id.")
+                raise ValueError(
+                    "Model provider should be supplied when passing a model ARN as model_id."
+                )
             model_parts = model_id.split(".")
-            values["provider"] = model_parts[-2] if len(model_parts) > 1 else model_parts[0]
+            values["provider"] = (
+                model_parts[-2] if len(model_parts) > 1 else model_parts[0]
+            )
 
         provider = values["provider"]
 
-        model_id_lower = values.get("base_model_id", values.get("base_model", model_id)).lower()
+        model_id_lower = values.get(
+            "base_model_id", values.get("base_model", model_id)
+        ).lower()
 
         # Determine if the model supports plain-text streaming (ConverseStream)
         # Here we check based on the updated AWS documentation.
@@ -532,7 +538,8 @@ class ChatBedrockConverse(BaseChatModel):
             (provider == "meta")
             or
             # All Mistral models
-            (provider == "mistral") or
+            (provider == "mistral")
+            or
             # DeepSeek-R1 models
             (provider == "deepseek" and "r1" in model_id_lower)
         ):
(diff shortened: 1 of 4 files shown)


libs/aws/langchain_aws/function_calling.py
```diff
@@ -63,6 +63,17 @@
...
+class CachePointType(TypedDict):
+    """Configuration for prompt caching in Bedrock toolConfig."""
+    
+    type: Literal["default"]
+
+
+class CachePoint(TypedDict):
+    """Configuration for prompt caching in Bedrock."""
+    
+    cachePoint: CachePointType
```
3coins, line 75: This might be an overkill for such a small input structure. 
```suggestion
class CachePointType(TypedDict):
    """Configuration for prompt caching in Bedrock toolConfig."""
    
    type: Literal["default"]


class CachePoint(TypedDict):
    """Configuration for prompt caching in Bedrock."""
    
    cachePoint: CachePointType
```

libs/aws/langchain_aws/function_calling.py
```diff
@@ -203,12 +214,23 @@ def _pydantic_parse(self, tool_call: ToolCall) -> BaseModel:
...
 
+def is_cache_point(cache_point: Any) -> bool:
+    return (
+        isinstance(cache_point, dict) 
+        and len(cache_point) == 1
+        and "cachePoint" in cache_point 
+        and isinstance(cache_point["cachePoint"], dict)
+        and len(cache_point["cachePoint"]) == 1
+        and "type" in cache_point["cachePoint"]
+    )
```
3coins, line 225: Can simplify this check. Also, can move this to bedrock_converse.py file, since this will not be used outside.

```suggestion
def is_cache_point(cache_point: Any) -> bool:
    return (
        isinstance(cache_point, dict) 
        and "cachePoint" in cache_point
        and cache_point.get("cachePoint").get("type") is not None
    )
```

libs/aws/langchain_aws/function_calling.py
```diff
@@ -203,12 +214,23 @@ def _pydantic_parse(self, tool_call: ToolCall) -> BaseModel:
...
 
 def convert_to_anthropic_tool(
     tool: Union[Dict[str, Any], TypeBaseModel, Callable, BaseTool],
-) -> AnthropicTool:
+) -> Union[AnthropicTool, CachePoint]:
```
3coins, line 229:
```suggestion
) -> Union[AnthropicTool, Dict]:
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -460,6 +460,16 @@ class Joke(BaseModel):
...
 
+    @classmethod
+    def create_cache_point(cls, cache_type: str = "default") -> Dict[str, Any]:
+        """Create a prompt caching configuration for Bedrock.
+        Args:
+            cache_type: Type of cache point. Default is "default".
+        Returns:
+            Dictionary containing prompt caching configuration.
+        """
+        return {"cachePoint": {"type": cache_type}}
```
3coins, line 471: This is great, can be used for system and message cache points as well. 
**nit**: Rather than a class or static method, should we just expose a function that users can directly call, since nothing in this function depends on the class.
```python
# Simpler to just use
create_cache_point(...)

# vs

BedrockConverse.create_cache_point(...)
```

3coins (commented):
@sharonxiaohanli 
Thanks for submitting this change. Added some suggestions to improve the code, the overall methodology to add cache_point looks great!

libs/aws/langchain_aws/chat_models/bedrock_converse.py
sharonxiaohanli: I thought about it but the syntax above only applies to Converse API. [InvokeModel API has a different syntax depending on Claude models vs. Nova Models](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-get-started). 

e.g. Claude models InvokeModel API
```
                    "cache_control": {
                        "type": "ephemeral"
                    }
```

In a sense, this method should only work with Converse API.

krshubham:
Hi, while using this I see that it might work well in the tests, but on a uber level with langchain agents due to validation of the tools it fails with:

```
   v_list = validator(input_value)
             ^^^^^^^^^^^^^^^^^^^^^^
TypeError: Can't instantiate abstract class BaseTool without an implementation for abstract method '_run'

```

PR #466, opened 2025-04-29
Added streaming support for writer palmyra models

edvinhallvaxhiu:
Added streaming support for writer palmyra models for ChatBedrockConverse.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -532,6 +532,9 @@ class ChatBedrockConverse(BaseChatModel):
             or
             # Cohere Command R models
             (provider == "cohere" and "command-r" in model_id_lower)
+            or
+            # Writer Palmyra models
+            (provider == "writer" and "palmyra" in model_id_lower)
         ):
             streaming_support = True
         elif (


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -532,6 +532,9 @@ def set_disable_streaming(cls, values: Dict) -> Any:
             or
             # Cohere Command R models
             (provider == "cohere" and "command-r" in model_id_lower)
+            or
+            # Writer Palmyra models
+            (provider == "writer" and "palmyra" in model_id_lower)
         ):
             streaming_support = True
```
michaelnchin, line 539: Converse API doesn't support streaming tool use for Writer Palymra models:

Let's move this to the conditional block that sets `streaming_support = "no_tools"`.

michaelnchin (changes requested):
@edvinhallvaxhiu, thank you for the submission!

libs/aws/langchain_aws/chat_models/bedrock_converse.py
edvinhallvaxhiu: @michaelnchin - thank you for the feedback! pull request updated.

michaelnchin (approved):
Tested and LGTM - thanks @edvinhallvaxhiu !

PR #473, opened 2025-05-08
aws[patch]: (ChatBedrockConverse) add response metadata in streaming case

ccurme:


michaelnchin (approved):
Thanks @ccurme !

Integration tests pass: https://github.com/langchain-ai/langchain-aws/actions/runs/14918228782/job/41908425887

PR #475, opened 2025-05-09
Update ChatBedrock for Llama 4 model support

michaelnchin:
Updating ChatBedrock to correctly format prompts for Llama 4 models, following the official documentation:

https://www.llama.com/docs/model-cards-and-prompt-formats/llama4/

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -112,12 +112,49 @@ def _convert_one_message_to_text_llama3(message: BaseMessage) -> str:
 
 
 def convert_messages_to_prompt_llama3(messages: List[BaseMessage]) -> str:
-    """Convert a list of messages to a prompt for llama."""
+    """Convert a list of messages to a prompt for Llama 3."""
 
     return "\n".join(
         ["<|begin_of_text|>"]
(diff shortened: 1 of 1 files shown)


libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -121,6 +121,43 @@ def convert_messages_to_prompt_llama3(messages: List[BaseMessage]) -> str:
     )
 
 
+def _convert_one_message_to_text_llama4(message: BaseMessage) -> str:
+    if isinstance(message, ChatMessage):
+        message_text = (
+            f"<|header_start|>{message.role}<|header_end|>{message.content}<|eot|>"
```
3coins, line 127: Are arbitrary roles supported?
https://www.llama.com/docs/model-cards-and-prompt-formats/llama4/#-llama-4---prompt-formats-

michaelnchin: That's a good point, I'm honestly not too sure. 

This is in line with the [llama3 ChatMessage translator](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock.py#L92) translator implementation, where the model also does not explicitly support arbitrary roles. But in practice, at worst, llama3 and llama4 seem to ignore the custom role.

PR #477, opened 2025-05-12
Correctly handle empty input in ChatBedrock's Anthropic prompt adapter

rotemruso:
Making convert_messages_to_prompt_anthropic work with messages or empty message list

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -158,9 +158,8 @@ def convert_messages_to_prompt_anthropic(
         return ""
     
     messages = messages.copy()  # don't mutate the original list
-    if messages:
-        if not isinstance(messages[-1], AIMessage):
-            messages.append(AIMessage(content=""))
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -154,10 +154,13 @@ def convert_messages_to_prompt_anthropic(
...
-
+    if messages is None:
+        return ""
+    
     messages = messages.copy()  # don't mutate the original list
-    if not isinstance(messages[-1], AIMessage):
-        messages.append(AIMessage(content=""))
+    if messages:
+        if not isinstance(messages[-1], AIMessage):
+            messages.append(AIMessage(content=""))
```
michaelnchin, line 163: Can simplify this:
```suggestion
    if len(messages) > 0 and not isinstance(messages[-1], AIMessage):
        messages.append(AIMessage(content=""))
```

michaelnchin (changes requested):

michaelnchin (approved):
Thanks @rotemruso !

PR #482, opened 2025-05-15
rerank: preserve Document-fields

georg-wind:
Document(BaseMedia) has at least one more property that is currently dropped: "id". Could also be added explicitly.


michaelnchin (approved):
Thanks @georg-wind ! Updated the unit tests.

PR #496, opened 2025-05-27
aws[patch]: fix tool calling and structured output for Claude 4 / ChatBedrock

ccurme:
Tests pass for ChatBedrockConverse

### DIFF
--- a/libs/aws/tests/integration_tests/chat_models/test_standard.py
+++ b/libs/aws/tests/integration_tests/chat_models/test_standard.py
@@ -16,7 +16,7 @@ class TestBedrockStandard(ChatModelIntegrationTests):
 
     @property
     def chat_model_params(self) -> dict:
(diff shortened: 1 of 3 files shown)


libs/aws/tests/integration_tests/chat_models/test_standard.py
```diff
@@ -16,7 +16,7 @@ def chat_model_class(self) -> Type[BaseChatModel]:
 
     @property
     def chat_model_params(self) -> dict:
-        return {"model_id": "anthropic.claude-3-sonnet-20240229-v1:0"}
+        return {"model_id": "us.anthropic.claude-sonnet-4-20250514-v1:0"}
```
ccurme, line 19: Don't necessarily need to test Claude 4 here but seemed preferable to Claude 3.0.

PR #513, opened 2025-06-13
Change to be able to pass stream param during ainvoke

aagarwal-amit:
I want to be able to control whether to stream or not using ainvoke method. 
However, current code gives following error - 

ChatBedrockConverse._converse_params() got an unexpected keyword argument 'stream'

Hence, adding a check to ignore stream param.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -652,7 +652,7 @@ class ChatBedrockConverse(BaseChatModel):
         params = self._converse_params(
             stop=stop,
             **_snake_to_camel_keys(
-                kwargs, excluded_keys={"inputSchema", "properties", "thinking"}
+                kwargs, excluded_keys={"inputSchema", "properties", "thinking", "stream"}
             ),
         )
         logger.debug(f"Input params: {params}")
@@ -676,7 +676,7 @@ class ChatBedrockConverse(BaseChatModel):
         params = self._converse_params(
             stop=stop,
             **_snake_to_camel_keys(
-                kwargs, excluded_keys={"inputSchema", "properties", "thinking"}
+                kwargs, excluded_keys={"inputSchema", "properties", "thinking", "stream"}
             ),
         )
         response = self.client.converse_stream(


michaelnchin (commented):
Hi @aagarwal-amit , thank you for submitting this PR.

Just for some context, could you provide a small code example to reproduce the issue that this aims to address?

aagarwal-amit:
> Hi @aagarwal-amit , thank you for submitting this PR.
> 
> Just for some context on your use case, could you provide a small code example to reproduce the issue that this aims to address?

Sure, this is the code. Here stream parameter is not supported and gives the exception.
```
llm_model = ChatBedrockConverse(
            region_name=config.get("REGION"),
            model=config.get("DEPLOYMENT_NAME"),
            base_url=config.get("BASE_URL", None),
            provider=config.get("PROVIDER", ""),
            aws_access_key_id=key["AWS_ACCESS_KEY_ID"],
            aws_secret_access_key=key["AWS_SECRET_ACCESS_KEY"],
        )
response = await llm_model.ainvoke(messages, **stream=False**)
```

michaelnchin:
Sorry - I'm still a bit unclear on the usage of stream param with `ainvoke`. Are you using streaming elsewhere in your application, and running into LangChain's [auto-streaming](https://python.langchain.com/docs/concepts/streaming/#auto-streaming-chat-models) on the invoke call?

Also, you can disable streaming on the ChatBedrockConverse model by passing `disable_streaming=True` - would this work for your case?

aagarwal-amit:
> Sorry - I'm still a bit unclear on the usage of stream param with `ainvoke`. Are you using streaming elsewhere in your application, and running into LangChain's [auto-streaming](https://python.langchain.com/docs/concepts/streaming/#auto-streaming-chat-models) on the invoke call?
> 
> Also, you can disable streaming on the ChatBedrockConverse model by passing `disable_streaming=True` - would this work for your case?

Yes, I am using `ainvoke` to control whether I want the streaming response or not. This is basically useful when I have a sub agent which is used to return a partial result vs when it is used to give final result. So I can decide at runtime, that if it is expected to return partial response, then I dont stream, but if its final response, then it streams (which gets streamed to UI directly by reading the langgraph's stream updates).

michaelnchin (approved):
Thanks @aagarwal-amit !

PR #516, opened 2025-06-16
aws[patch]: (Converse) support camel case tool call args

ccurme:


ccurme:
Integration tests pass: https://github.com/langchain-ai/langchain-aws/actions/runs/15690046073

Issue #519, opened 2025-06-17
Cohere provider not recognized by BedrockEmbeddings when using inference_profile (model_arn)


julienfoenet:
In the BedrockEmbedding class ([here](https://github.com/langchain-ai/langchain-aws/blob/644f22c144c90fcf98197d1a9059795cf139b723/libs/aws/langchain_aws/embeddings/bedrock.py#L127)) there is a parsing done on the `model_id` given as input, in order to define the provider's name. For example, if your model_id is `cohere.embed-multilingual-v3`, the then propetry `provider` will be set as `cohere`. In the rest of the code, this is used as switch, for example when formatting the input used to invoke the model (with cohere models, inputs are formatted in a special way).

The issue is that when using Bedrock inference_profiles, you don't use anymore a **model id** as input to BedrockEmbedding class, but a **model arn** instead, which does not start with the provider name (ex: `arn:aws:bedrock:eu-central-1:0123456789:application-inference-profile/ey3gaah56kd`). In that case `cohere` provider is not recognized and the formatting of input to invoke method does not work.
Any of the two model identifier, **model id** or **model arn** can be passed to Bedrock invoke method as `modelId` input argument (see [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model.html#)).

To fix that, maybe we should be able to pass directly the provider as input to the class `BedrockEmbedding`, and set the property `provider` as either this input or the existing parsing.

julienfoenet:
PR linked to this issue:
https://github.com/langchain-ai/langchain-aws/pull/522

PR #522, opened 2025-06-20
add provider input arg to BedrockEmbeddings class

julienfoenet:
This PR solves the issue https://github.com/langchain-ai/langchain-aws/issues/519

### DIFF
--- a/libs/aws/langchain_aws/embeddings/bedrock.py
+++ b/libs/aws/langchain_aws/embeddings/bedrock.py
@@ -106,9 +106,9 @@ class BedrockEmbeddings(BaseModel, Embeddings):
     model_kwargs: Optional[Dict] = None
     """Keyword arguments to pass to the model."""
 
-    provider_name: Optional[str] = None
+    provider: Optional[str] = None
     """Name of the provider, e.g., amazon, cohere, etc..
-    If not specified, the provider property will be inferred from the model_id."""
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/embeddings/bedrock.py
```diff
@@ -107,6 +106,10 @@ class BedrockEmbeddings(BaseModel, Embeddings):
     model_kwargs: Optional[Dict] = None
     """Keyword arguments to pass to the model."""
 
+    provider_name: Optional[str] = None
```
michaelnchin: Can we change this to just `provider`? This would be more consistent with the equivalent options for [`ChatBedrockConverse`](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L403) and [`BedrockLLM`](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/llms/bedrock.py#L649)/`ChatBedrock`.

julienfoenet: Ok then I'll have to rename the property method otherwise there will be a conflict between the property and the attribute with same name "provider". Let's rename the property `_inferred_provider` if it's ok for you.

michaelnchin (approved):
Thanks @julienfoenet !

PR #527, opened 2025-06-26
Fix the last commit as it was not removing the stream parameter.

aagarwal-amit:
Misunderstood the exclude_keys part as its not removing the stream key from kwargs. Hence adding stream as parameter to _converse_params function to pass the pydantic check.
Now, tested it and works as expected.


aagarwal-amit:
@michaelnchin

michaelnchin (approved):
Thanks @aagarwal-amit !

Issue #552, opened 2025-07-24
NeptuneGraph class does not support parameterized queries


bechbd:
The NeptuneGraph class does not currently pass through the params argument to the underlying query, preventing it from running parameterized queries.

PR #553, opened 2025-07-24
Fixed Issue #552 - Parameterized Query Support in NDB

bechbd:
Fixed Issue #552 by adding proper passing of the parameters through into the query call when using Neptune database


michaelnchin (approved):
Thank you @bechbd !

PR #556, opened 2025-07-28
Support prompt caching on system messages

aminah-ankoush:
Prompt caching was initially added [here](https://github.com/langchain-ai/langchain-aws/commit/0067b861a47af295683e101c98f9f6719f514728#diff-78c2abdea066a69391d4a156ea3d401a6fec2cb8ea8f578e2d42e7888a615d7c) but seems like it wasn't supported on system messages at that time. It is now!

https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#large-context-caching-example

This PR preserves `SystemMessage` content as a list of blocks rather than converting it into a single concatenated string.  This will make sure the caching param is retained.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -379,9 +379,9 @@ def _merge_messages(
 
 def _format_anthropic_messages(
     messages: List[BaseMessage],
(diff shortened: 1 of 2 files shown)


michaelnchin (approved):
Thanks @aminah-ankoush !

Issue #535, opened 2025-07-06
Bug: Bedrock Application Inference ARN breaks Claude Sonnet tools usage (regional inference works)


edibusl:
# 🐞 Bug: Bedrock Application Inference ARN breaks Claude Sonnet tools usage (regional inference works)

## Issue description

When running a LangChain agent using **tools** with **Bedrock Claude Sonnet 3.7**, the behavior differs between regional inference and application inference profiles.

- With **regional inference** (`us.anthropic.claude-3-7-sonnet-20250219-v1:0`), tools are passed properly and the agent calls tools as expected.
- With **application inference ARN** (`arn:aws:bedrock:us-east-1:111111484058:application-inference-profile/c3myu2h6fllr`), the LLM does not use the tools properly - tools are not passed and the output ignores them.

This makes it impossible to use tools with Claude when specifying an application inference ARN.

I confirmed with an **AWS Bedrock specialist** that the Application Inference Profile has exactly the same API as the regional inference and supports all features, including the tools API. Therefore, this appears to be a bug in how `langchain-aws` detects and handles the model name when using an ARN.

## Investigation

In [`bedrock.py`](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock.py#L1003), the relevant logic is here:

```python
if "claude-" in self._get_base_model():
```

When using a regional inference name (`us.anthropic.claude-3-7-sonnet-20250219-v1:0`), this condition passes, so tools are serialized properly using the Anthropic tools JSON API.  

When using an application inference ARN (`arn:aws:bedrock:...`), the ARN string does *not* contain `"claude-"`, so the condition fails, and tools are not passed correctly - which results in the LLM ignoring them.

In my older local version of the package (about 1 month old), the check was even stricter:

```python
if "claude-3" in self._get_base_model():
```

So it seems this code is actively evolving - but the issue remains: using the ARN prevents tools from being passed properly.

### Regional inference (works):

```python
model = ChatBedrock(
    model="anthropic.claude-3-7-sonnet-20250219-v1:0",
    ...
)
agent = model.bind_tools(...)
# tools invoked as expected
```

### Application ARN (broken):

```python
model = ChatBedrock(
    model="arn:aws:bedrock:us-east-1:111111484058:application-inference-profile/c3myu2h6fllr",
    ...
)
agent = model.bind_tools(...)
# tools ignored
```

## Expected behavior

[...]

PR #561, opened 2025-08-01
Fix: Bedrock Application Inference ARN breaks Claude Sonnet tools usage

0x-fang:
### Description

This PR fixes the issue when running a agent using tools with application inference profile of Claude Sonnet 3.7 as input. Note that using regional inference profile is working well, and only application inference profile (AIP) will not return any tools use information. 

### Root Cause

Langchain needs to identify the model id of the foundation model but fails to detect the foundation model from the input AIP ARN (e.g., `arn:aws:bedrock:us-east-1:111111484058:application-inference-profile/c3myu2h6fllr`), while regional inference profile contains the foundation model name `us.anthropic.claude-3-7-sonnet-20250219-v1:0` and can be easily used to identify the base model id. Therefore, regional inference profile is working well while AIP is not. 

### Solution

To identify the foundation model used in AIP, we need to call  Bedrock `get_inference_profile` control plane [API](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock/client/get_inference_profile.html), and parse the model id from the response. 

### Issue 

https://github.com/langchain-ai/langchain-aws/issues/535

### DIFF
--- a/libs/aws/langchain_aws/llms/bedrock.py
+++ b/libs/aws/langchain_aws/llms/bedrock.py
@@ -591,6 +591,10 @@ class BedrockBase(BaseLanguageModel, ABC):
     """Base class for Bedrock models."""
 
     client: Any = Field(default=None, exclude=True)  #: :meta private:
+    """The bedrock runtime client for making data plane API calls"""
+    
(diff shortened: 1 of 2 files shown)


michaelnchin (approved):
Thanks @0x-fang !

Issue #538, opened 2025-07-10
Application Inference Profile (AIP) is not streaming response with ChatBedrockConverse


Aruk0903:
When using an Application Inference Profile (AIP) with LangChain's ChatBedrockConverse, the response is not being streamed incrementally as expected. Instead, the entire response is delivered as a single chunk.

However CRIS is streaming response in multiple chunks.

On checking the logs, It seems Application Inference Profile (AIP) is calling Converse API in the backend. While CRIS is calling ConvereStream. This explains the scenario.

0x-fang:
looking into this issue

0x-fang:
bugfix for this issue https://github.com/langchain-ai/langchain-aws/pull/568

PR #568, opened 2025-08-05
Fix: Bedrock Application Inference Profile (AIP) is not streaming response with ChatBedrockConverse

0x-fang:
### Description

This PR fixed the streaming issue with ChatBedrockConverse when input is Application Inference Profile (AIP). The issue happened because langchain-aws fails to identify the foundation model used in AIP (e.g., `arn:aws:bedrock:us-east-1:111111484058:application-inference-profile/c3myu2h6fllr`), therefore, it cannot set the streaming_support flag for the AIP correctly. So the entire response was returned to user as a whole rather than streaming the response back. 

### Solution

We have to create a Bedrock client to call Bedrock get_inference_profile control plane [API](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock/client/get_inference_profile.html), and parse the foundation model id from the response. However, the existing `set_disable_streaming` function works on the raw user input before the ChatBedrockConverse object is instantiated, so we have to extract the logic of determining streaming_support for models into a common function `_get_streaming_support` and invoke it in both places (i.e., the original place that works on raw user input, and the new place that works on the resolved model Id from get_inference_profile API call).

### Issue 

https://github.com/langchain-ai/langchain-aws/issues/538

### Test

* add new unit tests in libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py and all passed
* run through the integration test for ChatBedrockConverse and all passed except one existing failure
   *  The main branch has failed this integration test already, which is not caused by new change. 
   ```
   FAILED tests/integration_tests/chat_models/test_bedrock_converse.py::test_structured_output_tool_choice_not_supported - assert 1 == 0
   ```

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -312,6 +312,10 @@ class ChatBedrockConverse(BaseChatModel):
     """  # noqa: E501
 
     client: Any = Field(default=None, exclude=True)  #: :meta private:
+    """The bedrock runtime client for making data plane API calls"""
+
+    bedrock_client: Any = Field(default=None, exclude=True)  #: :meta private:
+    """The bedrock client for making control plane API calls"""
 
     model_id: str = Field(alias="model")
     """Id of the model to call.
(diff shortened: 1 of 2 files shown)


michaelnchin:
Integ CI passes: https://github.com/langchain-ai/langchain-aws/actions/runs/16761621041

michaelnchin (approved):
Thanks @0x-fang !

PR #579, opened 2025-08-14
fix: Multimodal type conversion to Converse API format

### LINKED ISSUE #574: ConverseAPI format incorrectly constructed from MimeType
In [chat_model/bedrock_converse.py](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py), you define a function to go from langchain message format to Converse API format. In particular when dealing with multimodal request, there is the need to determine the format of the image or file. For this you are doing:

`"format": block["mimeType"].split("/")[1]`

(example [here](https://github.com/langchain-ai/langchain-aws/blob/b51aae07cf520b5b5bf4e0fa0adb50d131ea3913/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L1227C1-L1228C1))

Now here is the problem, the list of MimeTypes can be found here: 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types
https://www.iana.org/assignments/media-types/media-types.xhtml

**For markdown:**
Correct mime type:`text/markdown`
Resulting `"format": "markdown"`
Expected format: `md`

**For xslx:**
Correct mime type:`application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`
Resulting `"format": "vnd.openxmlformats-officedocument.spreadsheetml.sheet"`
Expected format: `xslx`

Same for docx, xls,...

The expected format from [Converse API](https://botocore.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html) must be in the following enums:
For image: `'png'|'jpeg'|'gif'|'webp'`
For documents: `'pdf'|'csv'|'doc'|'docx'|'xls'|'xlsx'|'html'|'txt'|'md'`

It's lucky it's working for pdf, images as the mime types as the mime types are application/pdf, image/png, etc...

michaelnchin:
Fixes #574.

This PR updates the LangChain -> Bedrock message conversion utility in ChatBedrockConverse to correctly convert multimodal input MIME types to Converse API format (where supported).


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -1200,6 +1200,48 @@ def _parse_stream_event(event: Dict[str, Any]) -> Optional[BaseMessageChunk]:
...
+        # Video formats
+        "video/x-matroska": "mkv",
+        "video/quicktime": "mov",
+        "video/mp4": "mp4",
+        "video/webm": "webm",
+        "video/x-flv": "flv",
+        "video/mpeg": "mpeg",
+        "video/x-ms-wmv": "wmv",
+        "video/3gpp": "three_gp",
+    }
```
3coins: minor: Can this be moved to a module level constant?

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -1200,6 +1200,48 @@ def _parse_stream_event(event: Dict[str, Any]) -> Optional[BaseMessageChunk]:
         raise ValueError(f"Received unsupported stream event:\n\n{event}")
 
 
+def _mime_type_to_format(mime_type: str) -> str:
```
3coins, line 1232: To optimize set creation and splits at each call, consider using a cache.
```python
@functools.cache
def _mime_type_to_format(mime_type: str) -> str:
```

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
```diff
@@ -1107,6 +1107,70 @@ def test__lc_content_to_bedrock_reasoning_content_signature() -> None:
     assert expected_system == actual_system
 
 
+def test__lc_content_to_bedrock_mime_types() -> None:
```
3coins, line 1110: nit: Some failure cases are not tested here.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -1200,6 +1200,48 @@ def _parse_stream_event(event: Dict[str, Any]) -> Optional[BaseMessageChunk]:
...
+        "video/x-ms-wmv": "wmv",
+        "video/3gpp": "three_gp",
+    }
+    
+    if mime_type in mime_to_format:
+        return mime_to_format[mime_type]
+    
+    # Fallback to original method of splitting on "/" for simple cases
+    all_formats = set(mime_to_format.values())
+    format_part = mime_type.split("/")[1]
```
3coins, line 1243: This will fail without meaningful error, if input does not have `/` in it. Some simple input validation can avoid this.
```python
if "/" not in mime_type:
  raise ValueError(
      f"Invalid MIME type format: {mime_type}. Expected format: 'type/subtype'"
  )
```

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
michaelnchin: Added a new `test__lc_content_to_bedrock_mime_types_invalid` case, includes the below 👍

libs/aws/langchain_aws/chat_models/bedrock_converse.py
michaelnchin: Good catch - thanks!

Issue #544, opened 2025-07-18
Cannot use streaming with Amazon Nova models through ChatBedrock when tools are bound


Freybsnapsoft:
When tools are bound to Amazon Nova models through `ChatBedrock` class or `bedrock:` provider prefix, they return responses in one chunk when using `.stream`.
Using `ChatBedrockConverse` class or `bedrock_converse:` provider prefix works correctly.

Example code to reproduce:
:x: Returns response in one chunk
```
from langchain_aws import ChatBedrock
from langchain_core.tools import tool

@tool
def get_weather(city: str) -> str:
    """Get weather for a given city."""
    return f"It's always sunny in {city}!"
    
llm = ChatBedrock(
    model="us.amazon.nova-pro-v1:0",
).bind_tools([get_weather])

for message_chunk in llm.stream( 
    [
        {"role": "user", "content": f"How's the weather in San Francisco"}
    ],
):
    if message_chunk.content:
        print(message_chunk.content, flush=True)
```
:white_check_mark: Streams response
```
from langchain_aws import ChatBedrockConverse
from langchain_core.tools import tool

@tool
def get_weather(city: str) -> str:
    """Get weather for a given city."""
    return f"It's always sunny in {city}!"
    
llm = ChatBedrockConverse(
    model="us.amazon.nova-pro-v1:0",
).bind_tools([get_weather])

for message_chunk in llm.stream( 
    [
        {"role": "user", "content": f"How's the weather in San Francisco"}
    ],
):
    if message_chunk.content:
        print(message_chunk.content, flush=True)
```

The root cause of the problem is that when using Amazon Nova models through ChatBedrock or `bedrock:` provider prefix, LangChain forces them to use the Converse API.

https://github.com/langchain-ai/langchain-aws/blob/d97e1b3493369261eaa45ca7d37117784c67b659/libs/aws/langchain_aws/chat_models/bedrock.py#L695-L702

When calling stream on the model, the `_as_converse` creates a new `ChatBedrockConverse` class and sets the provider to an empty string - since `self.provider` is `None`.

https://github.com/langchain-ai/langchain-aws/blob/d97e1b3493369261eaa45ca7d37117784c67b659/libs/aws/langchain_aws/chat_models/bedrock.py#L1198-L1211

This makes `ChatBedrockConverse.set_disable_streaming` set the `disable_streaming` property to False.
[...]

PR #580, opened 2025-08-14
Fix: fix the streaming issue for nova models

0x-fang:
### Description

This PR fixes the bug that cause streaming is broken for Amazon Nova model. The response was returned as a whole rather than streaming back chunks to customers, even though `.stream()` is used. 

### Root Cause and Fix

First of all, when using Amazon Nova model, langchain-aws will use ChatBedrockConverse even though we create ChatBedrock ([code](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock.py#L704)). Inside ChatBedrockConverse, the input parameter `provider` is default to empty string rather than None ([code](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L407)), therefore, the [set_disable_streaming](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L582) function will use `provider=""` to disable the streaming, even though the model can support streaming. To fix it, this PR updates **set_disable_streaming** function to extract provider from model Id correctly when `provider` is empty string as default value, so it can  setup the streaming flag correctly. 

Alternatively, user can set the `provider` parameter explicitly when initializing the ChatBedrock as below, which can make streaming work for Nova models without any code change.
```
llm = ChatBedrock(
    provider="amazon",
    model="us.amazon.nova-pro-v1:0",
)

for chunk in llm.stream(prompt):
    print(chunk)
```

### Issue

https://github.com/langchain-ai/langchain-aws/issues/544

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -584,7 +584,7 @@ class ChatBedrockConverse(BaseChatModel):
         
         # Extract provider from the model_id
         # (e.g., "amazon", "anthropic", "ai21", "meta", "mistral")
-        if "provider" not in values:
+        if "provider" not in values or values["provider"] == "":
             if model_id.startswith("arn"):
                 raise ValueError(
                     "Model provider should be supplied when passing a model ARN as model_id."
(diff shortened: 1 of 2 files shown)


michaelnchin (approved):
Thank you @0x-fang !

PR #581, opened 2025-08-16
Support OpenAI gpt-oss models on ChatBedrock

michaelnchin:
Adds base support for using [gpt-oss](https://openai.com/index/introducing-gpt-oss/) on ChatBedrock. Tools and streaming support to be implemented in later PRs.

Note that Bedrock currently uses the [Chat Completions API](https://platform.openai.com/docs/api-reference/chat) for gpt-oss models, and not the newer [Responses API](https://platform.openai.com/docs/api-reference/responses), so we have to format the input (i.e. messages only) and parameters in the request body accordingly.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -35,6 +35,7 @@ from langchain_core.messages import (
 )
 from langchain_core.messages.ai import UsageMetadata
 from langchain_core.messages.tool import ToolCall, ToolMessage
+from langchain_core.messages.utils import convert_to_openai_messages
 from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
 from langchain_core.runnables import Runnable, RunnableMap, RunnablePassthrough
 from langchain_core.tools import BaseTool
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -285,6 +286,48 @@ def convert_messages_to_prompt_writer(messages: List[BaseMessage]) -> str:
...
+        message_text = (
+            f"<|start|>user<|message|>{message.content}<|end|>"
+        )
+    elif isinstance(message, AIMessage):
+        message_text = (
+            f"<|start|>assistant<|channel|>final<|message|>{message.content}<|end|>"
+        )
+    elif isinstance(message, ToolMessage):
+        # TODO: should start with "<|start|>{toolname} to=assistant."
+        #  How do we get toolname from the ToolMessage content?
```
3coins:
```suggestion
      # TODO: Tool messages in the OpenAI format should use "<|start|>{toolname} to=assistant<|message|>"
      # Need to extract the tool name from the ToolMessage content or tool_call_id
      # For now using generic "to=assistant" format as placeholder until we implement tool calling
      # Will be resolved in follow-up PR with full tool support
```

Issue #411, opened 2025-03-17
botocore.errorfactory.ValidationException with Claude 3.5 models and below [Langchain-aws 0.2.15]


sibijr:
After upgrading to 0.2.15, I started seeing the below errors while working with Claude 3.5 or below models. The same code works fine with 3.7.

```
    | botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the InvokeModelWithResponseStream operation: Malformed input request: #: subject must not be valid against schema {"required":["messages"]}#/messages/1/content: expected type: String, found: JSONArray#/messages/1/content: expected minimum item count: 1, found: 0, please reformat your input and try again.
```

Any help?

3coins:
@sibijr 
Can you provide a small code sample to reproduce. Thanks.

sibijr:
Below the code sample to reproduce, It works with 3.7 but failed with 3.5. Please check


```

import json
from typing import Any, Dict, List, Optional, Tuple

import botocore
from langchain.agents import AgentExecutor
from langchain.agents.format_scratchpad import format_xml
from langchain_aws import ChatBedrock
from langchain_core.agents import AgentFinish
from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, SystemMessage
from langchain_core.prompts import ChatPromptTemplate, MessagesPlaceholder
from langchain_core.runnables import RunnableLambda
from pydantic import BaseModel, Field

from app.utils.sanitizer_util import clean_backtick_sequences

# model_id = "us.anthropic.claude-3-7-sonnet-20250219-v1:0"
model_id = "us.anthropic.claude-3-5-sonnet-20240620-v1:0"


template = """
Below is the current codebase of the user: 

---------------------------------------

{codebase}

---------------------------------------
Codebase ends here. 

"""

conversational_prompt = ChatPromptTemplate.from_messages(
    [
        ("system", template),
        MessagesPlaceholder(variable_name="chat_history", optional=True),
        ("user", "{question}"),
        ("ai", "{agent_scratchpad}"),
    ]
)

def get_model_with_thinking(thinking_enabled=False):
    """
    Create a ChatBedrock model with thinking enabled or disabled based on UI toggle
    and model support
    """
    # Basic parameters for both model versions
    model_kwargs={"max_tokens": 8192, "temperature": 0.3, "top_k": 15}

    # Only add thinking parameter when explicitly enabled in UI AND using supported model
[...]

Issue #403, opened 2025-03-13
ChatBedrockConverse ValidationException on Non-Empty AI Message


shahrukh-shaik:
Sample Code:

```python
from langchain_aws import ChatBedrockConverse
model_id="us.anthropic.claude-3-5-haiku-20241022-v1:0"
llm1 = ChatBedrockConverse(model_id=model_id,disable_streaming=True)

llm1.invoke([HumanMessage(content=[{'type': 'text', 'text': 'For the earlier part, can you see if there are any updates'}], additional_kwargs={}, response_metadata={}, id='dbda63b2-4ed8-4d7c-ab51-219f82353cb6'),
 AIMessage(content=[{'type': 'text', 'text': '\n\n', 'index': 0}], additional_kwargs={}),
 HumanMessage(content="try again")])
```

Original Conversation 
```python
[HumanMessage(content=[{'type': 'text', 'text': 'For the earlier part, can you see if there are any updates'}], additional_kwargs={}, response_metadata={}, id='dbda63b2-4ed8-4d7c-ab51-219f82353cb6'),
 AIMessage(content=[{'type': 'text', 'text': "I'd be happy to check for any updates regarding your earlier query. Let me search for the most current information.", 'index': 0}], additional_kwargs={}, response_metadata={'stopReason': 'tool_use', 'metrics': {'latencyMs': 4917}, 'llm': 'us.anthropic.claude-3-7-sonnet-20250219-v1:0'}, name='general', id='run-12a3ef27-eaa6-4ea8-857e-1031b066560c', tool_calls=[{'name': 'get_current_date', 'args': {}, 'id': 'tooluse_8Ke4pZdcS2C9DklgqMNFvw', 'type': 'tool_call'}], usage_metadata={'input_tokens': 39897, 'output_tokens': 77, 'total_tokens': 39974}),
 ToolMessage(content='2025-03-13', name='get_current_date', id='006b1fc4-0162-42ae-a778-1e2eb9767e2b', tool_call_id='tooluse_8Ke4pZdcS2C9DklgqMNFvw'),
[...]

PR #590, opened 2025-08-22
fix: Avoid ValidationException for empty message content passed to ChatBedrock

BumaldaOverTheWater94:
https://github.com/langchain-ai/langchain-aws/issues/411

### DIFF
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
@@ -1034,7 +1034,7 @@ def test__format_anthropic_messages_empty_content_fix() -> None:
     assert isinstance(ai_content, list)
     assert len(ai_content) > 0  # Should not be empty
     assert ai_content[0]["type"] == "text"
-    assert ai_content[0]["text"] == "[empty]"
+    assert ai_content[0]["text"] == "."
 
 
 def test__format_anthropic_messages_whitespace_only_content() -> None:
@@ -1051,7 +1051,7 @@ def test__format_anthropic_messages_whitespace_only_content() -> None:
     assert isinstance(ai_content, list)
     assert len(ai_content) > 0
     assert ai_content[0]["type"] == "text"
-    assert ai_content[0]["text"] == "[empty]"
+    assert ai_content[0]["text"] == "."
 
 
 def test__format_anthropic_messages_empty_string_content() -> None:
@@ -1068,7 +1068,7 @@ def test__format_anthropic_messages_empty_string_content() -> None:
     assert isinstance(ai_content, list)
     assert len(ai_content) > 0
     assert ai_content[0]["type"] == "text"
-    assert ai_content[0]["text"] == "[empty]"
+    assert ai_content[0]["text"] == "."
 
 
 def test__format_anthropic_messages_mixed_empty_content() -> None:
@@ -1090,7 +1090,7 @@ def test__format_anthropic_messages_mixed_empty_content() -> None:
     assert isinstance(ai_content, list)
     assert len(ai_content) > 0
     assert ai_content[0]["type"] == "text"
-    assert ai_content[0]["text"] == "[empty]"
+    assert ai_content[0]["text"] == "."
 
 
 def test__format_anthropic_messages_mixed_type_blocks_and_empty_content() -> None:


blanketspy99:
I hope the #403 issue has been reviewed. Bedrock with Anthropic throws ValidationException when content text is escapable character like `\n` `\t` `\n\t` or more

michaelnchin:
@blanketspy99 thanks for the callout, that issue is still also present with `ChatBedrock` in this PR. Even with the added message content object to fix the original error, Invoke API still complains about the empty content value:
```
ValidationException: An error occurred (ValidationException) when calling the InvokeModelWithResponseStream operation: messages: text content blocks must contain non-whitespace text
```
TBH this is really a Bedrock side issue - Invoke/Converse API should preferably handle these cases gracefully (or accept them outright) instead of terminating immediately, but for now, it looks like input needs to be validated/updated manually by the user.

michaelnchin:
For now, I'm going to set any empty/whitespace-only input message content to a placeholder string `[empty]` so that Invoke API doesn't throw an error. 

This fix seems to work for messages to ChatBedrockConverse as well, though I'm not sure if it would have an effect on model inference results. @blanketspy99 let me know what you think.

shahrukh-shaik:
```python

def is_empty_response(text):
    """
    Uses a regex pattern to identify responses that are effectively empty.
    
    This checks for:
    - Completely empty strings
    - Strings with only whitespace
    - Strings with only escape sequences and their combinations
    - Strings with only newlines, tabs and spaces
    """
    if not text:
        return True
        
    # Pattern to match strings that ONLY contain:
    # - whitespace (\s)
    # - escaped characters like \n, \t, \r (either as actual escapes or as literals)
    pattern = r'^[\s\\ntr]*$'
    
    return bool(re.match(pattern, text))

def extract_message_content_by_types(content:list[dict[str, Any]],types:list=['text']):
    if isinstance(content, str):
        return content
    try:
        result = []
        if content == []:
            content = [{'type': 'text', 'text': "", 'index': 0}]

        

        for item in content:
            # Check if the dictionary has any key that's in filter_keys
            if item['type'] in types:
                if item.get('type',None) == 'text':
                    if is_empty_response(item.get('text')):
                        item['text']="""."""
                        # print("empty response")
                # print(item)
                result.append(item)
        return result
    except Exception as e:
        return content
```

blanketspy99:
> ```python
> 
> def is_empty_response(text):
>     """
>     Uses a regex pattern to identify responses that are effectively empty.
>     
>     This checks for:
>     - Completely empty strings
>     - Strings with only whitespace
>     - Strings with only escape sequences and their combinations
>     - Strings with only newlines, tabs and spaces
>     """
>     if not text:
>         return True
>         
>     # Pattern to match strings that ONLY contain:
>     # - whitespace (\s)
>     # - escaped characters like \n, \t, \r (either as actual escapes or as literals)
>     pattern = r'^[\s\\ntr]*$'
>     
>     return bool(re.match(pattern, text))
> 
> def extract_message_content_by_types(content:list[dict[str, Any]],types:list=['text']):
>     if isinstance(content, str):
>         return content
>     try:
>         result = []
>         if content == []:
>             content = [{'type': 'text', 'text': "", 'index': 0}]
> 
>         
> 
>         for item in content:
>             # Check if the dictionary has any key that's in filter_keys
>             if item['type'] in types:
>                 if item.get('type',None) == 'text':
>                     if is_empty_response(item.get('text')):
>                         item['text']="""."""
>                         # print("empty response")
>                 # print(item)
>                 result.append(item)
>         return result
>     except Exception as e:
>         return content
> ```

@michaelnchin - this is how I'm currently handling the empty content.

As per the PR I see it only checks when content is empty list but there are cases where content is not present or empty in the AIMessage when it's making tool call (tool_use)

blanketspy99:
> > ```python
> > 
> > def is_empty_response(text):
> >     """
> >     Uses a regex pattern to identify responses that are effectively empty.
> >     
> >     This checks for:
> >     - Completely empty strings
> >     - Strings with only whitespace
> >     - Strings with only escape sequences and their combinations
> >     - Strings with only newlines, tabs and spaces
> >     """
> >     if not text:
> >         return True
> >         
> >     # Pattern to match strings that ONLY contain:
> >     # - whitespace (\s)
> >     # - escaped characters like \n, \t, \r (either as actual escapes or as literals)
> >     pattern = r'^[\s\\ntr]*$'
> >     
> >     return bool(re.match(pattern, text))
> > 
> > def extract_message_content_by_types(content:list[dict[str, Any]],types:list=['text']):
> >     if isinstance(content, str):
> >         return content
> >     try:
> >         result = []
> >         if content == []:
> >             content = [{'type': 'text', 'text': "", 'index': 0}]
> > 
> >         
> > 
> >         for item in content:
> >             # Check if the dictionary has any key that's in filter_keys
> >             if item['type'] in types:
> >                 if item.get('type',None) == 'text':
> >                     if is_empty_response(item.get('text')):
> >                         item['text']="""."""
> >                         # print("empty response")
> >                 # print(item)
> >                 result.append(item)
> >         return result
> >     except Exception as e:
> >         return content
> > ```
> 
> @michaelnchin - this is how I'm currently handling the empty content.
> 
> As per the PR I see it only checks when content is empty list but there are cases where content is not present or empty in the AIMessage when it's making tool call (tool_use)
> 
> 

I meant,

If the AIMessage(content=[{type: text, text: \n\n},{type: tool_use, tool_use: xxxxxxx}]). We could have this case where there are other content type in AIMessage.

And, I guess we still have to replace the `[empty]` with some preferable placeholder at the client side to keep the conversation history tidy.

michaelnchin:
The check that we added in this PR is just to compensate for the case where empty text type blocks are the only blocks present in a message. This is the only scenario where `[empty]` will be seen. 

ChatBedrock by default removes empty/whitespace text blocks from message content, so if there is an extra `tool_use` block, the tool block will remain untouched: 
https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock.py#L544-L554

Will add another test to ensure we're validating this case.

PR #592, opened 2025-08-26
fix: Set bedrock_client region name to runtime client value when needed

### LINKED ISSUE #586: Failed to Load AWS Credentials for Authentication in Virtual Environment
When upgrading to `langchain-aws==0.2.31`, an error occurs during AWS client authentication. The issue is specific to virtual environments and does not appear when using version `0.2.30`. The error message indicates that credentials could not be loaded to authenticate with the AWS client, and a region must be specified.

### Error details
```
1 validation error for ChatBedrock Value error, Could not load credentials to authenticate with AWS client. 
Please check that the specified profile name and/or its credentials are valid. 
Service error: You must specify a region. [type=value_error, input_value={'client': <botocore.clie...se_converse_api': False}, input_type=dict]
For further information visit [https://errors.pydantic.dev/2.11/v/value_error]
```

### Workaround
Downgrading to `langchain-aws==0.2.30` resolves the issue, as the error does not occur in this version. Alternatively, explicitly specifying the region in the `ChatBedrock` configuration resolves the problem, as the component fails to retrieve the region from the `~/.aws/config` file in version `0.2.31`.

michaelnchin:
Fixes #586.


3coins (approved):
@michaelnchin 
LGTM!

Issue #591, opened 2025-08-24
bedrock_converse:  ValidationException missing toolConfig when using toolUse toolResult


00LAVA00:
When using the summarizer node with the React agent, Bedrock gives an error saying it needs `toolConfig`. This happens because some messages that are passed for summarization contain React model tool calls `ToolMessage` but with no toolConfig params in the summarizers' converse api model call.

Error:
```
ValidationException: An error occurred (ValidationException) when calling the ConverseStream operation: The toolConfig field must be defined when using toolUse and toolResult content blocks.
During task with name 'pre_model_hook' and id '04e2b263-........
```

00LAVA00:
Code to reproduce the error:


```
from langchain_aws.chat_models.bedrock_converse import ChatBedrockConverse
from langchain_core.messages import SystemMessage, HumanMessage
from langchain_core.prompts.chat import ChatPromptTemplate
from langchain_core.messages.utils import count_tokens_approximately
from langgraph.prebuilt import create_react_agent
from langmem.short_term.summarization  import SummarizationNode
from langgraph.checkpoint.memory import InMemorySaver

model = "us.anthropic.claude-3-5-haiku-20241022-v1:0"
llm_c = ChatBedrockConverse(model_id=model)


def add_tool(a, b):
    """
    Add two numbers together.
    """
    return a + b

def multiply_tool(a, b):
    """
    Multiply two numbers together.
    """
    return a * b

summarization_node = SummarizationNode(
    token_counter=count_tokens_approximately,
    model=llm_c,
    max_tokens=150,
    max_tokens_before_summary=70,
    max_summary_tokens=30,
    output_messages_key="llm_input_messages",
)

checkpointer = InMemorySaver()
agent = create_react_agent(
    model=llm_c,
    tools=[add_tool, multiply_tool],
    pre_model_hook=summarization_node,
    prompt=SystemMessage(content="You are a helpful math assistant."),
    checkpointer=checkpointer,
)


async def run():
    config = {"configurable": {"thread_id": "test"}}
    
    async for event in agent.astream_events(
        {"messages": [HumanMessage(content="test, tell me about paris - 50 words <test prompt, not related to math>")]},
        config=config,
        version="v2",
    ):
        if event.get("event") == "on_chat_model_stream":
            chunk = event["data"]["chunk"]
            print(chunk.content if hasattr(chunk, 'content') else chunk)
    

    async for event in agent.astream_events(
[...]

PR #595, opened 2025-08-27
fix(bedrock_converse): handle tool blocks without toolConfig

00LAVA00:
Fixes: #591 

The ToolConfig ValidationException happens when using summarization nodes with `create_react_agent`. The issue is that the summarization node gets messages from the react agent that contain tool calls, but the summarizer is usually a different LLM client and isn't bound with the main agent's tools. So when the summarization model calls Bedrock Converse API, it fails because there are tool blocks but no `toolConfig`.

Instead of just throwing a ValidationException, this change converts tool blocks to text format when `toolConfig` is missing. This way the summarization model still gets the tool call context but in a text format.

I considered just removing the tool messages completely, but that would make the summarization node lose context about what tools were called and their results.

This fix happens at API call level so it won't interfere with other LangChain/LangGraph modules, and it only runs when there is a mismatch between having tool blocks but no toolConfig, which shouldn't happen in normal circumstances where llms are properly bound with tools.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -65,7 +65,7 @@ _BM = TypeVar("_BM", bound=BaseModel)
 MIME_TO_FORMAT = {
     # Image formats
     "image/png": "png",
-    "image/jpeg": "jpeg", 
+    "image/jpeg": "jpeg",
     "image/gif": "gif",
     "image/webp": "webp",
     # File formats
@@ -533,9 +533,11 @@ class ChatBedrockConverse(BaseChatModel):
         return values
 
     @classmethod
-    def _get_streaming_support(cls, provider: str, model_id_lower: str) -> Union[bool, str]:
+    def _get_streaming_support(
+        cls, provider: str, model_id_lower: str
+    ) -> Union[bool, str]:
         """Determine streaming support for a given provider and model.
-        
+
         Returns:
             True: Full streaming support
             "no_tools": Streaming supported but not with tools
@@ -609,7 +611,7 @@ class ChatBedrockConverse(BaseChatModel):
     @classmethod
     def set_disable_streaming(cls, values: Dict) -> Any:
         model_id = values.get("model_id", values.get("model"))
-        
+
         # Extract provider from the model_id
         # (e.g., "amazon", "anthropic", "ai21", "meta", "mistral")
         if "provider" not in values or values["provider"] == "":
@@ -649,8 +651,8 @@ class ChatBedrockConverse(BaseChatModel):
     @model_validator(mode="after")
     def validate_environment(self) -> Self:
         """Validate that AWS credentials to and python package exists in environment."""
-        
-         # Create bedrock client for control plane API call
+
+        # Create bedrock client for control plane API call
         if self.bedrock_client is None:
             self.bedrock_client = create_aws_client(
                 region_name=self.region_name,
@@ -662,7 +664,7 @@ class ChatBedrockConverse(BaseChatModel):
                 config=self.config,
                 service_name="bedrock",
             )
-            
+
         # Handle streaming configuration for application inference profiles
         if "application-inference-profile" in self.model_id:
             self._configure_streaming_for_resolved_model()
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -780,6 +785,23 @@ def _generate(
...
+            logger.warning(
+                "Tool messages (toolUse/toolResult) detected without toolConfig. Converting tool blocks to text format to avoid ValidationException."
+            )
+            warnings.warn(
+                "Tool messages were passed without toolConfig, converting to text format",
+                RuntimeWarning,
+            )
+
+            bedrock_messages = _convert_tool_blocks_to_text(bedrock_messages)
+            logger.debug(f"converted input messages: {bedrock_messages}")
```
michaelnchin: As is, this may be quite inefficient as it loops through and checks the entire message history with `_has_tool_use_or_result_blocks`, even if `toolConfig` is already present.

Should optimize to exit early if we have `toolConfig`:
```suggestion
        # Check for tool blocks without toolConfig and handle conversion
        if params.get("toolConfig") is None and _has_tool_use_or_result_blocks(bedrock_messages):
            logger.warning(
                "Tool messages (toolUse/toolResult) detected without toolConfig. "
                "Converting tool blocks to text format to avoid ValidationException."
            )
            warnings.warn(
                "Tool messages were passed without toolConfig, converting to text format",
                RuntimeWarning,
            )

            bedrock_messages = _convert_tool_blocks_to_text(bedrock_messages)
            logger.debug(f"converted input messages: {bedrock_messages}")
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -812,6 +834,24 @@ def _stream(
...
```
michaelnchin: ^Same exit early optimization change for `stream`:
```suggestion
        # Check for tool blocks without toolConfig and handle conversion
        if params.get("toolConfig") is None and _has_tool_use_or_result_blocks(bedrock_messages):
            logger.warning(
                "Tool messages (toolUse/toolResult) detected without toolConfig. "
                "Converting tool blocks to text format to avoid ValidationException."
            )
            warnings.warn(
                "Tool messages were passed without toolConfig, converting to text format",
                RuntimeWarning,
            )

            bedrock_messages = _convert_tool_blocks_to_text(bedrock_messages)
            logger.debug(f"converted input messages: {bedrock_messages}")
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -1750,3 +1796,61 @@ def _is_cache_point(cache_point: Any) -> bool:
...
+                # convert toolResult to indicate it's tool output without exposing internal details
+                tool_result = block["toolResult"]
+
+                result_content = ""
+                for content_block in tool_result.get("content", []):
+                    if "text" in content_block:
+                        result_content += content_block["text"]
+                    elif "json" in content_block:
+                        result_content += json.dumps(content_block["json"])
+                    # skip other internal content types
```
michaelnchin: Bit of a nit, but when concatenating strings in a loop like this, it's significantly more efficient to collect them in a list, then use `.join` and build out the final string at the end. 

(Ref: https://waymoot.org/home/python_string/)

```suggestion
                content_parts = []
                for content_block in tool_result.get("content", []):
                    if "text" in content_block:
                        content_parts.append(content_block["text"])
                    elif "json" in content_block:
                        content_parts.append(json.dumps(content_block["json"]))
                    # skip other internal content types
                result_content = "".join(content_parts)
```

michaelnchin (changes requested):
Thanks @00LAVA00 ! Works great functionally, had a few observations about optimization.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
00LAVA00: make sense, done 👍

00LAVA00:
> Thanks @00LAVA00 ! Works great functionally, had a few observations about optimization.

Thank you @michaelnchin, changes applied.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
00LAVA00: Thanks for the reference!

michaelnchin (approved):
Thanks @00LAVA00 !

PR #599, opened 2025-08-28
fix: Final assistant content cannot end with trailing whitespace

### LINKED ISSUE #532: final assistant content cannot end with trailing whitespace aws bedrock

michaelnchin:
Fixes #532 

Currently, Bedrock Invoke and Converse API will fail if the penultimate message in conversation history input is an AIMessage with trailing whitespace content:
```
ValidationException: An error occurred (ValidationException) when calling the InvokeModelWithResponseStream operation: messages: final assistant content cannot end with trailing whitespace
```

This PR updates ChatBedrock and ChatBedrockConverse to check for the presence of and strip said whitespace before submission to the Bedrock APIs.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -1070,7 +1070,18 @@ def _messages_to_bedrock(
     bedrock_system: List[Dict[str, Any]] = []
     # Merge system, human, ai message runs because Anthropic expects (at most) 1
     # system message then alternating human/ai messages.
-    messages = merge_message_runs(messages)
+    
+    # Check if the last message is an AIMessage with trailing whitespace
+    messages_copy = messages.copy()
+    if messages_copy and isinstance(messages_copy[-1], AIMessage):
+        if isinstance(messages_copy[-1].content, str):
+            messages_copy[-1].content = messages_copy[-1].content.rstrip()
+        elif isinstance(messages_copy[-1].content, list):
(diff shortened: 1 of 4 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -1105,7 +1105,18 @@ def _messages_to_bedrock(
...
+    messages_copy = messages.copy()
+    if messages_copy and isinstance(messages_copy[-1], AIMessage):
+        if isinstance(messages_copy[-1].content, str):
+            messages_copy[-1].content = messages_copy[-1].content.rstrip()
+        elif isinstance(messages_copy[-1].content, list):
+            for j, block in enumerate(messages_copy[-1].content):
+                if isinstance(block, dict) and block.get("type") == "text" and isinstance(block.get("text"), str):
+                    messages_copy[-1].content[j]["text"] = block["text"].rstrip()
+    
+    messages = merge_message_runs(messages_copy)
```
3coins: minor: A slight optimization (for large message lists) would be to only copy and update the last message or directly set the content of the last message. Also, extracting this into a utility function would avoid duplication with bedrock.py.

michaelnchin: @3coins thanks! Refactored and optimized per suggestions.

michaelnchin:
Integ run: https://github.com/langchain-ai/langchain-aws/actions/runs/17782436270

PR #600, opened 2025-08-29
Adding support for Bedrock Agentcore Memory Tools

jgordley:
This PR is based on @3coins WIP draft here: https://github.com/langchain-ai/langchain-aws/pull/559 and adds support for [Bedrock Agentcore Memory](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/memory.html) operations by providing both helper functions and tool factories for the following operations:
- `store_agentcore_memory_events`
- `list_agentcore_memory_events`
- `retrieve_agentcore_memories`

## Major Differences 
- improved error handling and type hints
- helper functions to convert back and forth between LangChain `BaseMessage` and Agentcore Memory `Events`
- sample notebooks that demonstrate end to end flow using both tools and functions in LangGraph
- unit tests

<hr>

## Sample Usage

For each helper function, a tool factory is provided that can be provided to the LLM. These each depend on a pre-created Memory [See CreateMemory docs](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/create-memory.html) with strategies defined such as `Summarization` or `UserPreferences`. 

```
REGION = "us-west-2"

# Memory configuration where short term memories are stored and accessed
MEMORY_ID = "YOUR_MEMORY_ID"
SESSION_ID = "session-1"
ACTOR_ID = "user-1"

# Sample namespaces where long term memories are processed and stored by Agentcore
SUMMARY_NAMESPACE = f"/summaries/{ACTOR_ID}/{SESSION_ID}"
USER_PREFERENCES_NAMESPACE = f"/userPreferences/{ACTOR_ID}/{SESSION_ID}"

# Initialize the memory client
memory_client = MemoryClient(region_name=REGION)
```

Then once these values are defined at the start of the flow, the helper functions can be used during pre/post model hooks like in the first notebook example, or used as individual tools themselves as shown below (and in the second notebook example). Each tool factory returns a LangChain `StructuredTool`:
```
# sample creation of a tool to store memories
store_short_term_memories_tool = def create_store_memory_events_tool(
    memory_client=memory_client,
    memory_id=MEMORY_ID,
    session_id=SESSION_ID,
    actor_id=ACTOR_ID
)

# sample creation of a tool to search long term memories in the user preferences namespace
user_preferences_search_tool = create_retrieve_memory_tool(
    memory_client=memory_client,
    memory_id=MEMORY_ID,
    namespace=USER_PREFERENCES_NAMESPACE,
    tool_name="retrieve_user_preferences",
    tool_description="Search for past user preferences related to a current query",
)
```

<hr>

## Caveats/Decisions
- the AgentCore memory [CreateEvent payload](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-agentcore/client/create_event.html) does not support metadata, so for LangChain `ToolMessages` only the content can be stored as of now, not the `tool_call_id`. A TODO was left to add support for storing this attribute in the future once metadata is supported, but for now the value is just retrieved as `unknown` for tool_id. 
- the AgentCore memory short term memory event storage only supports `'ASSISTANT'|'USER'|'TOOL'|'OTHER'` as message types, so LangChain `SystemMessage` do not have a direct correlation. For implementation, I'm not sure exactly when a user would want to save a `SystemMessage` for short term memory but it could provide important context in the event of a flow restarting. For that reason, including `SystemMessage` is enabled through the a flag `include_system_messages` in the `store_agentcore_memory_events` helper function, and optionally saves those messages as `OTHER` type.
- This PR does not yet include support for the advanced memory feature `branch` filters in the ListEvents API. This will need to be added as a follow up.

### DIFF
--- a/libs/aws/langchain_aws/memory/bedrock_agentcore.py
+++ b/libs/aws/langchain_aws/memory/bedrock_agentcore.py
@@ -1,0 +1,385 @@
+"""Module for AWS Bedrock Agent Core memory integration.
+
+This module provides tools to allow agents to use the AWS Bedrock Agent Core
+memory API to manage and search memories.
+"""
+
+import logging
+from typing import List, Any, Dict
+
+from bedrock_agentcore.memory import MemoryClient
+from langchain_core.messages import (
+    AIMessage,
+    BaseMessage,
+    HumanMessage,
+    ToolMessage,
+    SystemMessage,
+)
+from langchain_core.tools import StructuredTool
+from pydantic import BaseModel, Field
+
+logger = logging.getLogger(__name__)
+
+# TODO: Once Bedrock AgentCore introduces metadata to store the tool call ID,
+# implement logic to properly save and load for ToolCall messages
+TOOL_CALL_ID_PLACEHOLDER = "unknown"
+
+
+def convert_langchain_messages_to_events(
+    messages: List[BaseMessage], include_system_messages=False
+) -> List[Dict[str, Any]]:
+    """Convert LangChain messages to Bedrock Agent Core events
+
+    Args:
+        messages: List of Langchain messages (BaseMessage)
+        include_system_messages: Flag for whether to include system messages in the conversion (as OTHER) or skip them
+
+    Returns:
+        List of AgentCore event tuples (text, role)
+    """
+    converted_messages = []
+    for msg in messages:
+        # Skip if event already saved
+        if msg.additional_kwargs.get("event_id") is not None:
+            continue
+
+        text = msg.text()
+        if not text.strip():
+            continue
+
+        # Map LangChain roles to Bedrock Agent Core roles
+        if msg.type == "human":
+            role = "USER"
+        elif msg.type == "ai":
+            role = "ASSISTANT"
+        elif msg.type == "tool":
+            role = "TOOL"
+        elif msg.type == "system" and include_system_messages:
+            role = "OTHER"
+        else:
+            logger.warning(f"Skipping unsupported message type: {msg.type}")
+            continue
+
+        converted_messages.append((text, role))
+
+    return converted_messages
+
+
+def convert_events_to_langchain_messages(
+    events: List[Dict[str, Any]]
+) -> List[BaseMessage]:
+    """Convert Bedrock Agent Core events back to LangChain messages.
+
+    Args:
+        events: List of event dictionaries with 'payload' containing conversational data
+
+    Returns:
+        List of LangChain BaseMessage objects
+    """
+    messages = []
+
+    for event in events:
+        if "payload" not in event:
+            continue
+
+        for payload_item in event.get("payload", []):
+            if "conversational" not in payload_item:
+                continue
+
+            conv = payload_item["conversational"]
+            role = conv.get("role", "")
+            content = conv.get("content", {}).get("text", "")
+
+            if not content.strip():
+                continue
+
+            message = None
+            if role == "USER":
+                message = HumanMessage(content=content)
+            elif role == "ASSISTANT":
+                message = AIMessage(content=content)
+            elif role == "TOOL":
+                # As of now, the tool_call_id is not stored or returned by the Memory API
+                message = ToolMessage(
+                    content=content, tool_call_id=TOOL_CALL_ID_PLACEHOLDER
+                )
+            elif role == "OTHER":
+                message = SystemMessage(content=content)
+            else:
+                logger.warning(f"Skipping unknown message role: {role}")
+                continue
+
+            # Preserve event metadata
+            if message and "eventId" in event:
+                message.additional_kwargs["event_id"] = event["eventId"]
+
+            if message:
+                messages.append(message)
+
+    return messages
+
+
+def store_agentcore_memory_events(
+    memory_client: MemoryClient,
+    messages: List[BaseMessage],
+    memory_id: str,
+    actor_id: str,
+    session_id: str,
+    include_system_messages: bool = False,
+) -> str:
+    """Stores Langchain Messages as Bedrock AgentCore Memory events in short term memory
+
+    Args:
+        memory_client: Initialized MemoryClient instance
+        memory_id: Memory identifier (e.g., "test-memory-id")
+        actor_id: Actor identifier (e.g., "user")
+        session_id: Session identifier (e.g., "session-1")
+        include_system_messages: Flag for whether to save system messages (as OTHER) or skip them
+
+    Returns:
+        The ID of the event that was created
+    """
+
+    if len(messages) == 0:
+        raise ValueError("The messages field cannot be empty.")
+
+    if not memory_id or not memory_id.strip():
+        raise ValueError("memory_id cannot be empty")
+    if not actor_id or not actor_id.strip():
+        raise ValueError("actor_id cannot be empty")
+    if not session_id or not session_id.strip():
+        raise ValueError("session_id cannot be empty")
+
+    events_to_store = convert_langchain_messages_to_events(
+        messages, include_system_messages
+    )
+    if not events_to_store:
+        raise ValueError(
+            "No valid messages to store. All messages were either empty, "
+            "already stored, or filtered out."
+        )
+
+    response = memory_client.create_event(
+        memory_id=memory_id,
+        actor_id=actor_id,
+        session_id=session_id,
+        messages=events_to_store,
+    )
+    event_id = response.get("eventId")
+    if not event_id:
+        raise RuntimeError("AgentCore did not return an event ID")
+
+    return event_id
+
+
+def list_agentcore_memory_events(
+    memory_client: MemoryClient,
+    memory_id: str,
+    actor_id: str,
+    session_id: str,
+    max_results: int = 100,
+) -> List[BaseMessage]:
+    """Lists the events in short term memory from Bedrock Agentcore Memory as Langchain Messages
+
+    Args:
+        memory_client: Initialized MemoryClient instance
+        memory_id: Memory identifier (e.g., "test-memory-id")
+        actor_id: Actor identifier (e.g., "user")
+        session_id: Session identifier (e.g., "session-1")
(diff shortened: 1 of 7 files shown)


samples/memory/langgraph_agent_with_memory_search.ipynb
```diff
@@ -0,0 +1,314 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "4e8f3d89-8698-4fab-8224-d0a4d0f78b8a",
+   "metadata": {},
+   "source": [
+    "## Configure Agentcore Memory\n",
+    "\n",
+    "AgentCore short-term memories are organized by a Memory ID (overall memory store) and then categorized by Actor ID (i.e which user) and Session ID (which chat session). \n",
+    "\n",
+    "Long term memories are stored in namespaces. By configuring different strategies (i.e. Summarization, User Preferences) these short term memories are processed async as long term memories in the specified namespace. We will create a separate search tool for each namespace to differentiate between `UserPreferences` and `Summaries`. The tool factories ensure that the LLM will only have to worry about sending the query and limit for searching memories and not worry about any of the configuration IDs."
```
bergjaak, line 73: "Long term memories are stored in namespaces (https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/session-actor-namespace.html)" 

that is, it would be great to link to that so the user has some clarification on what a namespace is, which I think is our most difficult concept

bergjaak: "... not worry about any of the configuration IDs." does "configuration IDs" refer to Actor, Memory and Session? if so, I think it would be good to say "... not recall the actor, session, and memory IDs."

bergjaak: "... these short term memories are processed async as long term memories... " --> "... these short term memories (i.e., conversation history) are asynchronously converted into long-term memories (i.e., memory records)..."

samples/memory/langgraph_agent_with_memory_search.ipynb
```diff
@@ -0,0 +1,314 @@
+{
+ "cells": [
```
bergjaak, line 2: should this example also explain how the events get into short-term memory (i.e. conversation history)? As the example is, I am confused how the memories are retrieved because I don't know how the short-term memory was created that enabled extraction of memory records in long-term memory

samples/memory/langgraph_stateful_agent_with_memories.ipynb
```diff
@@ -0,0 +1,542 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "820010bb-6c4e-4911-aeea-e7aa4b496109",
+   "metadata": {},
+   "source": [
+    "## Pre/Post Model Hooks\n",
+    "\n",
+    "For this implementation, short term memories (message events) are stored before and after model invocation. Before the model runs, the previous user message is saved. After the model runs, the LLM message is saved. \n",
```
bergjaak, line 123: "For this implementation, short term memories (i.e., conversation history) are stored..."

I think it's best toe refer to short term memory as conversation history, so it's clear that we're offering hosted conversation history

bergjaak: "Before the model runs, the previous user message is saved. After the model runs, the LLM message is saved"


would it be possible to async CreateEvent during the LLM invocation so that the client application doesnt have any extra time to first LLM byte? And then maybe also make it async to save the LLM event... but I think that's less important

bergjaak: also, I am making those comments assuming it will affect the actual integration code, not just this example code. Please let me know if that is wrong

samples/memory/langgraph_stateful_agent_with_memories.ipynb
```diff
@@ -0,0 +1,542 @@
...
+    "                Content: {content}\n",
+    "                ---\"\"\"\n",
+    "                \n",
+    "                context_content += \"\"\"\n",
+    "                \n",
+    "                === END HISTORY ===\n",
+    "                \n",
+    "                Please use this context to continue our conversation naturally. You should reference relevant parts of this history when appropriate, but don't explicitly mention that you're loading from memory unless asked.\n",
+    "                \"\"\"\n",
+    "                # Create a special system message with the previous context\n",
```
bergjaak, line 175: Does it make the most sense to do this as a system prompt? Or would it be better to load this into the messages array that is sent to the LLM to resemble the conversation history? (Or does LangChain not allow that?)

libs/aws/langchain_aws/memory/bedrock_agentcore.py
```diff
@@ -0,0 +1,385 @@
...
+
+    if len(messages) == 0:
+        raise ValueError("The messages field cannot be empty.")
+
+    if not memory_id or not memory_id.strip():
+        raise ValueError("memory_id cannot be empty")
+    if not actor_id or not actor_id.strip():
+        raise ValueError("actor_id cannot be empty")
+    if not session_id or not session_id.strip():
+        raise ValueError("session_id cannot be empty")
```
bergjaak, line 151: should this be a function? It is repeated a lot

libs/aws/langchain_aws/memory/bedrock_agentcore.py
```diff
@@ -0,0 +1,385 @@
...
+        include_payload=True,
+    )
+
+    return convert_events_to_langchain_messages(events)
+
+
+def retrieve_agentcore_memories(
+    memory_client: MemoryClient,
+    memory_id: str,
+    namespace_str: str,
```
bergjaak, line 215: should the function take an optional `score_threshold` to specify the required score for a memory to be returned?

libs/aws/langchain_aws/memory/bedrock_agentcore.py
```diff
@@ -0,0 +1,385 @@
...
+        messages=events_to_store,
+    )
+    event_id = response.get("eventId")
+    if not event_id:
+        raise RuntimeError("AgentCore did not return an event ID")
+
+    return event_id
+
+
+def list_agentcore_memory_events(
```
bergjaak, line 175: is it possible to paginate over events if the user wants more than the latest 100 events?

libs/aws/langchain_aws/memory/bedrock_agentcore.py
```diff
@@ -0,0 +1,385 @@
...
+        raise ValueError("The messages field cannot be empty.")
+
+    if not memory_id or not memory_id.strip():
+        raise ValueError("memory_id cannot be empty")
+    if not actor_id or not actor_id.strip():
+        raise ValueError("actor_id cannot be empty")
+    if not session_id or not session_id.strip():
+        raise ValueError("session_id cannot be empty")
+
+    events_to_store = convert_langchain_messages_to_events(
```
bergjaak, line 153: nit: we would call these AgentCore `event messages`

bergjaak: that is, an event takes an array of messages

libs/aws/langchain_aws/memory/bedrock_agentcore.py
```diff
@@ -0,0 +1,385 @@
...
+
+            conv = payload_item["conversational"]
+            role = conv.get("role", "")
+            content = conv.get("content", {}).get("text", "")
+
+            if not content.strip():
+                continue
+
+            message = None
+            if role == "USER":
```
bergjaak, line 97: since these role strings are repeated, should they be defined as variables and then referenced?

libs/aws/langchain_aws/memory/bedrock_agentcore.py
```diff
@@ -0,0 +1,385 @@
...
+    namespace_str: str,
+    query: str,
+    limit: int = 3,
+) -> List[Dict[str, Any]]:
+    """Search for memories in AWS Bedrock Agentcore Memory
+
+    Args:
+        memory_client: The AgentCore memory client
+        memory_id: The memory identifier in AgentCore
+        namespace_str: The namespace to be searched
```
gowri100, line 224: should there be a method / some way of resolving the namespace returned by [get_memory](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-agentcore-control/client/get_memory.html)? 

It returns the memory in the format: `/strategies/{memoryStrategyId}/actors/{actorId}`

libs/aws/tests/integration_tests/memory/test_agentcore_memory.py
```diff
@@ -0,0 +1,99 @@
+from unittest.mock import Mock
```
gowri100, line 1: nit: can we make the naming consistent (use bedrock agentcore everywhere instead of agentcore in some places and bedrock agent core in others?

libs/aws/langchain_aws/memory/bedrock_agentcore.py
jgordley: The MemoryClient helper from bedrock_agentcore actually handles this under the hood already: https://github.com/aws/bedrock-agentcore-sdk-python/blob/main/src/bedrock_agentcore/memory/client.py#L1148

libs/aws/langchain_aws/memory/bedrock_agentcore.py
jgordley: Good point, there are actually constants already defined for this in the agentcore SDK we can use: https://github.com/aws/bedrock-agentcore-sdk-python/blob/main/src/bedrock_agentcore/memory/constants.py#L59

libs/aws/langchain_aws/memory/bedrock_agentcore.py
jgordley: The memory client from bedrock agentcore itself does not support this yet, this may need to be on the customer to filter or only take into account messages `>n` for example, but in any case this function does return the scores with the memories so the developer can do that.

PR #601, opened 2025-09-01
feat: Add document creation for ChatBedrockConverse

hatMatch:
### Add document creation for ChatBedrockConverse

Implements document creation to add to message content. Does moderate error handling on input types though a pydantic model for the various parameters or document itself might be preferable.

### Usage
```python
# Create document
llm = ChatBedrockConverse(
    model="anthropic.claude-3-sonnet-20240229-v1:0",
)

document = ChatBedrockConverse.create_document(name="NoCite", source={"text": "we don't talk about me :("})
document_to_cite = ChatBedrockConverse.create_document(name="BigCite", source={"text": "People love talking about me"}, enable_citations=False) 

messages = [HumanMessage(content=[
    "How are my documents?", document,
    document_to_cite
])]
```


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -512,6 +512,73 @@ def create_cache_point(cls, cache_type: str = "default") -> Dict[str, Any]:
...
+            source: The source of the document.
+            context: Info for the model to understand the document for citations.
+            format: The format of the document, or its extension.
+        Returns:
+            Dictionary containing a properly formatted to add to message content."""
+        if re.match(r"[^\w\[\]\(\)-]|[\s]{2,}", name):
+            raise ValueError(
+                "Name must be only alphanumeric characters,"
+                " whitespace characters (no more than one in a row),"
+                " hyphens, parantheses, or square brackets."
```
michaelnchin: Nit: `parentheses`

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -512,6 +512,73 @@ def create_cache_point(cls, cache_type: str = "default") -> Dict[str, Any]:
...
+
+        if source.get("s3Location") and not isinstance(
+            source.get("s3Location").get("uri"), str
+        ):
+            raise ValueError(
+                "Document source with type s3Location"
+                " must have a dictionary with a valid s3 uri as a dict."
+            )
+
+        if source.get("content") and not isinstance(source.get("content", list)):
```
michaelnchin, line 566: Passing `content` source currently fails because `isinstance` is missing the second argument for type here

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -512,6 +512,73 @@ def create_cache_point(cls, cache_type: str = "default") -> Dict[str, Any]:
...
+        format: Optional[
+            Literal["pdf", "csv", "doc", "docx", "xls", "xlsx", "html", "txt", "md"]
+        ] = None,
+    ) -> Dict[str, Any]:
+        """Create a document configuration for Bedrock.
+        Args:
+            name: The name of the document.
+            source: The source of the document.
+            context: Info for the model to understand the document for citations.
+            format: The format of the document, or its extension.
```
michaelnchin, line 531: Add an `enable_citations` docstring for completeness

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -512,6 +512,73 @@ def create_cache_point(cls, cache_type: str = "default") -> Dict[str, Any]:
...
+        Returns:
+            Dictionary containing a properly formatted to add to message content."""
+        if re.match(r"[^\w\[\]\(\)-]|[\s]{2,}", name):
```
michaelnchin: `re.match` won't work here for all cases here, as this will only matches for specific invalid chars/sequences starting from the first character (for example, `No     Cite` won't be caught).

You should use `re.search` instead (+ simplify a bit):
```suggestion
        if not re.search(r"[^A-Za-z0-9 \[\]()\-]|\s{2,}", name):
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -512,6 +512,73 @@ def create_cache_point(cls, cache_type: str = "default") -> Dict[str, Any]:
...
+    @classmethod
+    def create_document(
+        cls,
+        name: str,
+        source: dict[str, Any],
+        context: Optional[str] = None,
+        enable_citations: Optional[bool] = False,
+        format: Optional[
+            Literal["pdf", "csv", "doc", "docx", "xls", "xlsx", "html", "txt", "md"]
+        ] = None,
```
michaelnchin: When I try querying Converse API without including `format` (both with `create_document` and a manually created document object), I get the following error:
```
An error occurred (ValidationException) when calling the Converse operation: Document format is required.
```

As of [boto3 1.39.0](https://boto3.amazonaws.com/v1/documentation/api/1.39.0/reference/services/bedrock-runtime/client/converse.html), this *shouldn't* be a required parameter anymore, so I'm a bit perplexed.

Would you be able to share a complete example that's working for you without passing `format` ?

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
```diff
@@ -1431,6 +1424,21 @@ def test_create_cache_point() -> None:
...
+        name="MyDoc", source={"text": "Cite me"}, enable_citations=True
+    )
+    expected_doc = {
+        "document": {
+            "name": "MyDoc",
+            "source": {"text": "Cite me"},
+            "citations": {"enabled": True},
+        }
+    }
+    assert document == expected_doc
```
michaelnchin, line 1531: This specific example fails for me, Converse API doesn't appear to recognize `text` as a valid source:
```
DocumentSource object at messages.0.content.1.document.source must set one of the following keys: bytes, s3Location.
```

Could you share a full working example with `text` if possible (or better, create an integration test to validate this)?

michaelnchin (changes requested):
Thank you for your contribution @hatMatch !

libs/aws/langchain_aws/chat_models/bedrock_converse.py
hatMatch: Ohh yes this is a strange one. ~~I believe that this depends on the type of document that you upload. If I recall correctly, it becomes non-optional when the doc is either a bytes or s3Location document. Can double check. Unfortunately will be out for a bit with an upcoming surgery.~~

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
hatMatch: Hey! Can confirm that the following code does work.
```python
client = boto3.client("bedrock-runtime", region_name="us-west-2")
response = client.converse(
    modelId="us.anthropic.claude-3-5-sonnet-20241022-v2:0",
    messages=[
        {
            "role": "user",
            "content": [
                {"text": "hi cite my source"},
                {
                    "document": {
                        "name": "hi",
                        "format": "txt",
                        "source": {"text": "hi"},
                        "citations": {"enabled": True},
                    }
                },
            ],
        }
    ],
)
```
I have a feeling there might be model specific nuance, but the docs don't really outline what that might be.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
hatMatch: In terms of a working example without format, I think that there might have been some changes to the API (or I am hallucinating). Looks like format is required. When trying the example in the comments below without format, the following error pops up:

```
ValidationException: An error occurred (ValidationException) when calling the Converse operation: Document format is required when citations are enabled and document source is bytes
```
My guess is that converse is just converting it to bytes behind the scenes.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
hatMatch: will do!

libs/aws/langchain_aws/chat_models/bedrock_converse.py
hatMatch: whoops good catch thanks!

libs/aws/langchain_aws/chat_models/bedrock_converse.py
hatMatch: ohh man my english skills are slipping away fast!

hatMatch:
Looking to follow up with this sometime next week. Will be out for a few days. Cheers and thanks for looking through this!

libs/aws/langchain_aws/chat_models/bedrock_converse.py
hatMatch: fixed in 1842cfcb450f3ab3091e14cdb0029d7d285048e1

libs/aws/langchain_aws/chat_models/bedrock_converse.py
hatMatch: fixed in 4812a051d75a468093eafb582eaa3ac73a40aa35

libs/aws/langchain_aws/chat_models/bedrock_converse.py
hatMatch: I think not was accidentally added to the front. Resolved in d7578511b46df2ddddee2cc2dea4d0e084b5565c

libs/aws/langchain_aws/chat_models/bedrock_converse.py
hatMatch: Changed format to be required in 58c7ecba0b9dcb045f75419f00189313c9cf3f8a
Also changed code since it is required in 003d288514ae4dc2e2cd9854621392b44a2d2b78

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
hatMatch: Added integration test in d648de1416a759536d561ce8285636880b7126f4 (fixed typo in 42ce46006481d19166223c9f9b9b92c279a4628c)
Updated unit test to have format in 114bc30e983e43e9aa41ae9968250f32a23fe61a

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -657,19 +721,6 @@ def set_disable_streaming(cls, values: Dict) -> Any:
     def validate_environment(self) -> Self:
         """Validate that AWS credentials to and python package exists in environment."""
 
-        # Skip creating new client if passed in constructor
-        if self.client is None:
-            self.client = create_aws_client(
```
michaelnchin, line 662: Looks like this was removed unintentionally in the 40241bd1e4bddf4c6b5f693e701fed538e4ee613 merge, let's put it back to fix the tests

hatMatch: added it back in but I did change my setup so hoping nothing else broke.

PR #607, opened 2025-09-05
fix: add support for `amer` and `global` bedrock cross region inference profiles

mikesjanebitbucketaccount:
AWS has added inference profiles that begin with "amer" and "global". This change allow ChatBedrock to correctly recognize such inference profiles


mikesjanebitbucketaccount:
@michaelnchin may be nice to get some confirmation of the `amer` endpoint as I think it's not yet officially documented

michaelnchin (approved):
Thanks @mikesjanebitbucketaccount !

Issue #604, opened 2025-09-03
ChatBedrock for Nova Micro does not work with inference profile


kodepapa:
I started using inference profiles on bedrock and had to adapt my `ChatBedrock` call accordingly:

Before:
```
llm = ChatBedrock(
    model=<modelId>,
    model_kwargs=model_kwargs,
    **provider_args,
)
```

After:
```
llm = ChatBedrock(
    model=<inferenceProfileArn>,
    provider=<provider>,
    model_kwargs=model_kwargs,
    **provider_args,
)
```

For the anthropic models, this transition worked seemlessly and everything works as before. For Amazon Nova Mirco i receive the following error:
```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: Malformed input request: #: required key [messages] not found, please reformat your input and try again.
```

I tracked it down to the actual `self.client.invoke_model(**request_options)` call in BedrockBase... but now Im clueless

michaelnchin:
Hi @kodepapa, thank you for submitting an issue. 

I couldn't reproduce the error with a with a basic test on the default Nova Micro AIP:
```
from langchain_aws import ChatBedrock
from langchain_core.messages import HumanMessage

chat = ChatBedrock(
    provider="nova",
    model_id="arn:aws:bedrock:us-east-1::foundation-model/amazon.nova-micro-v1:0",
    region_name="us-east-1"
)

message = HumanMessage(content="Hello, what's your name?")

res = chat.invoke([message])
print(res.content)
```

This might be more to do with the messages being in a format that Claude models accept, but Nova does not. Could you share a larger sample containing the input you're submitting to the model? 

Also - could you try switching to the ChatBedrockConverse model, and see if it works then? This model uses the [Bedrock Converse API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html) which is generally better equipped to handle model-specific requirements than the Invoke API that ChatBedrock uses.

kodepapa:
@michaelnchin Thanks for the swift reply.  Why do you use provider `nova`, shouldnt it be `amazon`?

here is an extended example of how i call it.
```
msg_prompt = [SystemMessage(content='You are a friendly assistant.'), HumanMessage(content='Hi')]
result = await llm.ainvoke(msg_prompt)
```

[...]

PR #608, opened 2025-09-05
fix: Ensure ChatBedrock uses Converse API for Nova-based inference profiles

michaelnchin:
Fixes: #604

ChatBedrock has a `set_beta_use_converse_api` validation method that is meant to enforce usage of Bedrock Converse API when calling Amazon Nova models. 

However, the validator only checks the value of `model_id`. If `model_id` is passed as an ambiguous Nova-based inference profile ARN, ChatBedrock will still attempt to use InvokeModel API, and eventually fail because BedrockLLM's input adapter for the `amazon` provider is only compatible with Titan models.

This PR addresses this gap in the `set_beta_use_converse_api` method by ensuring that:
1) We also check against the `base_model_id`, if provided
2) If using an inference profile that is not clearly `nova`, we call [GetInferenceProfile](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_GetInferenceProfile.html) API to retrieve the base model identity.

### DIFF
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
@@ -539,12 +539,12 @@ def test_standard_tracing_params() -> None:
 
 
 def test_beta_use_converse_api() -> None:
-    llm = ChatBedrock(model_id="nova.foo", region_name="us-west-2")  # type: ignore[call-arg]
+    llm = ChatBedrock(model_id="amazon.nova.foo", region_name="us-west-2")  # type: ignore[call-arg]
     assert llm.beta_use_converse_api
 
     llm = ChatBedrock(
         model="foobar",
-        base_model="nova.foo",
+        base_model="amazon.nova.foo",
         region_name="us-west-2")  # type: ignore[call-arg]
     assert llm.beta_use_converse_api
 


kodepapa:
Hey @michaelnchin, are you waiting for any action from my side?

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -751,9 +752,32 @@ def get_lc_namespace(cls) -> List[str]:
...
+                        aws_access_key_id=values.get("aws_access_key_id"),
+                        aws_secret_access_key=values.get("aws_secret_access_key"),
+                        aws_session_token=values.get("aws_session_token"),
+                        endpoint_url=values.get("endpoint_url"),
+                        config=values.get("config"),
+                        service_name="bedrock",
+                    )
+                response = bedrock_client.get_inference_profile(
+                    inferenceProfileIdentifier=model_id
+                )
```
dlqqq: This is a network call right? Ideally this should be async to avoid blocking the event loop, but it's fine if this is the pattern followed throughout this codebase. It makes sense given that `boto3` still lacks async support.

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -751,9 +752,32 @@ def get_lc_namespace(cls) -> List[str]:
     @classmethod
     def set_beta_use_converse_api(cls, values: Dict) -> Any:
         model_id = values.get("model_id", values.get("model"))
+        base_model_id = values.get("base_model_id", values.get("base_model", ""))
 
         if model_id and "beta_use_converse_api" not in values:
-            values["beta_use_converse_api"] = "nova" in model_id
+            values["beta_use_converse_api"] = False
+            if "nova" in model_id or "nova" in base_model_id:
```
dlqqq: Would it be safer to check for the presence of `amazon.nova` instead of just `nova`?

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -751,9 +752,32 @@ def get_lc_namespace(cls) -> List[str]:
     @classmethod
     def set_beta_use_converse_api(cls, values: Dict) -> Any:
         model_id = values.get("model_id", values.get("model"))
+        base_model_id = values.get("base_model_id", values.get("base_model", ""))
 
         if model_id and "beta_use_converse_api" not in values:
```
dlqqq: You may be able to make this method more readable by inverting this condition, returning early, and removing 1 level of indentation from the rest of this method. For example:

```suggestion
        if not model_id or "beta_use_converse_api" in values:
            return values
```

libs/aws/tests/unit_tests/chat_models/test_bedrock.py
```diff
@@ -542,18 +542,82 @@ def test_beta_use_converse_api() -> None:
...
+    mock_bedrock_client.get_inference_profile.return_value = {
+        "models": [
+            {
+                "modelArn": "arn:aws:bedrock:us-west-2::foundation-model/amazon.nova-micro-v1:0"
+            }
+        ]
+    }
+    mock_create_aws_client.return_value = mock_bedrock_client
+
+    aip_model_id = "arn:aws:bedrock:us-west-2:123456789012:application-inference-profile/my-nova-profile"
```
dlqqq: This AIP model ID has 'nova' in its name. I believe this won't call `client.get_inference_profile()` because this is caught in the first `if "nova" in model_id or "nova" in base_model_id:` branch.

It would be good to add an assertion that `get_inference_profile()` is called & handled correctly. I recommend removing 'nova' from `aip_model_id` to assert this, or add a new test that does the same thing without 'nova' in `aip_model_id`.

dlqqq (commented):
@michaelnchin Thanks for putting this together! Left a few comments. I have less experience than you have with this codebase, so all of my comments should be considered non-blocking. 🤗

libs/aws/tests/unit_tests/chat_models/test_bedrock.py
michaelnchin: Nice catch, thanks - updated to ambiguous AIP name, and added extra assertions for the `get_inference_profile` calls

libs/aws/langchain_aws/chat_models/bedrock.py
michaelnchin: Thanks! Updated as suggested.

libs/aws/langchain_aws/chat_models/bedrock.py
michaelnchin: Agreed, done 👍

libs/aws/langchain_aws/chat_models/bedrock.py
michaelnchin: Keeping this as is for now for consistency with the other API calls

PR #615, opened 2025-09-09
fix: Optimize base model identification for application inference profiles

michaelnchin:
This PR updates BedrockBase and ChatBedrockConverse to optimize the identification of base model IDs for application inference profiles. 

Currently, the `base_model_id` check and potential Bedrock GetInferenceProfile API call are done inside the `_get_base_model` method, which may cause a new API request to be executed on every invocation of the chat model. We can instead relocate this mechanism to the environment pre-validator method - this will ensure that we make the extra API request only once, on instantiation of ChatBedrock/ChatBedrockConverse.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -668,6 +668,16 @@ class ChatBedrockConverse(BaseChatModel):
                 service_name="bedrock",
             )
 
+        # For AIPs, pull base model ID via GetInferenceProfile API call
+        if self.base_model_id is None and 'application-inference-profile' in self.model_id:
+            response = self.bedrock_client.get_inference_profile(
+                inferenceProfileIdentifier=self.model_id
+            )
+            if "models" in response and len(response["models"]) > 0:
+                model_arn = response["models"][0]["modelArn"]
+                # Format: arn:aws:bedrock:region::foundation-model/provider.model-name
+                self.base_model_id = model_arn.split("/")[-1]
+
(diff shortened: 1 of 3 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -668,6 +668,16 @@ def validate_environment(self) -> Self:
                 service_name="bedrock",
             )
 
+        # For AIPs, pull base model ID via GetInferenceProfile API call
+        if self.base_model_id is None and 'application-inference-profile' in self.model_id:
+            response = self.bedrock_client.get_inference_profile(
+                inferenceProfileIdentifier=self.model_id
```
3coins: Slight formatting issue here.

```suggestion
                inferenceProfileIdentifier = self.model_id
```

libs/aws/langchain_aws/llms/bedrock.py
```diff
@@ -803,6 +803,15 @@ def validate_environment(self) -> Self:
                 service_name="bedrock",
             )
 
+        if self.base_model_id is None and 'application-inference-profile' in self.model_id:
+            response = self.bedrock_client.get_inference_profile(
+                inferenceProfileIdentifier=self.model_id
```
3coins: Slight formatting issue here.

```suggestion
                inferenceProfileIdentifier = self.model_id
```

3coins (approved):
@michaelnchin 
Overall looks good!

michaelnchin:
Integ run: https://github.com/langchain-ai/langchain-aws/actions/runs/17777999950/job/50530387878

PR #593, opened 2025-08-27
fix: Avoid ValidationException for empty content in ChatBedrockConverse

michaelnchin:
Fixes #403 

Bedrock Converse API throws a `ValidationError` if it receives input messages with empty or whitespace-only content. This PR updates ChatBedrockConverse to update said message content to an acceptable placeholder string (`[empty]`) for Converse API.

PR #618, opened 2025-09-10
aws(patch): remove empty content when upserting tool calls

ccurme:
`AIMessage("", tool_calls=[...])` is a common representation for messages with tool calls— e.g., from chat completions APIs.

Following https://github.com/langchain-ai/langchain-aws/pull/593, these get translated into
```
{'role': 'assistant', 'content': [{'text': '.'}, {'toolUse': ...}]}
```
which some Converse providers reject (Meta and Mistral, see failed test run [here](https://github.com/langchain-ai/langchain-aws/actions/runs/17602834165/job/50007989000)).

Here we patch this specific case, removing `{"text": "."}` blocks if there are tool calls and that's the only content.

This is a hack, open to other solutions.


ccurme:
Integration tests passing: https://github.com/langchain-ai/langchain-aws/actions/runs/17618829124

Issue #620, opened 2025-09-10
Outdated `supports_tool_choice` logic in ChatBedrockConverse


awg66:
You can see from the comment here: https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L677 that the logic for determining the tool choices supported by specific models is quite outdated. Specifically, it hasn't been updated since December 3rd, 2024. Specifically:
- there is no logic here for disabling specific tool choice options for claude 4 models, which only support the tool choices "auto" and "none"
- the logic here is incorrect for llama models: llama models support the "auto" tool choice, but based on the current code their `supports_tool_choice` would be an empty set

michaelnchin:
Hi @awg66, thanks for submitting an issue!

For Claude 4 models, Converse API should currently support all [`ToolChoice`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolChoice.html) values (except Anthropic's `none` option which is unsupported by Bedrock). I tested the `any` and `tool` options against Sonnet 4 and Opus 4/4.1 and things seemed to work okay.

Will add handling for llama model tool_choice - thanks!

michaelnchin:
Also we have a workaround for models whose `tool_choice` values have not been updated yet. 

You can pass the [`supports_tool_choice_values`](https://python.langchain.com/api_reference/aws/chat_models/langchain_aws.chat_models.bedrock_converse.ChatBedrockConverse.html#langchain_aws.chat_models.bedrock_converse.ChatBedrockConverse.supports_tool_choice_values) option to ChatBedrockConverse, for example:
```python
llm = ChatBedrockConverse(
    model="us.meta.llama4-maverick-17b-instruct-v1:0",
    region_name="us-west-2",
    supports_tool_choice_values=("auto",)
)
```

PR #627, opened 2025-09-13
Add tool_choice config for Llama models

michaelnchin:
Related: #620 

Allows passing `tool_choice` values for compatible Llama4 and Llama3 models when using ChatBedrockConverse.


mdrxy:
Similar here - anything blocking @michaelnchin or can we merge for release? Many thanks

mdrxy (approved):

PR #611, opened 2025-09-05
feat: Adding Bedrock AgentCore Memory Store for long term memory

jgordley:
## Summary

LangGraph memory stores are intended to be used hand in hand with checkpointers to process and create long term memories from conversations. This fits well into the Bedrock AgentCore Memory framework where chat events are saved and processed async into long term memories. 

https://langchain-ai.github.io/langgraph/concepts/persistence/#memory-store

> With this all in place, we use the in_memory_store in LangGraph. The in_memory_store works hand-in-hand with the checkpointer: the checkpointer saves state to threads, as discussed above, and the in_memory_store allows us to store arbitrary information for access across threads. We compile the graph with both the checkpointer and the in_memory_store as follows.

## Differences with LangGraph

So for this implementation, the `AgentCoreMemoryStore` was created as an alternative to `InMemoryStore`. It overrides the `put`, `search`, `get`, and `delete` methods to interact with Bedrock AgentCore Memory instead of storing the messages in a list and directly performing the embedding. 

There are some nuances with Bedrock AgentCore namespaces and the difference between short term events and long term memory records. In LangGraph, long term memories are stored directly and managed through CRUD operations (with the help of an embedding model if the user chooses). For this implementation, I played around with Bedrock AgentCore events being saved with the `put` operation, then the `search` operation will query long term memories or simply list long term memories if no query is provided.

A sample notebook walks through usage of this implementation and how it fits into a LangGraph agent.

PR #628, opened 2025-09-15
feat: Add AgentCore Memory Checkpointer

jgordley:
This PR adds the `AgentCoreMemorySaver` to be used as a LangGraph checkpointer. It implements the LangGraph [BaseCheckpointSaver](https://langchain-ai.github.io/langgraph/reference/checkpoints/#langgraph.checkpoint.base.BaseCheckpointSaver). Before reviewing, I highly recommend reading over the LangGraph Checkpointer documentation here for persistence: https://langchain-ai.github.io/langgraph/concepts/persistence/#checkpoints.

Based on https://github.com/langchain-ai/langchain-aws/pull/610 from @akshseh, without `branch` usage and with refactored serialization and helper classes.

## How it works

When initializing the `AgentCoreMemorySaver`, you must specify a `memory_id` for a preconfigured memory resource in AgentCore memory. Then when running the agent, your `thread_id` (already set in LangGraph) and `actor_id` (set by the user to associate in AgentCore memory) must be set in the runtime config for invoking the agent. This is so checkpoints are saved to unique `actor`/`session` combinations and can be retrieved efficiently.

### Saving checkpoints
In LangGraph, two types of checkpoints are saved. `put` is for whole checkpoints which contain `checkpoint_id` and channel values at that point in time. `put_writes` are intermediate writes associated with a previous `checkpoint_id` and channel value. 

Each new checkpoint and intermediate write is serialized and saved to an event in AgentCore memory in order. 

### Loading checkpoints
When a new thread is invoked, it will call `get_tuple` on the checkpointer to retrieve the latest checkpoint. To do this, AgentCore memory events are listed and pending writes are associated with the latest full checkpoint, then returned to the agent to resume execution.

Agents may also request a specific checkpoint ID, or call `get_state` or `get_execution_history`. To do this, AgentCore memory events are listed and processed to construct a full history of the `actor_id`/`thread_id` combination. There is room for caching here which has not yet been implemented. 

## New Files
- `saver.py` - the main `AgentCoreMemorySaver` class that implements `BaseCheckpointSaver`
- `helpers.py` - contains three main helper classes `EventSerializer`, `EventProcessor`, and `CheckpointEventClient`. These helpers bridge the gap between serializing, storing, and loading between LangGraph checkpoints and AgentCore Memory event operations
- `models.py` - contains the main pydantic objects for categorizing checkpoint writes, channel value writes, and pending writes
- `constants.py` - Contains a few new exceptions for error handling and a constant for empty channel values `_empty`

## Checkpointer Demo
A demo notebook has been included that shows how to use the checkpointer and how it works. Please run through this as it will tie everything together. 

## What's not included yet
- Client side caching. There is room in the implementation to implement caching based on most recent checkpoint and pending writes associated with it. For distributed agentcore deployments, this may need to be properly designed, as it would require the assumption that `actor_id`/`checkpoint_id` combinations are handled by the same agentcore runtime instance, otherwise, the data may be outdated. TBD
- Storing messages as plain text for long term memory processing - this feature is more analogous to LangGraph BaseStore, for which there is an implementation here: https://github.com/langchain-ai/langchain-aws/pull/611. However, it would be cool to identify on pending writes that a channel value was for `messages` and then store the plaintext for embedding and long term memory retrieval later, perhaps through an initialization parameter like `process_messages_for_long_term` or something like that. Will require more discussion.

Thanks for reviewing!


libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/helpers.py
```diff
@@ -0,0 +1,307 @@
...
+class EventSerializer:
+    """Handles serialization and deserialization of events to store in AgentCore Memory."""
+
+    def __init__(self, serde: SerializerProtocol):
+        self.serde = serde
+
+    def serialize_value(self, value: Any) -> Dict[str, Any]:
+        """Serialize a value using the serde protocol."""
+        type_tag, binary_data = self.serde.dumps_typed(value)
+        return {"type": type_tag, "data": base64.b64encode(binary_data).decode("utf-8")}
```
bergjaak, line 45: is there any way to do this without base64 encoding? I wonder because it increases the size of the data by 33%, so it would be great if there is some way of doing this without base64

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/helpers.py
```diff
@@ -0,0 +1,307 @@
...
+                        except EventDecodingError as e:
+                            logger.warning(f"Failed to decode event: {e}")
+
+            next_token = response.get("nextToken")
+            if not next_token or len(all_events) >= max_results:
+                break
+
+        return all_events[:max_results]
+
+    def delete_events(self, session_id: str, actor_id: str) -> None:
```
bergjaak, line 212: is this something that we expect will happen often? my concern is that this will perhaps get throttled or be reeaaalllyy slow if there are many, many events

bergjaak: It is probably fine. But I think it would be a good idea to make sure the boto3 client that is used here is configured to have longer backoff with retries (we want to make sure we're using the boto3 config because the clients that are generated by AWS have a lot of retry best practices built in, like adaptive retries [a token bucket that throttles the client if too many retried requests fail, which protects service health] and backoff with jitter) (it's usually an anti-pattern when someone is rolling their own retry code)

bergjaak: ... though these boto3 docs are not very reassuring about token bucket... it's supposed to be the default https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/helpers.py
```diff
@@ -0,0 +1,307 @@
...
+
+        self.client.create_event(
+            memoryId=self.memory_id,
+            actorId=actor_id,
+            sessionId=session_id,
+            eventTimestamp=datetime.now(UTC),
+            payload=[{"blob": serialized}],
+        )
+
+    def store_events_batch(
```
bergjaak: what happens if any of the events are too large? (~10mb should be upper bound for a CreateEvent request)
and what happens if there are more than 100 events here? (we should make sure this integration has some way of handling both of those cases instead of just failing)

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/helpers.py
```diff
@@ -0,0 +1,307 @@
...
+                            parsed_event = self.serializer.deserialize_event(blob)
+                            all_events.append(parsed_event)
+                        except EventDecodingError as e:
+                            logger.warning(f"Failed to decode event: {e}")
+
+            next_token = response.get("nextToken")
+            if not next_token or len(all_events) >= max_results:
+                break
+
+        return all_events[:max_results]
```
bergjaak: nit: I think we don't need to do this array slice since we're stopping when we have max_results, right? (the performance difference would be super negligible so only a nit)

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/helpers.py
```diff
@@ -0,0 +1,307 @@
...
+        """Retrieve events from AgentCore Memory."""
+        all_events = []
+        next_token = None
+
+        while len(all_events) < max_results:
+            params = {
+                "memoryId": self.memory_id,
+                "actorId": actor_id,
+                "sessionId": session_id,
+                "maxResults": min(100, max_results - len(all_events)),
```
bergjaak: im wondering if this is the right behavior. Is there any reason to not make a single call to ListEvents and just use the max_results from server side? (per the AWS API standards, max_results is a target and not a guarantee, so I think by moving this pagination to the client, we're sort of hacking around that API standard, which I think only makes sense if we have ListAllEvents, in which case you'd have to paginate till no NextToken)

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
```diff
@@ -0,0 +1,279 @@
...
+        config: RunnableConfig,
+        checkpoint: Checkpoint,
+        metadata: CheckpointMetadata,
+        new_versions: ChannelVersions,
+    ) -> RunnableConfig:
+        """Save a checkpoint to AgentCore Memory."""
+        checkpoint_config = CheckpointerConfig.from_runnable_config(config)
+
+        # Extract channel values
+        checkpoint_copy = checkpoint.copy()
```
bergjaak, line 160: just curious, why do we need to copy the whole checkpoint? If checkpoints can be larger than 1mb, I wonder if it'd be worth it to avoid a copy (unless there's some reason it needs to be a copy)

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
```diff
@@ -0,0 +1,279 @@
...
+            if limit is not None and count >= limit:
+                break
+
+            writes = writes_by_checkpoint.get(checkpoint_id, [])
+
+            yield self.processor.build_checkpoint_tuple(
+                checkpoint_event, writes, channel_data, checkpoint_config
+            )
+
+            count += 1
```
bergjaak, line 147: do we do anything with this count?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
```diff
@@ -0,0 +1,279 @@
...
+        if not checkpoints:
+            return None
+
+        # Find the specific checkpoint if `checkpoint_id` is provided or return the latest one
+        if checkpoint_config.checkpoint_id:
+            checkpoint_event = checkpoints.get(checkpoint_config.checkpoint_id)
+            if not checkpoint_event:
+                return None
+        else:
+            latest_checkpoint_id = max(checkpoints.keys())
```
bergjaak, line 92: what are the checkpoint keys? Do we have a guarantee that they're always increasing?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
```diff
@@ -0,0 +1,279 @@
...
+        checkpoints, writes_by_checkpoint, channel_data = self.processor.process_events(
+            events
+        )
+
+        # Build and yield CheckpointTuples
+        count = 0
+        before_checkpoint_id = get_checkpoint_id(before) if before else None
+
+        # Sort checkpoints by ID in descending order (most recent first)
+        for checkpoint_id in sorted(checkpoints.keys(), reverse=True):
```
3coins, line 129: Aren't events returned by default in reverse order?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/helpers.py
3coins: +1, if saving data (as blob) without encoding works.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
3coins: The code at line 157 seems to mutate the checkpoint_copy. I am not familiar with the internals of how LangGraph exactly uses the checkpointer object, but mutating the original object could cause unintended issues.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
```diff
@@ -0,0 +1,279 @@
...
+        writes: Sequence[tuple[str, Any]],
+        task_id: str,
+        task_path: str = "",
+    ) -> None:
+        return self.put_writes(config, writes, task_id, task_path)
+
+    async def adelete_thread(self, thread_id: str, actor_id: str) -> None:
+        return self.delete_thread(thread_id, actor_id)
+
+    def get_next_version(self, current: str | None, channel: None) -> str:
```
3coins, line 273: Would the base implementation suffice here?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/__init__.py
```diff
@@ -2,5 +2,15 @@
 LangGraph Checkpoint AWS - A LangChain checkpointer implementation using Bedrock Session Management Service.
 """
 
+from langgraph_checkpoint_aws.checkpoint.agentcore_memory.saver import (
```
3coins: We can simplify the organization by dropping the `checkpoint` folder. Let's just add `agentcore` and all files under it. Also it seems to be missing `__init__.py`. The final import state should look like:
```python
from langgraph_checkpoint_aws import AgentCoreMemorySaver
```

samples/memory/checkpointer-demo.ipynb
```diff
@@ -0,0 +1,486 @@
...
+    "3. An AgentCore Memory Resource configured (see below section for details)\n",
+    "\n",
+    "### AgentCore Memory Resource\n",
+    "\n",
+    "Either in the AWS developer portal or using the boto3 library you must create an [AgentCore Memory Resource](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-agentcore-control/client/create_memory.html). For just using the `AgentCoreMemorySaver` checkpointer in this notebook, you do not need to specify any specific long-term memory strategies. However, it may be beneficial to supplement this approach with the `AgentCoreMemoryStore` to save and extract conversational insights, so you may want to enable strategies for that use case.\n",
+    "\n",
+    "Once you have the Memory enabled and in a `ACTIVE` state, take note of the `memoryId`, we will need it later."
+   ]
+  },
+  {
```
3coins: We need another cell before imports, that installs LangChain `%pip install langchain langchain-aws`.

samples/memory/checkpointer-demo.ipynb
```diff
@@ -0,0 +1,486 @@
...
+   "cell_type": "code",
+   "execution_count": 4,
+   "id": "a8a1e31f-137a-4bd6-b959-4dc7d8176949",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Define agent state and LangGraph graph builder\n",
+    "class State(TypedDict):\n",
+    "    messages: Annotated[list, add_messages]\n",
+    "\n",
```
3coins: Rather than defining a custom graph, would prebuilt `create_react_agent` work here?
https://langchain-ai.github.io/langgraph/reference/agents/#langgraph.prebuilt.chat_agent_executor.create_react_agent

3coins (commented):
@jgordley 
Thanks for submitting these changes. Left some suggestions on reorganizing the modules to simplify the setup, and updates on the sample notebook. 

I will do some testing on checking the event size with a long conversation, and report back. As discussed, one option to explore would be to save the checkpoint event and the values separately.

samples/memory/checkpointer-demo.ipynb
jgordley: Yes, that's a great idea. I'll implement the `create_react_agent` for this sample

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/helpers.py
jgordley: Did a deep dive here on this to try and use it without `base64`. The reason we need encoding is that we are storing blobs in AgentCore as JSON, which can only handle base types (`dict`, `list`, `string`, `int`, `float`, `bool`). We can't directly convert pydantic or arbitrary value objects such as `HumanMessage` to a JSON, so we have to use a `serde` serializer such as langchain's default [JsonPlusSerializer](https://langchain-ai.github.io/langgraph/reference/checkpoints/#langgraph.checkpoint.serde.jsonplus.JsonPlusSerializer). 

Example:
```python
my_message = HumanMessage("That's got to be the best pirate I've ever seen")

from langgraph.checkpoint.serde.jsonplus import JsonPlusSerializer
serde = JsonPlusSerializer()
type_tag, binary_data = serde.dumps_typed(my_message)

# Output:
# type_tag = 'msg_pack'
# binary_data = 'b"\xc7\xbe\x05\x94\xbdlangchain_core.messages.human\xacHumanMessage\x87\xa7content\xd9/That's got to be the best pirate I've ever seen\xb1additional_kwargs\x80\xb1response_metadata\x80\xa4type\xa5human\xa4name\xc0\xa2id\xc0\xa7example\xc2\xb3model_validate_json"
```

If we tried to pass these directly to AgentCore like so, we get a validation error related to the binary type not being allowed by JSON:
```python
client.create_event(
    memoryId=MEMORY_ID,
    actorId=ACTOR_ID,
    sessionId=SESSION_ID,
    payload=[{"blob": {"type": type_tag, "data": binary_data}}],
    eventTimestamp=datetime.now(UTC)
)

# Error: ParamValidationError: Parameter validation failed:
Invalid type for document parameter data, value: b"\xc7\xbe\x05\x94\xbdlangcha...", type: <class 'bytes'>, valid types: <class 'str'>, <class 'int'>, <class 'bool'>, <class 'float'>, <class 'list'>, <class 'dict'>
```

So to get around this, we need to use `base64` (or another encoder like `hex`) to convert this binary dump into a string before we send to AgentCore:
```python
client.create_event(
    memoryId=MEMORY_ID,
    actorId=ACTOR_ID,
    sessionId=SESSION_ID,
    payload=[{"blob": {"type": type_tag, "data": base64.b64encode(binary_data).decode('utf-8')}}],
    eventTimestamp=datetime.now(UTC)
)
```

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/helpers.py
jgordley: I did some load testing myself and found that I was able to delete a thread with `120` checkpoints. It took ~30 seconds but did not run into throttling. I think this could be improved with the introduction of a `batch_delete` API. But also note that this is not a normal pattern that is used when invoking langgraph agents, its something that needs to be manually done to delete the checkpoint history (i.e. by some sort of management application or the user)

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/helpers.py
jgordley: Agreed. I'll remove this logic and let the server handle the listing max events and then break if we surpass it in the results.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
jgordley: No, this was leftover from testing I will remove this. Good catch

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
jgordley: There are some cases where checkpoints are saved in rapid succession (such as transitions between nodes with no LLM calls and very minimal latency). Due to network latency and the async nature of saving checkpoints in langgraph, I've observed cases where checkpoints are saved and returned out of order by the AgentCore API, requiring this sort operation.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
jgordley: The BaseCheckpointSaver implementation is very minimal: https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint/langgraph/checkpoint/base/__init__.py#L353

I added this to be on par with the `InMemorySaver` implementation: https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint/langgraph/checkpoint/memory/__init__.py#L515

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
jgordley: Yes it's part of the `limit` logic in this function, i.e. someone calls `list(..., limit=3)` it will break and stop yielding after 3 loops.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore_memory/saver.py
jgordley: Yes, according to the [LangChain docstrings here](https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint/langgraph/checkpoint/base/__init__.py#L66)

> The ID of the checkpoint. This is both unique and monotonically
    increasing, so can be used for sorting checkpoints from first to last.

jgordley:
Added unit tests and integration tests and ran them locally:

```
➜  langgraph-checkpoint-aws git:(agentcore-memory-checkpointer) ✗ make integration_test TEST_FILE=tests/integration_tests/agentcore/

poetry run pytest tests/integration_tests/
========================================== test session starts ===========================================
platform darwin -- Python 3.11.10, pytest-8.4.1, pluggy-1.6.0
rootdir: /Users/jgordle/projects/jgordle-forks/langchain-aws/libs/langgraph-checkpoint-aws
configfile: pyproject.toml
plugins: anyio-4.10.0, cov-6.2.1, langsmith-0.4.13
collected 10 items

tests/integration_tests/agentcore/test_saver.py ......                                             [ 60%]
tests/integration_tests/saver/test_saver.py ...                                                    [ 90%]
tests/integration_tests/test_compile.py .                                                          [100%]

========================================== slowest 5 durations ===========================================
25.86s call     tests/integration_tests/agentcore/test_saver.py::TestAgentCoreMemorySaver::test_checkpoint_listing_with_limit
8.89s call     tests/integration_tests/agentcore/test_saver.py::TestAgentCoreMemorySaver::test_multiple_sessions_isolation
8.60s call     tests/integration_tests/agentcore/test_saver.py::TestAgentCoreMemorySaver::test_math_agent_with_checkpointing
8.11s call     tests/integration_tests/saver/test_saver.py::TestBedrockMemorySaver::test_weather_query_and_checkpointing
5.98s call     tests/integration_tests/agentcore/test_saver.py::TestAgentCoreMemorySaver::test_weather_query_with_checkpointing
===================================== 10 passed in 61.18s (0:01:01) ====================================
```

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
```diff
@@ -0,0 +1,309 @@
...
+                    blob = payload_item.get("blob")
+                    if blob:
+                        try:
+                            parsed_event = self.serializer.deserialize_event(blob)
+                            all_events.append(parsed_event)
+                        except EventDecodingError as e:
+                            logger.warning(f"Failed to decode event: {e}")
+
+            next_token = response.get("nextToken")
+            if not next_token or (limit and len(all_events) >= limit):
```
bergjaak: does anything bad happen if limit is set to 0?

jgordley: Good call, I'll switch the check to  != 0, in the rare event that someone does submit with limit=0

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
```diff
@@ -0,0 +1,309 @@
...
+            else:
+                raise EventDecodingError(f"Unknown event type: {event_type}")
+
+        except json.JSONDecodeError as e:
+            raise EventDecodingError(f"Failed to parse JSON: {e}")
+        except Exception as e:
+            raise EventDecodingError(f"Failed to deserialize event: {e}")
+
+
+class CheckpointEventClient:
```
3coins: Should we rename this to AgentCoreMemoryClient or simply MemoryClient? Also, would any part of this be shared with the Store?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
```diff
@@ -0,0 +1,309 @@
...
+class EventSerializer:
+    """Handles serialization and deserialization of events to store in AgentCore Memory."""
+
+    def __init__(self, serde: SerializerProtocol):
+        self.serde = serde
+
+    def serialize_value(self, value: Any) -> Dict[str, Any]:
+        """Serialize a value using the serde protocol."""
+        type_tag, binary_data = self.serde.dumps_typed(value)
+        return {"type": type_tag, "data": base64.b64encode(binary_data).decode("utf-8")}
```
3coins, line 45: Did you look into saving blob without b64 encode which was brought up in the previous comments?

samples/memory/agentcore_memory_checkpointer.ipynb
```diff
@@ -0,0 +1,465 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "eda8649a-d48f-467c-94b9-99dd8fb2c078",
+   "metadata": {},
+   "outputs": [],
+   "source": []
+  }
```
3coins, line 390: This section looks irrelevant. As long as we have correctly setup the saver to require actor_id, users can apply those based on their preferences.

samples/memory/agentcore_memory_checkpointer.ipynb
```diff
@@ -0,0 +1,465 @@
...
+    "@tool\n",
+    "def multiply(a: int, b: int):\n",
+    "    \"\"\"Multiply two integers and return the result\"\"\"\n",
+    "    return a * b\n",
+    "\n",
+    "\n",
+    "tools = [add, multiply]\n",
+    "\n",
+    "# Bind the tools to our LLM so it can understand their structure\n",
+    "llm_with_tools = llm.bind_tools(tools)"
```
3coins: React agent might be doing that on creation, can you check that, and make sure we are not double binding the tools.

samples/memory/agentcore_memory_checkpointer_human_loop.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "@tool\n",
+    "def multiply(a: int, b: int):\n",
+    "    \"\"\"Multiply two integers and return the result\"\"\"\n",
+    "    return a * b\n",
+    "\n",
+    "\n",
+    "tools = [add, multiply, human_assistance]\n",
+    "\n",
+    "# Bind the tools to our LLM so it can understand their structure\n",
+    "llm_with_tools = llm.bind_tools(tools)"
```
3coins: Is this required? Can you check if create_react_agent binds tools without you explicitly doing it.

samples/memory/agentcore_memory_checkpointer_human_loop.ipynb
```diff
@@ -0,0 +1,340 @@
...
+      "Tool Calls:\n",
+      "  human_assistance (tooluse_6FDef0XWQhGv0aiA3D3_IQ)\n",
+      " Call ID: tooluse_6FDef0XWQhGv0aiA3D3_IQ\n",
+      "  Args:\n",
+      "    query: User is frustrated with current support and has requested to speak with a human representative.\n"
+     ]
+    }
+   ],
+   "source": [
+    "user_input = \"I'm frustrated with my current support. Please loop in a human.\"\n",
```
3coins: We can change this rather to a more direct command.
```suggestion
    "user_input = \"I would like to work with a customer service human agent.\"\n",
```

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
jgordley: Yes I'll rename it to be more specific for agentcore and yes the store will make use of some of these methods too.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
jgordley: Yes see above https://github.com/langchain-ai/langchain-aws/pull/628#discussion_r2356498440

samples/memory/agentcore_memory_checkpointer.ipynb
jgordley: Ah yes it does. Good catch.

samples/memory/agentcore_memory_checkpointer_human_loop.ipynb
jgordley: Removed this in the latest commit.

samples/memory/agentcore_memory_checkpointer.ipynb
```diff
@@ -0,0 +1,413 @@
...
+     },
+     "execution_count": 14,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "graph = create_react_agent(\n",
+    "    model=llm,\n",
+    "    tools=[add, multiply],\n",
```
3coins:
```suggestion
    "    tools=tools,\n",
```

samples/memory/agentcore_memory_checkpointer.ipynb
```diff
@@ -0,0 +1,413 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e6acbf4e",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install langchain"
```
3coins:
```suggestion
    "%pip install langchain langchain-aws"
```

PR #638, opened 2025-09-16
fix(aws): (bedrock) support thinking blocks in streaming, multi-turn contexts

ccurme:
In this scenario:
- Thinking is enabled in ChatBedrock
- User is streaming
- User is in a multi-turn conversation (i.e., passing in chat history)

ChatBedrock breaks immediately due to the unrecognized `index` key which is added in streaming.

Here we also add a test for citations, which is currently skipped as it will require breaking changes— we need to not coerce content to string when citations are enabled. Citations are currently unsupported for `.invoke` and break when you run `.stream`.

### DIFF
--- a/libs/aws/tests/integration_tests/chat_models/test_bedrock.py
+++ b/libs/aws/tests/integration_tests/chat_models/test_bedrock.py
@@ -1,13 +1,14 @@
 """Test Bedrock chat model."""
 
 import json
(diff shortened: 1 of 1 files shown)


ccurme:
Integration tests pass: https://github.com/langchain-ai/langchain-aws/actions/runs/17771004268

PR #648, opened 2025-09-18
Format / lint the repo

awg66:
I noticed a few issues while doing my first PR:
- Ruff was locked to version 0.1.35. The latest version is 0.13.0!
- Running the format / lint commands results in a lot of failures with master. This means that when submitting PRs for actual code changes, if your editor is setup to auto lint / format on save (like mine) you end up picking up a lot of extraneous code changes

So this PR proposes two solutions:
- Update pyproject.toml to bring ruff up-to-date from the stone-age
- Run the formatter / linter and commit the results!

I also made a slight tweak to the pyproject.toml file (which can / probably should also be linted) to not use strings for boolean TOML settings

There are still linting errors but they mostly fall into two categories: very long strings and unused variables. Decided to leave those and tackle them in a separate review. At least they won't get picked up by auto formatters / linters, but you will just see them in your editor if you have ruff / mypy enabled


michaelnchin (approved):
Thanks @awg66! We're already planning on including the ruff upgrade and linting with the v1.0.0 release changes pending in PR #597, but probably no harm committing this now.

PR #652, opened 2025-09-20
fix(aws): (v1) Remove beta_use_converse_api from ChatBedrock

michaelnchin:
**Note:** Do not merge until week of Oct. 20.

This PR makes two changes to the ChatBedrock chat model in order to encourage users to migrate to the new, dedicated ChatBedrockConverse class:
- Removes the [`beta_use_converse_api`](https://python.langchain.com/api_reference/aws/chat_models/langchain_aws.chat_models.bedrock.ChatBedrock.html#langchain_aws.chat_models.bedrock.ChatBedrock.beta_use_converse_api) parameter.
- Adds a new `check_unsupported_model` Pydantic validator, which will immediately throw an exception if attempting to use a model that we don't support in ChatBedrock (ex. Nova models.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,12 +44,11 @@ pip install langgraph-checkpoint-aws
 Here's a simple example of how to use the `langchain-aws` package.
 
 ```python
-from langchain_aws import ChatBedrock
+from langchain_aws import ChatBedrockConverse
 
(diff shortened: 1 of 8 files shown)


3coins:
Did we add this in v1.0.0?

Issue #619, opened 2025-09-10
Taking boto3.Session as input for BedrockSessionSaver


whkt:
Currently, BedrockSessionSaver there is not an option to provide boto3.Session as input. Many long running application with customized credential provider initialized along with a boto3.Session thus it is crucial to have this option as an input.

Suggestion:

```python
saver = BedrockSessionSaver(session=boto3.Session())
```

00LAVA00:
Hi @michaelnchin , I found this one interesting. I have some free time to learn and would love to help implement session as input for BedrockSessionSaver. 

Would it make sense for me to pick this up? I previously worked on [#595] but am still learning, so any pointers or prerequisites you recommend would be great.

michaelnchin:
Hi @00LAVA00, please feel free to! You can take a look at the [Contributing](https://github.com/langchain-ai/langchain-aws/blob/main/libs/langgraph-checkpoint-aws/CONTRIBUTING.md) guide to get started on local development for the checkpointer package.

PR #654, opened 2025-09-23
feat(checkpointer): add session parameter support

00LAVA00:
Fixes: #619

Added an extra parameter as input for the BedrockSessionSaver to pass customized credential session info.

This will allow users to pass their own pre configured boto3 session like this:

```
# Before
saver = BedrockSessionSaver(region_name="us-west-2", aws_access_key_id=..., aws_secret_access_key=...)

# Now: Can pass custom session with any credential
session = boto3.Session(profile_name="my-custom-profile")
saver = BedrockSessionSaver(session=session)
```

So when a session is provided, it's used directly instead of creating a new one from individual credential parameters.

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/async_saver.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/async_saver.py
@@ -3,6 +3,7 @@ import json
 from collections.abc import AsyncIterator, Sequence
 from typing import Any, Optional
 
+import boto3
 from botocore.config import Config
 from botocore.exceptions import ClientError
 from langchain_core.runnables import RunnableConfig
@@ -48,6 +49,7 @@ class AsyncBedrockSessionSaver(BaseCheckpointSaver):
     It handles creating invocations, managing checkpoint data, and tracking pending writes.
 
     Args:
+        session: Pre-configured boto3 session instance for custom credential
         region_name: AWS region name
         credentials_profile_name: AWS credentials profile name
         aws_access_key_id: AWS access key ID
@@ -59,6 +61,7 @@ class AsyncBedrockSessionSaver(BaseCheckpointSaver):
 
     def __init__(
         self,
+        session: Optional["boto3.Session"] = None,
         region_name: Optional[str] = None,
         credentials_profile_name: Optional[str] = None,
         aws_access_key_id: Optional[SecretStr] = None,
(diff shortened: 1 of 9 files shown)


00LAVA00:
Hi @michaelnchin, I’ve implemented the **session** parameter. Should I also add support for passing a pre configured client directly (similar to how BedrockLLM works), or is the session based approach sufficient for now?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/async_saver.py
```diff
@@ -59,6 +61,7 @@ class AsyncBedrockSessionSaver(BaseCheckpointSaver):
 
     def __init__(
         self,
+        session: Optional["boto3.Session"] = None,
```
michaelnchin: Do we need to use quotes here and in `saver.py`? I see that the `SessionClient` classes don't use them for the same parameter.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/async_session.py
```diff
@@ -47,23 +49,45 @@ def __init__(
...
+            # Create session using provided credentials
+            _session_kwargs, self._client_kwargs = process_aws_client_args(
+                region_name,
+                credentials_profile_name,
+                aws_access_key_id,
+                aws_secret_access_key,
+                aws_session_token,
+                endpoint_url,
+                config,
+            )
```
michaelnchin: Can we keep this outside the conditional? I don't think the simplified/duplicated logic in the first block adds much value in regards to efficiency

michaelnchin (changes requested):
Thanks for the contribution @00LAVA00 ! Couple of minor comments.

A client parameter would also be good to have, but you can add that in a separate PR.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/async_saver.py
00LAVA00: Yes, removed quotes from saver files.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/async_session.py
00LAVA00: Understood, simplified the logic.

00LAVA00:
> Thanks for the contribution @00LAVA00 ! Couple of minor comments.
> 
> A client parameter would also be good to have, but you can add that in a separate PR.

Thank you @michaelnchin

For a separate PR, do I also have to first create an issue mentioning the client requirement and then link it with the pr?

michaelnchin:
Integ tests: https://github.com/langchain-ai/langchain-aws/actions/runs/17975349895

PR #667, opened 2025-10-01
Adding tool factories for the agentcore memory store

jgordley:
This PR adds LangMem style tool factories to provide to the agent for saving events and searching for memories at runtime.

It is based on the LangMem tool factories with the ability to inject the base store at runtime and use the runtime config `actor_id` and `session_id` to save and search for memories according to the runtime config.

Reference implementation: https://github.com/langchain-ai/langmem/blob/main/src/langmem/knowledge/tools.py

```python
# Create the memory save tool using the factory
# The actor ID and session ID are handled under the hood at runtime to save to the actor and session based on the invocation below
save_events_to_memory_tool = create_store_event_tool(
    name="save_events_to_memory",
)

# Create the memory search tool using the factory
retrieve_past_conversation_facts_tool = create_search_memory_tool(
    namespace=("facts", "{actor_id}"), # Placeholder for actor ID, specifying the namespace we defined /facts/{actorId} in AgentCore Memory
    instructions="Retrieve facts and user preferences about the user that might be helpful in answering vague questions",
    name="get_past_conversation_facts",
)
```

Comes with a sample notebook.

### DIFF
--- a/samples/memory/agentcore_memory_store_long_term_search_tools.ipynb
+++ b/samples/memory/agentcore_memory_store_long_term_search_tools.ipynb
@@ -33,7 +33,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -50,7 +50,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -81,7 +81,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 17,
+   "execution_count": null,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


bergjaak:
My thinking is, we designed our system so that CreateEvent should always be called for all LLM messages, and then the decision to extract a memory record from the message belongs to our service backend. So then I can see it being more straightforward for the customer if we just tell them: "Store all messages," and so if we take that stance, it would imply that we don't want to have a tool for calling CreateEvent (since then it confuses customers into adopting the anti-pattern where they don't store all events)

austinmw:
Hey @bergjaak , just to follow up and summarize what we talked about, I think users will want all of the following:

- Easily save and retrieve all messages (including tool calls and responses) with simple store/retrieve commands, without needing to parse the individual complex langchain message objects in `state['messages']`
- Also store other non-message information saved in langgraph `state` (probably with the same store/retrieve commands)
- Store messages via a single method instead of needing to store in two separate ways (blob for full state and conversational for extraction purposes)
- Support automatically extracting long-term memories from short-term memory (from `state['messages']`)

Issue #669, opened 2025-10-01
Cross-region inference (jp) support


masucamp:
Claude Sonnet 4.5, which is now supported on Amazon Bedrock, has a profile for cross-region inference that is limited to Japan.
The ID of the inference profile is jp.anthropic.claude-sonnet-4-5-20250929-v1:0.

If you configure this and then call LangChain, the following error will be returned:

```
NotImplementedError: Provider jp model does not support chat.
```

This is probably because jp is not included in the _get_provider function.

https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/llms/bedrock.py#L878

```
        return (
            parts[1]
            if (
                len(parts) > 1
                and parts[0].lower()
                in {"eu", "us", "us-gov", "apac", "sa", "amer", "global"}
            )
            else parts[0]
        )
```

PR #671, opened 2025-10-01
fix: Support `jp` cross region inference profiles

michaelnchin:
Addresses #669.

Updated ChatBedrock/ChatBedrockConverse to account for new `jp`Claude Sonnet 4.5 inference profiles.

See: https://docs.aws.amazon.com/bedrock/latest/userguide/inference-profiles-support.html


3coins (approved):
Looks good!

PR #692, opened 2025-10-14
docs: Migrate from reST to Markdown format

michaelnchin:
Addresses #687.


libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -1315,81 +1315,81 @@ def with_structured_output(
...
+            class AnswerWithJustification(BaseModel):
+                '''An answer to the user question along with justification for the answer.'''
+                answer: str
+                justification: str
 
-                class AnswerWithJustification(BaseModel):
-                    '''An answer to the user question along with justification for the answer.'''
-                    answer: str
-                    justification: str
+            llm =ChatBedrock(
```
mdrxy: One more thing that's come up since opening the issue - we're now using `model = ChatModel(...)` instead of `llm = C...`

michaelnchin: Updated, thanks!

PR #694, opened 2025-10-14
feat: Support ElastiCache Valkey LangGraph and Agentcore checkpointers

seaofawareness:
PR adds support for using ElastiCache [Valkey](https://valkey.io/) datastores as LangGraph checkpointer to store short-term (thread-scoped), long-term (user/application specific) memory, and LLM prompt caching. It also adds Agentcore checkpointer for Valkey.


3coins (changes requested):
@seaofawareness 
Thanks for submitting this change. It will help to split these changes into smaller PRs, which will allow us to iterate quickly on these changes. I would recommend submitting just the Checkpointer changes, and then move onto the Store and Cache implementations.

3coins:
Replaced by #697

Issue #713, opened 2025-10-20
Passing any value for `model_kwargs` in ChatBedrockConverse results in Circular reference error


peterkarman1:
When trying to pass model_kwargs, using this parameter at all causes a circular reference error. Here is a minimal reproduction

```
model = ChatBedrockConverse(
            model="qwen.qwen3-coder-480b-a35b-v1:0",
            model_kwargs={"test": AppConfig.MAX_OUTPUT_TOKENS},
            region_name=region_name,
            stop=None,
            client=client,
        )

model.invoke("hello world")
```

````
/lib/python3.12/site-packages/langchain_core/load/dump.py", line 64, in dumps\n    return json.dumps(obj, default=default, **kwargs)\n           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "/Users/peterkarman/.local/share/uv/python/cpython-3.12.8-macos-aarch64-none/lib/python3.12/json/__init__.py", line 238, in dumps\n    **kw).encode(obj)\n          ^^^^^^^^^^^\n  File "/Users/peterkarman/.local/share/uv/python/cpython-3.12.8-macos-aarch64-none/lib/python3.12/json/encoder.py", line 200, in encode\n    chunks = self.iterencode(o, _one_shot=True)\n             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File "/Users/peterkarman/.local/share/uv/python/cpython-3.12.8-macos-aarch64-none/lib/python3.12/json/encoder.py", line 258, in iterencode\n    return _iterencode(o, 0)\n           ^^^^^^^^^^^^^^^^^\nValueError: Circular reference detected\n'
```

michaelnchin:
@peterkarman1 thanks! A fix to handle this error is pending in #715.

[...]

PR #715, opened 2025-10-21
fix(aws): Correctly handle serialization for model_kwargs in ChatBedrockConverse

michaelnchin:
Fixes: #713

Updated the `build_extra` model args validator in `ChatBedrockConverse` to correctly handle direct `model_kwargs` user input. Previously, this would interfere with the call to `langchain-core`'s [`_build_model_kwargs`](https://github.com/langchain-ai/langchain/blob/d9e659ca4f020908a360118e7a2a225d86e4b555/libs/core/langchain_core/utils/utils.py#L217), which would throw a `Circular reference detected` error during JSON serialization.

Also added a user warning to switch to the correct `additional_model_request_fields` accepted by both `ChatBedrockConverse`, per the underlying API.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -543,12 +543,14 @@ class ChatBedrockConverse(BaseChatModel):
             "additional_model_request_fields", {}
         )
         if model_kwargs:
-            model_kwargs_msg = (
-                "Please use additional_model_request_fields instead of "
-                "model_kwargs for any extra inference parameters."
-            )
-            logger.warning(model_kwargs_msg)
-            warnings.warn(model_kwargs_msg)
+            if model_kwargs:
+                warnings.warn(
+                    "ChatBedrockConverse uses 'additional_model_request_fields' "
+                    "instead of 'model_kwargs'. Your parameters have been automatically"
+                    " converted.",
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -538,17 +538,25 @@ def create_cache_point(cls, cache_type: str = "default") -> Dict[str, Any]:
...
             "additional_model_request_fields", {}
         )
-        if additional_model_request_fields or model_kwargs:
+        if model_kwargs:
+            model_kwargs_msg = (
+                "Please use additional_model_request_fields instead of "
+                "model_kwargs for any extra inference parameters."
+            )
+            logger.warning(model_kwargs_msg)
+            warnings.warn(model_kwargs_msg)
```
3coins: Are these 2 redundant? Also, a form like this will help folks filter these.

```python
if model_kwargs:
        warnings.warn(
            "ChatBedrockConverse uses 'additional_model_request_fields' instead of 'model_kwargs'. "
            "Your parameters have been automatically converted.",
            UserWarning,
            stacklevel=2
        )

```

michaelnchin: Thanks 👍 Removed logger statement and updated with suggested params

Issue #712, opened 2025-10-20
Bug [langgraph_checkpoint_aws] -  agentcore serializer miss class serialization in writes event


pppazos:
The helper for serialize_event(), in the case the event is WritesEvent does not mantain the original class, serializing it as dictionary.

ex. event:
```python
 WritesEvent(
        event_type='writes',
        checkpoint_id='1f0ac518-084c-6d84-8001-d72be2d5493b',
        writes=[
            WriteItem(
                task_id='d3c2cbaa-5b30-0fae-94c3-f8a41da849d4',
                channel='__interrupt__',
                value=(Interrupt(
                    value="Trying to call tool xxxx. Please approve or suggest edits.",
                    id='0d1508ae263740a8ffc819cbe41e36ff'
                ),),
                task_path='~__pregel_push, 0000000000, 0000000000'
            )
        ]
    )
```
when serialized and deserialized, the value gets converted to dictionary:
```python
WritesEvent(
    event_type="writes",
    checkpoint_id="1f0ac518-084c-6d84-8001-d72be2d5493b",
    writes=[
        WriteItem(
            task_id="d3c2cbaa-5b30-0fae-94c3-f8a41da849d4",
            channel="__interrupt__",
            value=[
                {
                    "value": "Trying to call tool xxxx. Please approve or suggest edits.",
                    "id": "0d1508ae263740a8ffc819cbe41e36ff",
                }
            ],
            task_path="~__pregel_push, 0000000000, 0000000000",
        )
    ],
)
```
This creates an issue when ad example an Interrupt gets deserialized as dictionary and langgraph try to access a class property like ".id"

proposed solution in serialize_event() from
```python
        elif isinstance(event, WritesEvent):
            # The writes field is already properly serialized by model_dump()
            # We just need to serialize the value field in each write
            for write in event_dict["writes"]:
                val = write.get("value", EMPTY_CHANNEL_VALUE)
                write["value"] = self.serialize_value(val)
```

to
```python
    elif isinstance(event, WritesEvent):
        event_dict["writes"] = [
            {
                **write.model_dump(exclude_none=True),
                "value": self.serialize_value(write.value)
            }
            for write in event.writes
        ]
```

michaelnchin:
Hi @pppazos! Thank you for reporting this, and looking into a potential solution.

If you are able, could you submit a pull request with the proposed changes for us to review?

PR #716, opened 2025-10-21
fix(langgraph-checkpoint-aws): serialize writes in serialize_envet issue #712

pppazos:
Proposed fix for serialization issue reported in #712


michaelnchin (approved):
Thanks @pppazos !

Issue #261, opened 2024-10-29
Enhance Bedrock Agents support to accommodate Messages, ToolNode


3coins:
@kcgourishetti 
I took care of some of the formatting, linting, typing and CI errors. One last thing I noticed is that the notebook `bedrock_agent_langgraph.ipynb` seems to be incomplete, it doesn't have any of the code for the LangGraph executor, you seem to have some of this code in the integration test.

A couple of other observations, not blockers on this PR:
1. Even though there is no LLM present here (agent manages the LLM calls), a lot of the LangChain components work around `messages` which is central to it's working. We should follow up with a PR to support messages.
2. The LangGraph integration test is using the [ToolExecutor](https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/prebuilt/tool_executor.py#L17), which has been deprecated in favor of [ToolNode](https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/prebuilt/tool_node.py#L42), which works only with messages. To future proof this work, we should move to use ToolNode in LangGraph samples and tests.

_Originally posted by @3coins in https://github.com/langchain-ai/langchain-aws/issues/152#issuecomment-2297409880_

PR #718, opened 2025-10-21
chore(aws): Update BedrockAgents integ tests to use `langchain-classic`

michaelnchin:
Re-enabling BedrockAgents integ tests as the legacy `AgentExecutor` is now available with the `langchain-classic` release.


michaelnchin (commented):
(Non-blocking) Note: `test_bedrock_agent_langgraph` is [using `ToolExecutor`](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/tests/integration_tests/agents/test_bedrock_agents.py#L315), which was removed in LangGraph [v0.3](https://github.com/langchain-ai/langgraph/releases/tag/0.3.1).

Per issue #261, we can't switch to the replacement [`ToolNode`](https://langchain-ai.github.io/langgraph/how-tos/tool-calling/#toolnode) prebuilt yet because `BedrockAgentsRunnable` doesn't support messages yet. However, this isn't currently a blocker for the CI or builds since the test is skipped anyway.

The other changes in this PR are unrelated, and should be OK to merge.

Issue #719, opened 2025-10-22
Hardcoded limit to checkpoints retrieval for AgentCore LangGraph Saver


michaelnchin:
### Discussed in https://github.com/langchain-ai/langchain-aws/discussions/709

<div type='discussions-op-text'>

<sup>Originally posted by **ltoniazzi** October 20, 2025</sup>
Seems that the `get_events` method defaults to `100` limit of events ([here](https://github.com/langchain-ai/langchain-aws/blob/main/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py#L185)).

But it's always used without passing the limit param [here](https://github.com/langchain-ai/langchain-aws/blob/main/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py#L79).

So the default limit is always used and the user has no control over it.

**Question**:
Is this intentional? Because it seems that a `thread_id` memory with more than 100 nodes executions will be incomplete when retrieved from AgentCore.</div>

ltoniazzi:
I think that if you plan to keep the `limit` parameter, then there should be a warning [here](https://github.com/langchain-ai/langchain-aws/blob/main/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py#L223) if the pagination stops because of `limit` and not `next_token`:
```python
if next_token and (limit is not None and len(all_events) >= limit)
    logger.warning("Might be missing checkpoints...")
```

PR #728, opened 2025-10-25
fix(langgraph-checkpoint-aws): Add `max_results` and `limit` attributes to AgentCoreMemorySaver

michaelnchin:
Addresses #719.

This PR updates the AgentCore memory checkpointer integration to:
- Add an `max_results` attribute to `AgentCoreMemorySaver`, allowing control of the `maxResults` value passed to [ListEvents API](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_ListEvents.html). 
  - Effective default will remain at the previously hardcoded limit of 100 results.
- Add another `limit` attribute to control the maximum events parsed when paginating on the ListEvents response.
  - Effective default will move from 100 -> `None` (unlimited).
- In `AgentCoreEventClient,get_events()`, exit from the response processing loop early if the number of events found exceeds the `limit` attribute value. If this occurs, a warning will be returned to the user.


michaelnchin:
Hi @ltoniazzi, thanks for the suggestions in #719 - feel free to try out these changes and let me know if you have any additional feedback.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
```diff
@@ -253,8 +268,10 @@ def delete_thread(self, thread_id: str, actor_id: str = "") -> None:
         self.checkpoint_event_client.delete_events(thread_id, actor_id)
 
     # ===== Async methods ( TODO: NOT IMPLEMENTED YET ) =====
-    async def aget_tuple(self, config: RunnableConfig) -> CheckpointTuple | None:
-        return self.get_tuple(config)
+    async def aget_tuple(
+        self, config: RunnableConfig, *, limit: int | None = None
```
ltoniazzi: I think `limit` is like a connection parameter, so it could be good to set it as an attribute when instantiating the saver, like for the [postgres saver](https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py#L39).

Also the user will not use `aget_tuple`, but usually only instantiate the saver. Other methods, like  `aget`,  `aput`, are usually  only called by the langgraph internals.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
```diff
@@ -77,7 +90,9 @@ def get_tuple(self, config: RunnableConfig) -> CheckpointTuple | None:
         )
 
         events = self.checkpoint_event_client.get_events(
-            checkpoint_config.session_id, checkpoint_config.actor_id
+            checkpoint_config.session_id,
+            checkpoint_config.actor_id,
+            100 if limit is None else limit,
```
ltoniazzi: I think `None` could be allowed as a parameter?

If limit is an attribute it could just be `self.limit` here and set a default in the `__init__`?

michaelnchin: Done - set attribute default to `100`, can also pass `None` for no limit.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
michaelnchin: Great suggestion, thanks! Added `limit` as a class attribute instead, and updated `get_tuple`/`list` to pass this value to `get_events`.

michaelnchin:
@jgordley @bergjaak - ready for review

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
```diff
@@ -184,15 +185,25 @@ def store_blob_events_batch(
...
 
         if limit is not None and limit <= 0:
             return []
 
         all_events = []
         next_token = None
+        limit_reached = False
 
         while True:
             params = {
```
bergjaak, line 214: we should pass limit as maxResults in the params here

michaelnchin: Done, renamed the attribute to `max_results`. Also a good change to reduce confusion with the the base `list()` implementation's existing `limit` parameter , which applies specifically to checkpoints.

ltoniazzi:
I wonder if it should default to `None` instead of `100`, as I think saver always tries to list events for a `thread_id`, so I do not see a reason why one would not want to fetch the full checkpoint history by default (also `limit` in [other savers defaults to](https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint-postgres/langgraph/checkpoint/postgres/aio.py#L115) `None` as well).

Also, I'm a bit surprised that this change does not break any unit test (I can try add some next week)

michaelnchin:
@ltoniazzi actually, looking at the [ListEvents API](https://docs.aws.amazon.com/bedrock-agentcore/latest/APIReference/API_ListEvents.html) reference, `maxResults` has a hard upper limit of 100 events (defaults to 20 if we don't pass it). I think this is the reason for the original hardcoded limit. If the docs are correct, I think we need to do away with the no-limit `None` option.

ltoniazzi:
@michaelnchin Ok I get it, `maxResults` is the parameter for each `ListEvents` call, but it's unrelated to the pagination loop and it should not stop the pagination loop, right?

So I think `maxResults` can default/hardcoded to `<=100` as previously, but the previous `limit` parameter should still allow `None`, as it breaks the pagination.

michaelnchin:
@itoniazzi - if I'm understanding correctly, do we want two separate attributes here? `max_results` for the ListEvents API call parameter, and `limit` for the pagination limit?

ltoniazzi:
@michaelnchin I Think `limit` should be controlled and default to `None`.

`max_results`: I think both cases (hardcoded to be 100 or giving user control) are fine.

PR #738, opened 2025-10-29
feat(langgraph-checkpoint-aws): add Valkey checkpoint saver for AgentCore

seaofawareness:
This PR provides AgentCore-compatible checkpoint savers that use Valkey as the storage backend, combining AgentCore session management concepts with Valkey's storage and search capabilities.

### DIFF
--- a/samples/memory/agentcore_valkey_saver.ipynb
+++ b/samples/memory/agentcore_valkey_saver.ipynb
@@ -57,7 +57,7 @@
     "\n",
     "from langchain.chat_models import init_chat_model\n",
     "from langchain.tools import tool\n",
-    "from langgraph.prebuilt import create_react_agent\n",
+    "from langchain.agents import create_agent\n",
     "\n",
     "# Note: If you see a Pylance import error below, it's a false positive.\n",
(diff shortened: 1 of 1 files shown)


samples/memory/agentcore_valkey_checkpointer.ipynb
```diff
@@ -0,0 +1,938 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "import time\n",
+    "from typing import Dict, Any\n",
+    "\n",
+    "from langchain.chat_models import init_chat_model\n",
+    "from langchain.tools import tool\n",
+    "from langgraph.prebuilt import create_react_agent\n",
```
michaelnchin: `create_react_agent` is deprecated in LangGraph V1, we should be using `create_agent` instead (see docs here: https://docs.langchain.com/oss/python/releases/langchain-v1#create-agent)
```suggestion
    "from langchain.agents import create_agent\n",
```

@seaofawareness please replace all usages in this notebook, and regenerate the outputs to remove the deprecation warnings.

seaofawareness: Updated

Issue #731, opened 2025-10-28
System prompt parameter for ChatBedrockConverse


michaelnchin:
### Discussed in https://github.com/langchain-ai/langchain-aws/discussions/730

<div type='discussions-op-text'>

<sup>Originally posted by **Steinkreis** October 28, 2025</sup>
I think it would be beneficial to add a system_prompt parameter to the `ChatBedrockConverse` class, similar to how a system prompt can be defined for agents.

In the boto3 implementation, the client call works like this:

`client.converse(messages=bedrock_messages, system=system, **params)`

Here, the system message is handled separately from the rest of the messages. Allowing a `system_prompt` parameter would make the developer experience more consistent with using Bedrock directly.

This parameter could be implemented as an optional str, defaulting to None. That way, `ChatBedrockConverse` could be used in two ways:

1. Default behavior – the system prompt is provided via the messages list.
2. Parameter-based behavior – the system prompt is set using the system_prompt parameter, which is automatically added to every conversation.

I’ve looked into the source code, and with a small change to the _messages_to_bedrock function, this should be straightforward to implement. I’d also be happy to contribute this change.</div>

michaelnchin:
Hi @steinkreis! Thanks for opening the discussion - moved here for easier tracking.

Adding a `system` parameter to ChatBedrockConverse sounds like a good idea, given that it already supports equivalent parameters for most/all of the other Converse API options. Please do feel free to open a PR with the changes you have in mind👍

Steinkreis:
Hi @michaelnchin, i opened the PR [#743](https://github.com/langchain-ai/langchain-aws/pull/743)

PR #743, opened 2025-11-02
feat(aws): Add optional system parameter to ChatBedrockConverse

Steinkreis:
### Summary
related to [#731](https://github.com/langchain-ai/langchain-aws/issues/731)

This PR introduces a new optional `system_prompt` parameter to the `ChatBedrockConverse` class and the internal `_messages_to_bedrock` helper.

When provided, this parameter explicitly defines the system message used by the LLM.
If not provided, the system message continues to be inferred from the messages list, maintaining backward compatibility.

### Changes

- Added `system_prompt` field to `ChatBedrockConverse` with documentation.
- Updated `_messages_to_bedrock` to accept and prioritize the `system_prompt` parameter.
- `_messages_to_bedrock` is now always called with `self.system_prompt` inside `ChatBedrockConverse`
- Added new unit tests to verify correct handling and precedence of the system prompt parameter.

### Testing

- Added `test__messages_to_bedrock_prefers_parameter_over_system_message` to confirm the parameter takes precedence.
- Extended `test_model_kwargs` to validate initialization and storage of the system_prompt.
- Existing Tests passed

### Validation
Testing with the new parameter:
image

Testing the existing behaviour and backward compatibility:
image

### Potential Side Effects
When defining the `ChatBedrockConverse` as an LLM for an agent, the `system_prompt` parameter of `ChatBedrockConverse` will be prioritized:
image

Just in case if that would be a problem. I looked into the `create_agent` function. Right at the start it checks if `model` is a str:
```python
# init chat model
if isinstance(model, str):
    model = init_chat_model(model)
```

 and a possible solution could be to also handle the case where model is of type `BaseChatModel`:
```python
# init chat model
if isinstance(model, str):
    model = init_chat_model(model)
elif isinstance(model, BaseChatModel):
    # do something e.g. raise error or override system_prompt
```

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -371,11 +371,27 @@ class ChatBedrockConverse(BaseChatModel):
 
     """
 
-    system: Optional[List[str]] = None
-    """Optional system prompts for the LLM.
-    If provided, these prompts will be added as system messages,
-    and can be used with or without system prompts in the messages list.
+    system: Optional[List[Union[str, Dict[str, Any]]]] = None
+    """Optional list of system prompts for the LLM.
 
+    Each entry can be either:
+      - a simple string (for straightforward text-based system prompts), or
+      - a dictionary matching the Converse API system message schema, allowing
+        inclusion of additional fields like `guardContent`, `cachePoint`, etc.
+
+    Example:
+        system = [
+            "a simple system prompt",
+            {
+                "text": "another system prompt",
+                "guardContent": {"text": {"text": "string"}},
+                "cachePoint": {"type": "default"}
+            },
+        ]
+
+    String inputs will be internally converted to the appropriate message format,
+    while dict entries will be passed through as-is. Any invalid formats will be
+    rejected by the Converse API.
     """
 
     max_tokens: Optional[int] = None
@@ -1281,7 +1297,7 @@ class ChatBedrockConverse(BaseChatModel):
 
 def _messages_to_bedrock(
     messages: List[BaseMessage],
-    system: Optional[List[str]] = None,
+    system: Optional[List[Union[str, Dict[str, Any]]]] = None,
 ) -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]:
     """Handle Bedrock converse and Anthropic style content blocks"""
     for idx, message in enumerate(messages):
@@ -1305,7 +1321,13 @@ def _messages_to_bedrock(
     messages = merge_message_runs(trimmed_messages)
 
     if system:
-        bedrock_system.extend([{"text": s} for s in system])
+        sys_param_to_bedrock = []
+        for s in system:
+            if isinstance(s, str):
+                sys_param_to_bedrock.append({"text": s})
+            else:
+                sys_param_to_bedrock.append(s)
+        bedrock_system.extend(sys_param_to_bedrock)
 
     for msg in messages:
         content = _lc_content_to_bedrock(msg.content)
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -2071,15 +2071,49 @@ def test__messages_to_bedrock_preserves_whitespace_non_last_aimessage_blocks() -
 @pytest.mark.parametrize(
     "system_prompt_parameter, expected_system",
     [
-        (None, [{"text": "System message"}]),  # from messages list
+        # No system parameter → use only the system message from messages
+        (None, [{"text": "System message"}]),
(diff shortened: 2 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -371,6 +371,13 @@ class Joke(BaseModel):
 
     """
 
+    system_prompt: Optional[str] = None
```
michaelnchin: Two updates we should make to the parameter:
1) Rename to `system` to match the Converse API parameter name.
2) Update the type to list - `system` can accept multiple system prompts.
```suggestion
    system: Optional[list[str]] = None
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -1346,6 +1358,9 @@ def _messages_to_bedrock(
     if not bedrock_messages:
         bedrock_messages.append({"role": "user", "content": [{"text": EMPTY_CONTENT}]})
 
+    if system_prompt:
+        bedrock_system = [{"text": system_prompt}]
```
michaelnchin: Let's move this to before the `messages` processing loop. Since we can pass multiple prompts to `system` - if there are any additional `SystemMessage`s in `messages`, we should also include them in the final list.

Steinkreis:
> @Steinkreis thanks for putting this together!

Hey @michaelnchin, I implemented your suggestions and updated the tests accordingly

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -371,6 +371,13 @@ class Joke(BaseModel):
 
     """
 
+    system: Optional[List[str]] = None
```
michaelnchin: @Steinkreis sorry, one more requirement - `dict` type should also be accepted in the list, as `system` allows for `guardContent` and `cachePoint` blocks in addition to the base `text`.

Should look like:
```python
system: Optional[List[Union[str, Dict[str, Any]]]] = None
```

Full syntax from the boto3 Converse ref (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/converse.html):
```python
system=[
    {
        'text': 'string',
        'guardContent': {
            'text': {
                'text': 'string',
                'qualifiers': [
                    'grounding_source'|'query'|'guard_content',
                ]
            },
            'image': {
                'format': 'png'|'jpeg',
                'source': {
                    'bytes': b'bytes'
                }
            }
        },
        'cachePoint': {
            'type': 'default'
        }
    },
],
```

Steinkreis: I'm not entirely sure how I should implement this. How would you imagine `system` to look like for `ChatBedrockConverse`? Can you give an example?. I found the `_lc_content_to_bedrock` function that basically has the logic for transforming raw content into bedrock_format. But this expects blocks inside the content:
`for block in _snake_to_camel_keys(content)`
For example guardContent:
```python
        elif block["type"] == "guard_content":
            bedrock_content.append({"guardContent": {"text": {"text": block["text"]}}})
```
It seems off that for the system prompt the user only needs to provide a str but the rest should be handled differently.

michaelnchin: As I'm visualizing it, accepting `str` will just give the user an simpler way to pass text system prompts if the extra fields aren't needed.

An example input with both `str` and `dict` format would look like:
```python
system=[
    "a system prompt",
    {
        'text': 'another system prompt',
        'guardContent': {'text': {'text': 'string'}},
        'cachePoint': {'type': 'default'}
    },
]
```

We should continue handling/converting the `str` inputs as per your current implementation. The new `dict` inputs can be passed as-is without extra validation - we'll let the Converse API handle bouncing back an error to the user if the internal format is invalid.
```python
if system:
    sys_param_to_bedrock = []
    for s in system:
        if isinstance(s, str):
            sys_param_to_bedrock.append({"text": s})
        else:
            sys_param_to_bedrock.append(s)
    bedrock_system.extend(sys_param_to_bedrock)
```

Steinkreis: Gotcha! Thanks for the clarification 🙂

michaelnchin (approved):
Looks good - thanks for the changes @Steinkreis !

Steinkreis:
> Looks good - thanks for the changes @Steinkreis !

Thank you as well for the guidance @michaelnchin 🙂 Do you know when langchain-aws 1.0.1 will be released? I'm eagerly waiting for this feature 😄

michaelnchin:
Hi @Steinkreis, thanks for your patience - this feature is now available with langchain-aws [v1.1.0](https://github.com/langchain-ai/langchain-aws/releases/tag/langchain-aws%3D%3D1.1.0).

PR #755, opened 2025-11-10
fix: block forced tool_choice for Claude 4.5 thinking models and tests

jitokim:


libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -1239,7 +1239,8 @@ def bind_tools(
                     "claude-3-7-",
                     "claude-opus-4-",
                     "claude-sonnet-4-",
-                    "claude-haiku-4-",
+                    "claude-sonnet-4-5-",
+                    "claude-haiku-4-5-",
```
michaelnchin: This seems redundant with the existing checks for Claude 4-x model substrings? But the tests are probably useful to keep.
```suggestion
```

jitokim: @michaelnchin  thanks for review :)

PR #759, opened 2025-11-11
fix: Handling NS_SEP in langgraph correctly in session-id

ps06756:
Handling the "|" separator used by langgraph as "_" since session-id in bedrock doesn't support using "|" character


3coins (approved):
Looks good!

Would be good to add a unit test to avoid regression.

Issue #760, opened 2025-11-11
store with DynamoDB


5enxia:
### Package

langchain-checkpoint-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a feature request and not a bug report or question

### Feature Description

memory store with DynamoDB

### Use Case

There is already an implementation of store memory using Valkey, but there are cases where high read and write performance and vector store are not required.
I think it would be useful to have store memory using DynamoDB for such cases.

### Proposed Implementation (optional)

```python
from langgraph_checkpoint_aws import DynamoDBStore

# Basic usage
with DynamoDBStore.from_conn_string(
    "table-name",
) as store:
    # Setup vector search index
    store.setup()
    
    # Store documents
    store.put(
        ("documents", "user123"),
        "report_1",
        {
            "text": "Machine learning report on customer behavior analysis...",
            "tags": ["ml", "analytics", "report"],
            "author": "data_scientist"
        }
    )
    
    # get documents
    results = store.get(
        ("documents", "user123"),
        "report_1"
    )
```

#### related works

There is already an [implementation ](https://github.com/skamalj/langgraph_store_dynamodb)by skamalj, but it is unmaintained and support from the community would be greatly appreciated.

#### boto3 dynamodb

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb.html

#### LangGraph's BaseStore

https://github.com/langchain-ai/langgraph/blob/main/libs/checkpoint/langgraph/store/base/__init__.py

PR #764, opened 2025-11-14
feat(langgraph-checkpoint-aws): add DynamoDB store

5enxia:
This PR provides a DynamoDB-backed store implementation for LangGraph, enabling persistent, hierarchical key-value storage with features like TTL, filtering, and batch operations.

issue: https://github.com/langchain-ai/langchain-aws/issues/760

### DIFF
--- a/libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/test_async_dynamodb_store_integration.py
+++ b/libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/test_async_dynamodb_store_integration.py
@@ -9,12 +9,7 @@ import pytest
 
 from langgraph_checkpoint_aws.store.dynamodb import DynamoDBStore
 
-try:
-    import boto3
-
-    BOTO3_AVAILABLE = True
-except ImportError:
-    BOTO3_AVAILABLE = False
+from .conftest import BOTO3_AVAILABLE, DYNAMODB_AVAILABLE
 
 pytestmark = pytest.mark.skipif(
     not BOTO3_AVAILABLE,
@@ -23,39 +18,6 @@ pytestmark = pytest.mark.skipif(
 
 if BOTO3_AVAILABLE:
     from langgraph.store.base import Item, SearchItem
-
-
-def _is_dynamodb_available() -> bool:
-    """Check if a DynamoDB Local server is available for testing.
-
-    Verifies availability by making a real DynamoDB API call (list_tables)
-    with dummy credentials, rather than just checking if the port is open.
-    """
-    if not BOTO3_AVAILABLE:
-        return False
-
-    from .conftest import (
-        DYNAMODB_AWS_ACCESS_KEY_ID,
-        DYNAMODB_AWS_SECRET_ACCESS_KEY,
-        DYNAMODB_ENDPOINT_URL,
-        DYNAMODB_REGION,
-    )
-
-    try:
-        client = boto3.client(
-            "dynamodb",
-            region_name=DYNAMODB_REGION,
-            endpoint_url=DYNAMODB_ENDPOINT_URL,
-            aws_access_key_id=DYNAMODB_AWS_ACCESS_KEY_ID,
-            aws_secret_access_key=DYNAMODB_AWS_SECRET_ACCESS_KEY,
-        )
-        client.list_tables(Limit=1)
-        return True
-    except Exception:
-        return False
-
-
-DYNAMODB_AVAILABLE = _is_dynamodb_available()
 
 
 @pytest.fixture
--- a/libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/conftest.py
+++ b/libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/conftest.py
@@ -1,10 +1,16 @@
 """Shared fixtures and configuration for DynamoDB store integration tests."""
 
-from __future__ import annotations
-
 import os
 
 import pytest
+
+try:
+    import boto3
+
+    BOTO3_AVAILABLE = True
+except ImportError:
+    boto3 = None  # type: ignore[assignment]
+    BOTO3_AVAILABLE = False
 
 # Default configuration for DynamoDB Local
 DYNAMODB_HOST = os.getenv("DYNAMODB_HOST", "dynamodb-local")
@@ -27,6 +33,32 @@ def _set_aws_credentials(monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", DYNAMODB_AWS_SECRET_ACCESS_KEY)
     monkeypatch.setenv("AWS_SECURITY_TOKEN", "test")
     monkeypatch.setenv("AWS_SESSION_TOKEN", "test")
+
+
+def _is_dynamodb_available() -> bool:
+    """Check if a DynamoDB Local server is available for testing.
+
+    Verifies availability by making a real DynamoDB API call (list_tables)
+    with dummy credentials, rather than just checking if the port is open.
+    """
+    if not BOTO3_AVAILABLE:
+        return False
+
+    try:
+        client = boto3.client(
+            "dynamodb",
+            region_name=DYNAMODB_REGION,
+            endpoint_url=DYNAMODB_ENDPOINT_URL,
+            aws_access_key_id=DYNAMODB_AWS_ACCESS_KEY_ID,
+            aws_secret_access_key=DYNAMODB_AWS_SECRET_ACCESS_KEY,
+        )
+        client.list_tables(Limit=1)
+        return True
+    except Exception:
+        return False
+
+
+DYNAMODB_AVAILABLE = _is_dynamodb_available()
 
 
 @pytest.fixture
--- a/libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/test_dynamodb_store_integration.py
+++ b/libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/test_dynamodb_store_integration.py
@@ -11,12 +11,7 @@ import pytest
 from langgraph_checkpoint_aws.store.dynamodb import DynamoDBStore
 from langgraph_checkpoint_aws.store.dynamodb.exceptions import ValidationError
 
-try:
-    import boto3
-
-    BOTO3_AVAILABLE = True
-except ImportError:
-    BOTO3_AVAILABLE = False
+from .conftest import BOTO3_AVAILABLE, DYNAMODB_AVAILABLE
 
 pytestmark = pytest.mark.skipif(
     not BOTO3_AVAILABLE,
@@ -25,39 +20,6 @@ pytestmark = pytest.mark.skipif(
 
 if BOTO3_AVAILABLE:
     from langgraph.store.base import Item, SearchItem
-
-
-def _is_dynamodb_available() -> bool:
-    """Check if a DynamoDB Local server is available for testing.
-
-    Verifies availability by making a real DynamoDB API call (list_tables)
-    with dummy credentials, rather than just checking if the port is open.
-    """
-    if not BOTO3_AVAILABLE:
-        return False
-
-    from .conftest import (
-        DYNAMODB_AWS_ACCESS_KEY_ID,
-        DYNAMODB_AWS_SECRET_ACCESS_KEY,
-        DYNAMODB_ENDPOINT_URL,
-        DYNAMODB_REGION,
-    )
-
-    try:
-        client = boto3.client(
-            "dynamodb",
-            region_name=DYNAMODB_REGION,
-            endpoint_url=DYNAMODB_ENDPOINT_URL,
-            aws_access_key_id=DYNAMODB_AWS_ACCESS_KEY_ID,
-            aws_secret_access_key=DYNAMODB_AWS_SECRET_ACCESS_KEY,
-        )
-        client.list_tables(Limit=1)
-        return True
-    except Exception:
-        return False
-
-
-DYNAMODB_AVAILABLE = _is_dynamodb_available()
 
 
 @pytest.fixture
--- a/libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/test_dynamodb_sync_async_parity_integration.py
+++ b/libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/test_dynamodb_sync_async_parity_integration.py
@@ -11,50 +11,12 @@ import pytest
 
 from langgraph_checkpoint_aws.store.dynamodb import DynamoDBStore
 
-try:
-    import boto3
-
-    BOTO3_AVAILABLE = True
-except ImportError:
-    BOTO3_AVAILABLE = False
+from .conftest import BOTO3_AVAILABLE, DYNAMODB_AVAILABLE
 
 pytestmark = pytest.mark.skipif(
     not BOTO3_AVAILABLE,
     reason="boto3 not available. Install with: pip install boto3",
 )
-
-
-def _is_dynamodb_available() -> bool:
-    """Check if a DynamoDB Local server is available for testing.
-
-    Verifies availability by making a real DynamoDB API call (list_tables)
-    with dummy credentials, rather than just checking if the port is open.
-    """
-    if not BOTO3_AVAILABLE:
-        return False
-
-    from .conftest import (
-        DYNAMODB_AWS_ACCESS_KEY_ID,
-        DYNAMODB_AWS_SECRET_ACCESS_KEY,
-        DYNAMODB_ENDPOINT_URL,
-        DYNAMODB_REGION,
-    )
-
-    try:
-        client = boto3.client(
-            "dynamodb",
-            region_name=DYNAMODB_REGION,
-            endpoint_url=DYNAMODB_ENDPOINT_URL,
-            aws_access_key_id=DYNAMODB_AWS_ACCESS_KEY_ID,
-            aws_secret_access_key=DYNAMODB_AWS_SECRET_ACCESS_KEY,
-        )
-        client.list_tables(Limit=1)
-        return True
-    except Exception:
-        return False
-
-
-DYNAMODB_AVAILABLE = _is_dynamodb_available()
 
 
 @pytest.fixture


libs/langgraph-checkpoint-aws/docs/dynamodb_store.md
```diff
@@ -0,0 +1,265 @@
...
+Use **DynamoDB Store** when:
+- You need a fully managed solution
+- You don't require vector search capabilities
+- You want pay-per-request pricing
+- Your workload is unpredictable
+
+Use **Valkey Store** when:
+- You need vector search capabilities
+- You require ultra-low latency
+- You can manage your own infrastructure
```
seaofawareness: With ElastiCache valkey, infrastructure has shared management as self-designed clusters support vector search. Autoscaling support helps with unpredictable workloads as well. So both diff points are not relevant.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/__init__.py
```diff
@@ -8,8 +8,9 @@
 # Conditional imports for optional dependencies
 try:
     from .valkey import AsyncValkeyStore, ValkeyIndexConfig, ValkeyStore
+    from .dynamodb import DynamoDBStore
 
-    __all__ = ["AsyncValkeyStore", "ValkeyStore", "ValkeyIndexConfig"]
+    __all__ = ["AsyncValkeyStore", "ValkeyStore", "ValkeyIndexConfig", "DynamoDBStore"]
```
seaofawareness: nit: add in sorted order

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/base.py
```diff
@@ -0,0 +1,657 @@
...
+    - SK (Sort Key): Item key
+    - value: The stored dictionary
+    - created_at: ISO format timestamp
+    - updated_at: ISO format timestamp
+    - expires_at: Unix timestamp for TTL (optional)
+
+    Examples:
+        Basic usage:
+        ```python
+        from langgraph_checkpoint_aws.store.dynamodb import DynamoDBStore
```
seaofawareness: Remove .store as top level export allows use of class

libs/langgraph-checkpoint-aws/docs/dynamodb_store.md
5enxia: Thank you for review. I'll modify them.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/__init__.py
5enxia: OK. I'll sort order

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/DynamoDBStore.md
```diff
@@ -0,0 +1,259 @@
...
+pip install langgraph-checkpoint-aws
+```
+
+## Quick Start
+
+```python
+from langgraph_checkpoint_aws import DynamoDBStore
+
+# Create a store instance
+store = DynamoDBStore(table_name="my-store-table")
```
ltoniazzi: I think either `region_name` or `boto3_session` must be passed on init

5enxia: Thank you for your review!
I've modified the points you mentioned.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/base.py
```diff
@@ -0,0 +1,657 @@
...
+                UpdateExpression="SET expires_at = :expires_at, updated_at = :updated_at",
+                ExpressionAttributeValues={
+                    ":expires_at": expires_at,
+                    ":updated_at": datetime.now(timezone.utc).isoformat(),
+                },
+            )
+        except Exception as e:
+            logger.warning(f"Error refreshing TTL for {pk}/{sk}: {e}")
+
+    async def abatch(self, ops: Iterable[Op]) -> list[Result]:
```
ltoniazzi, line 759: I think one could run these in a thread, this should be already quite good as each sync method invokes the client once or twice, and would make these not blocking for parallel nodes and parallel graphs.

Let me know I can help add tests (something like [this](https://github.com/langchain-ai/langchain-aws/pull/776/files#diff-92dd6df4e7357e563228caf453632330be181482deb7d8745002e707f7d95735R666))

```python
from langchain_core.runnables import run_in_executor
import asyncio
...
async def abatch(self, ops: Iterable[Op]) -> list[Result]:
        """Execute batch operations asynchronously in parallel.

        This method leverages run_in_executor to execute synchronous DynamoDB
        operations concurrently in a thread pool. Each operation (GetOp, PutOp,
        SearchOp, ListNamespacesOp) is wrapped in an async function and executed
        in parallel using asyncio.gather, to improve throughput for
        batch operations compared to sequential execution. 
        TODO: should perform multiple dynamodb put_item with batch_write_item for better performance.

        Args:
            ops: Iterable of operations to execute. Supported operation types are
                GetOp, PutOp, SearchOp, and ListNamespacesOp.

        Returns:
            List of results corresponding to the input operations, in the same
            order as the input operations.

        Raises:
            NotImplementedError: If an unsupported operation type is encountered.
        """

        async def execute_op(op: Op) -> Result:
            """Execute a single operation in the executor."""
            if isinstance(op, GetOp):
                return await run_in_executor(None, self._batch_get_op, op)
            if isinstance(op, PutOp):
                return await run_in_executor(None, self._batch_put_op, op)
            if isinstance(op, SearchOp):
                return await run_in_executor(None, self._batch_search_op, op)
            if isinstance(op, ListNamespacesOp):
                return await run_in_executor(None, self._batch_list_namespaces_op, op)
            raise NotImplementedError(f"Operation type {type(op)} not supported")  # noqa: EM102

        # Execute all operations in parallel
        results = await asyncio.gather(*[execute_op(op) for op in ops])
        return list(results)
```

5enxia: Thank you for your feedback!
I implemented abatch based on your feedback. 
I also added tests.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/DynamoDBStore.md
```diff
@@ -0,0 +1,283 @@
...
+
+### DynamoDB Store vs Valkey Store
+
+| Feature | DynamoDB Store | Valkey Store |
+|---------|---------------|--------------|
+| Vector Search | ❌ No | ✅ Yes |
+| Performance | ✅ Good | ✅ Excellent |
+| Best For | cost-effective | Vector search, ultra-high performance |
+
+Use **DynamoDB Store** when:
```
ltoniazzi, line 344: Now there is no offloading to S3 for this DynamoDB Store right? (While it's the case for DynamoDBSaver)

So maybe mention the [Item limit of 400kb](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/bp-use-s3-too.html)?

(I would not add offloading to s3 in this PR to keep the PR simpler, but I am not a maintainer)

5enxia: Thank you for your feedback!

> Now there is no offloading to S3 for this DynamoDB Store right? (While it's the case for DynamoDBSaver)

Yes. In most cases, I think the size is sufficient for storing long-term memory.

> So maybe mention the Item limit of 400kb?

Good point. I agree that adding that information would be helpful.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/base.py
```diff
@@ -0,0 +1,702 @@
...
+        Make sure to call `setup()` before first use to create the necessary
+        DynamoDB table if it doesn't exist.
+
+    Warning:
+        DynamoDB charges are based on read/write capacity and storage.
+        Consider using on-demand billing for unpredictable workloads or
+        provisioned capacity for consistent traffic patterns.
+    """
+
+    MIGRATIONS: list[str] = []
```
ltoniazzi: I think `MIGRATIONS` is not used. What is it for?

libs/langgraph-checkpoint-aws/tests/unit_tests/store/dynamodb/test_dynamodb_store.py
```diff
@@ -0,0 +1,614 @@
...
+        
+        assert concurrent_time < max_expected_concurrent_time, (
+            f"Concurrent execution took {concurrent_time:.2f}s, "
+            f"expected < {max_expected_concurrent_time:.2f}s"
+        )
+
+    @pytest.mark.asyncio
+    async def test_abatch_get_concurrency(self, dynamodb_store):
+        """Test that abatch GetOp is non-blocking and runs concurrently."""
+        delay = 0.05
```
ltoniazzi: I think it's clearer if this `delay` variable is global and assigned once instead of in each method

libs/langgraph-checkpoint-aws/tests/unit_tests/store/dynamodb/test_dynamodb_store.py
```diff
@@ -0,0 +1,614 @@
...
+        
+        # If sequential, it would take at least n_async_calls * sleep_time
+        # We expect concurrent execution to be much faster, roughly sleep_time
+        # We'll allow a generous buffer (e.g. 2x sleep time + overhead) but definitely less than sequential
+        
+        # We assume the func has a sleep of at least 0.05s
+        # Total sequential time ~ n * 0.05
+        # Total concurrent time ~ 0.05 + overhead
+        
+        avg_single_call_time = 0.05
```
ltoniazzi: This can be the global delay variable

libs/langgraph-checkpoint-aws/tests/unit_tests/store/dynamodb/test_dynamodb_store.py
```diff
@@ -0,0 +1,614 @@
...
+        max_expected_concurrent_time = avg_single_call_time * n_async_calls * 0.5 # Should be significantly less than sequential
```
ltoniazzi: This should be `DELAY` (of single call) + a small overhead, lay `DELAY/2`

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/base.py
```diff
@@ -0,0 +1,702 @@
...
+    supports_ttl = True
```
ltoniazzi, line 110: This seems to also not be used?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/base.py
```diff
@@ -0,0 +1,702 @@
...
+        self.max_write_capacity_units = max_write_capacity_units or 10
+
+        # Initialize boto3 session and resources
+        if boto3_session:
+            self.session = boto3_session
+        else:
+            self.session = boto3.Session(region_name=region_name)
+
+        try:
+            self.dynamodb = self.session.resource("dynamodb")
```
ltoniazzi: I think it should use a `.client` instead of `.resource`, as [`client` is thread-safe](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/clients.html#multithreading-or-multiprocessing-with-clients).

Also should be able to pass `client_kwargs`, like in [this implementation for the saver](https://github.com/langchain-ai/langchain-aws/blob/main/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/utils.py#L119).

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/base.py
5enxia: It was a bit of trial and error.
I removed it.

libs/langgraph-checkpoint-aws/tests/unit_tests/store/dynamodb/test_dynamodb_store.py
5enxia: Thank you. I refactored it.

libs/langgraph-checkpoint-aws/tests/unit_tests/store/dynamodb/test_dynamodb_store.py
5enxia: Thank you. I refactored it too.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/base.py
5enxia: I would like to use the utils, but it's outside the scope of the module. What do you think? Should I define it separately?

ltoniazzi: I think it's a good idea, and it should always be reachable as would be part of `langgraph-checkpoint-aws` package no?


(Btw thanks 🙇  for this PR and sorry for the sporadic comments, but I'm using your implementation as custom code at the moment, so I'm giving comments as I find them, not doing a full review in one go)

5enxia: > I think it's a good idea, and it should always be reachable as would be part of langgraph-checkpoint-aws package no?

I may have caused some confusion.
What I meant to say is that within `langgraph-checkpoint-aws`, it might be better to create a separate directory like `utils` at the repository root or a similar location. This is because it spans both the store and the checkpointers.

Thank you. For now, I'll use the checkpointer's util.

5enxia: > (Btw thanks 🙇 for this PR and sorry for the sporadic comments, but I'm using your implementation as custom code at the moment, so I'm giving comments as I find them, not doing a full review in one go)

No problem.
I'm glad my work is needed.
I look forward to your feedback.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/__init__.py
```diff
@@ -26,6 +26,9 @@
         ValkeyIndexConfig,
         ValkeyStore,
     )
+    from langgraph_checkpoint_aws.store.dynamodb import (
+        DynamoDBStore,
+    )
```
michaelnchin: I think this should not be here since this try/catch is specifically for Valkey dependencies (unless there's some dependency I'm missing). Move this out next to the DynamoDBSaver import.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/__init__.py
```diff
@@ -7,9 +7,10 @@
 
 # Conditional imports for optional dependencies
 try:
+    from .dynamodb import DynamoDBStore
```
michaelnchin: Same as in `dynamodb/__init__.py` - should we move this handling out of the Valkey import try/catch and into its own try/catch statement?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/base.py
```diff
@@ -0,0 +1,799 @@
...
+        except Exception as e:
+            raise DynamoDBConnectionError(
+                f"Failed to initialize DynamoDB connection: {e}"
+            ) from e
+
+    @classmethod
+    @contextmanager
+    def from_conn_string(
+        cls,
+        table_name: str,
```
michaelnchin: I see this mirrors the same [class from ValkeyStore](https://github.com/langchain-ai/langchain-aws/blob/ef318092060e0e7e2324e90a4018e8e6babf51e6/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/valkey/store.py#L197-L199), but we use `table_name` instead of `conn_string` here so the name is no longer accurate - we should rename it to `from_table_name`

libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/test_async_dynamodb_store_integration.py
```diff
@@ -0,0 +1,296 @@
+"""Async integration tests for the DynamoDBStore implementation."""
+
+from __future__ import annotations
```
michaelnchin: Nit: shouldn't need this anymore since `langchain-aws` requires Python >=3.10 now

libs/langgraph-checkpoint-aws/tests/integration_tests/store/dynamodb/test_async_dynamodb_store_integration.py
```diff
@@ -0,0 +1,296 @@
...
+pytestmark = pytest.mark.skipif(
+    not BOTO3_AVAILABLE,
+    reason="boto3 not available. Install with: pip install boto3",
+)
+
+if BOTO3_AVAILABLE:
+    from langgraph.store.base import Item, SearchItem
+
+
+def _is_dynamodb_available() -> bool:
```
michaelnchin: I see `_is_dynamodb_available` is duplicated in all 3 of the integration test files - can we move it out to the shared `conftest.py` instead?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/__init__.py
5enxia: Thank you for your review!

5enxia:
@michaelnchin san
Thanks for your review !
I've fixed the issues pointed out in the reviews.

michaelnchin (commented):
Integ wf passes: https://github.com/langchain-ai/langchain-aws/actions/runs/22123849598

michaelnchin (approved):
LGTM - thanks @5enxia !

PR #774, opened 2025-11-22
fix: Store vectors as binary for FT index compatibility ValkeyStore

seaofawareness:
Store vectors as binary for FT index compatibility ValkeyStore

### DIFF
--- a/.github/workflows/_compile_integration_test.yml
+++ b/.github/workflows/_compile_integration_test.yml
@@ -26,7 +26,7 @@ jobs:
           - "3.13"
     name: "uv run pytest -m compile tests/integration_tests #${{ matrix.python-version }}"
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Set up Python ${{ matrix.python-version }} + uv
         uses: "./.github/actions/uv_setup"
--- a/.github/workflows/_integration_test.yml
+++ b/.github/workflows/_integration_test.yml
@@ -40,7 +40,7 @@ jobs:
       actions: write
     name: "make integration_test"
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           repository: "${{ inputs.fork }}/langchain-aws"
           ref: "${{ inputs.branch }}"
--- a/.github/workflows/_lint.yml
+++ b/.github/workflows/_lint.yml
@@ -33,7 +33,7 @@ jobs:
         python-version:
           - "3.12"
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Set up Python ${{ matrix.python-version }} + uv
         uses: "./.github/actions/uv_setup"
--- a/.github/workflows/_release.yml
+++ b/.github/workflows/_release.yml
@@ -40,7 +40,7 @@ jobs:
       version: ${{ steps.check-version.outputs.version }}
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Set up Python + uv
         uses: "./.github/actions/uv_setup"
@@ -95,7 +95,7 @@ jobs:
       contents: read
       id-token: write # Needed for aws-actions/configure-aws-credentials
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       # We explicitly *don't* set up caching here. This ensures our tests are
       # maximally sensitive to catching breakage.
@@ -231,7 +231,7 @@ jobs:
         working-directory: ${{ inputs.working-directory }}
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Set up Python + uv
         uses: "./.github/actions/uv_setup"
@@ -269,7 +269,7 @@ jobs:
         working-directory: ${{ inputs.working-directory }}
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Set up Python + uv
         uses: "./.github/actions/uv_setup"
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -26,7 +26,7 @@ jobs:
           - "3.13"
     name: "make test #${{ matrix.python-version }}"
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Set up Python ${{ matrix.python-version }} + uv
         uses: "./.github/actions/uv_setup"
--- a/.github/workflows/_test_release.yml
+++ b/.github/workflows/_test_release.yml
@@ -29,7 +29,7 @@ jobs:
       version: ${{ steps.check-version.outputs.version }}
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
 
       - name: Set up Python + uv
         uses: "./.github/actions/uv_setup"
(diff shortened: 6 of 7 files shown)


libs/langgraph-checkpoint-aws/tests/unit_tests/store/valkey/test_valkey_index_config_unit.py
```diff
@@ -743,7 +747,11 @@ def test_list_fields_handled_correctly(self, fake_valkey_client):
         hash_fields_str = {}
         for k, v in hash_fields.items():
             key_str = k.decode("utf-8") if isinstance(k, bytes) else k
-            val_str = v.decode("utf-8") if isinstance(v, bytes) else v
+            # Skip binary vector field - it's not UTF-8 text
+            if key_str == "vector" and isinstance(v, bytes):
+                val_str = v  # Keep as bytes
+            else:
+                val_str = v.decode("utf-8") if isinstance(v, bytes) else v
```
3coins: nit: Would be nice if this is not duplicated in multiple places.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/valkey/search_strategies.py
```diff
@@ -142,11 +144,28 @@ def _execute_vector_search(self, op: SearchOp) -> list[SearchItem]:
...
+                    index_name=index_name,
+                    index_operation="embedding_generation",
+                )
+
+            # Generate embedding for the query
+            if hasattr(self.store.embeddings, "embed_query"):
+                query_vector = self.store.embeddings.embed_query(op.query)
+            else:
+                # Fallback to async if only async method available
+                query_vector = asyncio.run(self.store.embeddings.aembed_query(op.query))
```
3coins: This is going to cause issues (RuntimeError) if there is an existing event loop. Is the valkey `aembed_query` method built on top of `embed_query` (like in LangChain) or has completely different implementation. Would be helpful to know how other implementations are handling this.

libs/langgraph-checkpoint-aws/tests/unit_tests/store/valkey/test_valkey_index_config_unit.py
seaofawareness: Fixed

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/valkey/search_strategies.py
seaofawareness: Added proper event loop detection with asyncio.get_running_loop()

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/valkey/search_strategies.py
```diff
@@ -142,11 +144,48 @@ def _execute_vector_search(self, op: SearchOp) -> list[SearchItem]:
...
+                        "Cannot generate embeddings: sync method not available "
+                        "and already in async context. Use AsyncValkeyStore instead.",
+                        index_name=index_name,
+                        index_operation="embedding_generation",
+                    )
+                except RuntimeError:
+                    # No running event loop, safe to create one
+                    query_vector = asyncio.run(
+                        self.store.embeddings.aembed_query(op.query)
+                    )
```
3coins: It's still not clear why we need all this, isn't this a configuration error if sync method is not present? How would this work in environments where there is an existing event loop? If we are relying on `AsyncValkeyStore` for handling the async version, would the right thing to do here is just raise that exception for the async case?

seaofawareness: We don't need this, simplified this flow to raise exception, so customers can only call sync embed from sync APIs.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/valkey/async_store.py
```diff
@@ -441,26 +442,23 @@ async def _generate_embeddings_async(self, op: PutOp) -> list[float] | None:
...
+            # Use async embeddings method
+            if hasattr(self.embeddings, "aembed_documents"):
+                vectors = await self.embeddings.aembed_documents(texts)
+                return vectors[0] if vectors else None
+            else:
+                raise EmbeddingGenerationError(
+                    "Cannot generate embeddings: embeddings object only has "
+                    "sync methods (embed_documents). "
+                    "Use ValkeyStore for sync embedding generation.",
+                    text_content=" ".join(texts[:3]) if texts else None,
```
3coins: Is this line relevant to the raised exception?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/valkey/search_strategies.py
3coins: There doesn't seem to be any change from the previous version here, did you apply all your changes?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/valkey/async_store.py
seaofawareness: Yes, this line is essential for providing meaningful error context when embedding generation fails due to sync/async method compatibility issues. This line provides the text_content parameter to the [EmbeddingGenerationError](https://github.com/langchain-ai/langchain-aws/blob/main/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/valkey/exceptions.py#L81) constructor. The exception class already truncates long text content for logging, so developer can understand content that was processed,
```
self.text_content = (
    text_content[:100] + "..."
    if text_content and len(text_content) > 100
    else text_content
)
```

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/valkey/search_strategies.py
seaofawareness: No changes here as this code is still needed as is. Changes are in async_store.py and store.py to throw EmbeddingGenerationError on sync/async mismatch rather than use other access pattern.

3coins: Correct me if I am wrong, but I see couple of issues with this approach:
1. It always favors sync method, even though an async method is available.
2. AsyncValkeyStore never uses the strategy in it's async paths, plus, it's sync method though present will raise NotImplementedError.

This means, that this code will never support async methods, so that whole block is redundant. Users should follow this for sync/async use cases.

Option 1: Use embeddings that have a sync embed_query() method (standard in LangChain)
```python
async def main():
    embeddings = BedrockEmbeddings()  # Has both embed_query and aembed_query
    store = ValkeyStore(client, embeddings=embeddings)
    results = store.search(...)  # Works - uses embed_query
```

Option 2: Use AsyncValkeyStore with asearch() (but note: vector search may not work currently)
```python
async def main():
    embeddings = BedrockEmbeddings()  # Has both embed_query and aembed_query
    store = AsyncValkeyStore(client, embeddings=embeddings)
    results = await store.asearch(...)  # Different code path entirely
```

seaofawareness: Thanks for catching this bug, during refactoring etc seems this was left out. I have refactored code to address this and also added integration test to help us catch regression from sync/async search differences.

3coins (approved):
@seaofawareness 
Thanks for making all the updates. Approving to move this forward as there are no API changes, but feel that the adapter approach is an overkill for handling combination of embeddings and store with async/sync methods in sync/async contexts. It is unusual for users to have these misconfigured, and a simpler approach would be preferred.

Issue #749, opened 2025-11-06
Feature Request: Async `aput_writes` for AgentCoreMemorySaver


ltoniazzi:
**Feature request**

I think that LangGraph nodes that run in parallel trigger, [upon completion](https://github.com/langchain-ai/langgraph/blob/9fd3dfc54223f082823dacd3c1e78ac0d9852122/libs/langgraph/langgraph/pregel/_runner.py#L104), writing with the checkpointer with [aput_writes](https://github.com/langchain-ai/langgraph/blob/9fd3dfc54223f082823dacd3c1e78ac0d9852122/libs/langgraph/langgraph/pregel/main.py#L2046). 

But `AgentCoreMemorySaver`'s `aput_writes` implementation falls back to the `boto3` sync client [here](https://github.com/langchain-ai/langchain-aws/blob/4062b71d3d213867046bb3c84958fb2ff279a75b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py#L307).

So I think this makes the writes blocking for parallel nodes.

If correct, is it possible to get the async implementation?

michaelnchin:
cc: @jgordley @bergjaak

PR #776, opened 2025-11-23
feat(langgraph-checkpoint-aws): Run AgentCoreSaver async methods with executor

ltoniazzi:
## Proposal

Currently `AgentCoreSaver` async methods fall back to the sync ones. 

But this can cause blocking i/o calls, for example when parallel nodes trigger `aput_writes` (which I think runs async like in [here](https://github.com/langchain-ai/langgraph/blob/d45f52ab0cac3d036c960cb344095294b537c3fe/libs/langgraph/langgraph/pregel/_loop.py#L370), using [this](https://github.com/langchain-ai/langgraph/blob/d45f52ab0cac3d036c960cb344095294b537c3fe/libs/langgraph/langgraph/pregel/_executor.py#L142)).

By sending the sync methods to another thread we solve the above problems, and we should be already quite efficient as the sync methods mostly perform a single sync call to AgentCore Memory.

## Changes

1. Wrap sync methods in `run_in_executor` for async versions
2. Add two tests for each method, (a) assert the sync method is called correctly, (b) assert the total time of multiple async execution is equal to a single one.

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
@@ -4,6 +4,7 @@ AgentCore Memory Checkpoint Saver implementation.
 
 from __future__ import annotations
 
+import asyncio
 import random
 from collections.abc import AsyncIterator, Iterator, Sequence
 from typing import Any, TypeAlias, cast
@@ -285,9 +286,13 @@ class AgentCoreMemorySaver(BaseCheckpointSaver[str]):
         before: RunnableConfig | None = None,
         limit: int | None = None,
     ) -> AsyncIterator[CheckpointTuple]:
-        for item in await run_in_executor(
-            None, self.list, config, filter=filter, before=before, limit=limit
-        ):
+        loop = asyncio.get_running_loop()
+
+        def _sync_list():
+            return list(self.list(config, filter=filter, before=before, limit=limit))
+
+        items = await loop.run_in_executor(None, _sync_list)
+        for item in items:
             yield item
 
     async def aput(
--- a/libs/langgraph-checkpoint-aws/tests/unit_tests/agentcore/test_saver.py
+++ b/libs/langgraph-checkpoint-aws/tests/unit_tests/agentcore/test_saver.py
@@ -5,6 +5,7 @@ Unit tests for AgentCore Memory Checkpoint Saver.
 import asyncio
 import json
 import time
+from collections.abc import Iterator
 from unittest.mock import ANY, MagicMock, Mock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


ltoniazzi:
I fixed the title, so now the title test should pass.

ltoniazzi:
@michaelnchin Is it possible to get a review for this PR?

ltoniazzi:
Resolving https://github.com/langchain-ai/langchain-aws/issues/749

michaelnchin:
Hi @ltoniazzi! Sorry for the delay, just back from vacation. Thanks for submitting this PR - will have a look soon

michaelnchin:
cc: @bergjaak @jgordley

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
```diff
@@ -285,7 +285,9 @@ async def alist(
         before: RunnableConfig | None = None,
         limit: int | None = None,
     ) -> AsyncIterator[CheckpointTuple]:
-        for item in self.list(config, filter=filter, before=before, limit=limit):
+        for item in await run_in_executor(
+            None, self.list, config, filter=filter, before=before, limit=limit
+        ):
```
michaelnchin: The other async method implementations with `run_in_executor` look good. But for `alist`, I don’t think the new implementation offers much over the original, as we’re only using `run_in_executor` to create the generator object from `self.list()`. The outer `for` loop still iterates said generator in the main thread instead of the executor, so true async is still getting blocked here.

We should use something similar to the [AgentCoreValkeySaver’s alist implementation](https://github.com/langchain-ai/langchain-aws/blob/6ae571c73131c9d61c1c02eb30a9874cb7ff5631/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/valkey/saver.py#L672-L691), which does finish iterating the generator in a subthread and into a list object before yielding the results.

```suggestion
        import asyncio

        loop = asyncio.get_running_loop()
    
        def _sync_list():
            return list(self.list(config, filter=filter, before=before, limit=limit))
    
        items = await loop.run_in_executor(None, _sync_list)
        for item in items:
            yield item
```

libs/langgraph-checkpoint-aws/tests/unit_tests/agentcore/test_saver.py
```diff
@@ -145,6 +171,56 @@ def sample_checkpoint_metadata(self):
...
+
+    @pytest.fixture
+    def mock_slow_list(self, sample_checkpoint_tuple):
+        """Mock list with artificial delay for testing async concurrency."""
+
+        def _mock_slow_list(config, *, filter=None, before=None, limit=None):  # noqa: ARG001 A002
+            time.sleep(MOCK_SLEEP_DURATION)
+            return [sample_checkpoint_tuple]
+
+        return _mock_slow_list
```
michaelnchin, line 198: Following on the above comments about the `alist` implementation - the mock is incorrectly assuming that `self.list()` returns a list, when it actually returns a generator instead. Need to fix this - the below should fail with the current `alist` implementation, and succeed with the new one.

```suggestion
    @pytest.fixture
def mock_slow_list(self, sample_checkpoint_tuple):
    """Mock list that returns a generator, with delay."""
    
    def _mock_slow_list(config, *, filter=None, before=None, limit=None):
        def _generator():
            time.sleep(MOCK_SLEEP_DURATION)
            yield sample_checkpoint_tuple
        return _generator()
    
    return _mock_slow_list

```

ltoniazzi: Thanks a lot, fixed both

Issue #786, opened 2025-12-03
First class support for AWS Bedrock service_tier in the langchain-aws chat models


adamhadani:
### Package

langchain-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a feature request and not a bug report or question

### Feature Description

AWS Bedrock recently added support for service tiers to AWS Bedrock models (see for example [here](https://aws.amazon.com/bedrock/service-tiers/)).

We should support this.

### Use Case

This feature would allow users to use LLMs from AWS Bedrock with the different service tiers available.

### Proposed Implementation (optional)

```python
we should support this in the langchain-aws, ideally via a `service_tier` parameter that can be passed to the chat clients, e.g.:


llm = ChatBedrockConverse(
    ...
    service_tier="priority",
)
```

### Additional Context

the aws client classes support receiving a `config` that is passed as is (?) to the botocore. perhaps there is a way to use that for setting a service tier, but I could not find any documentation anywhere how to do that (maybe not supported by boto at all yet?).

Note also this is distinct from the (existing) `performance_config` / optimized latency feature (see [here](https://docs.aws.amazon.com/bedrock/latest/userguide/latency-optimized-inference.html)) which seems to be already supported

facundofleitas:
I need this feature, please!

adamhadani:
Added in #795 FYI @michaelnchin @facundofleitas

PR #795, opened 2025-12-09
feat(aws): Add support for AWS Bedrock serviceTier

adamhadani:
Addresses Issue #786 . See reference docs for AWS / boto [here](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-invoke.html).

- Add support for `serviceTier` parameter when invoking AWS Bedrock models.
- Add unit test coverage
- Bump `boto3` and related dependencies to latest version needed for `serviceTier` support, update `uv.lock` and `pyproject.toml` accordingly

### DIFF
--- a/libs/aws/langchain_aws/document_compressors/rerank.py
+++ b/libs/aws/langchain_aws/document_compressors/rerank.py
@@ -126,6 +126,10 @@ class BedrockRerank(BaseDocumentCompressor):
         if len(documents) == 0:
             return []
 
+        effective_top_n = top_n if top_n is not None else self.top_n
+        if effective_top_n is not None:
+            effective_top_n = min(effective_top_n, len(documents))
+
         # Serialize documents for the Bedrock API
         serialized_documents = [
             {"textDocument": {"text": doc.page_content}, "type": "TEXT"}
@@ -145,7 +149,7 @@ class BedrockRerank(BaseDocumentCompressor):
                         "additionalModelRequestFields": additional_model_request_fields
                         or {},
                     },
-                    "numberOfResults": top_n or self.top_n,
+                    "numberOfResults": effective_top_n,
                 },
                 "type": "BEDROCK_RERANKING_MODEL",
             },
--- a/libs/aws/tests/unit_tests/document_compressors/test_rerank.py
+++ b/libs/aws/tests/unit_tests/document_compressors/test_rerank.py
@@ -69,3 +69,89 @@ def test_compress_documents(mock_rerank: MagicMock, reranker: BedrockRerank) -> 
     assert len(compressed_docs) == 2
     assert compressed_docs[0].metadata["relevance_score"] == 0.95
     assert compressed_docs[1].metadata["relevance_score"] == 0.85
+
+
+@patch("langchain_aws.utils.create_aws_client")
+def test_rerank_clamps_top_n_to_document_count(
+    mock_create_client: MagicMock,
+) -> None:
+    mock_client = MagicMock()
+    mock_create_client.return_value = mock_client
+
+    captured_request: dict = {}
+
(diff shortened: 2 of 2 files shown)


libs/aws/pyproject.toml
```diff
@@ -13,7 +13,7 @@ version = "1.1.0"
 requires-python = ">=3.10"
 dependencies = [
     "langchain-core>=1.1.0",
-    "boto3>=1.40.19",
+    "boto3>=1.42.5",
```
adamhadani, line 16: newer `boto3` / `botocore` required to get the `serviceTier` parameter to be supported by boto

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -1227,6 +1247,7 @@ def _converse_params(
                 ),
                 "guardrailConfig": guardrailConfig or self.guardrail_config,
                 "performanceConfig": performanceConfig or self.performance_config,
+                "serviceTier": {"type": tier} if tier else None,
```
adamhadani, line 1250: ^ `botocore` expects the tier to be given as a small struct of the form `{type: "foo"}`  for bedrock converse

adamhadani:
fixed typing / mypy errors FYI @michaelnchin

michaelnchin:
Thanks @adamhadani - could you rebase once more on main? Just merged a different change also updating uv.lock and it's causing some conflicts

adamhadani:
@michaelnchin done

adamhadani:
all tests pass cc/ @michaelnchin

libs/aws/langchain_aws/llms/bedrock.py
```diff
@@ -1070,13 +1085,16 @@ def _prepare_input_and_invoke(
...
-        request_options = {
+        request_options: dict[str, Any] = {
             "body": body,
             "modelId": self.model_id,
             "accept": accept,
             "contentType": contentType,
         }
 
+        if self.service_tier:
+            request_options["serviceTier"] = {"type": self.service_tier}
```
michaelnchin: Per Boto3 reference for [invoke_model](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model.html), `serviceTier` expected format is slightly different from that of Converse:
```
serviceTier='priority'|'default'|'flex'|'reserved'
```
So I think this should instead be:
```suggestion
            request_options["serviceTier"] = self.service_tier
```

libs/aws/langchain_aws/llms/bedrock.py
```diff
@@ -1216,13 +1234,16 @@ def _prepare_input_and_invoke_stream(
...
+        request_options: dict[str, Any] = {
             "body": body,
             "modelId": self.model_id,
             "accept": "application/json",
             "contentType": "application/json",
         }
 
+        if self.service_tier:
+            request_options["serviceTier"] = {"type": self.service_tier}
+
```
michaelnchin: Ditto as for `invoke_model`, [`invoke_model_with_response_stream`](invoke_model_with_response_stream) also expects tier as str input:
```suggestion
            request_options["serviceTier"] = self.service_tier
```

michaelnchin (changes requested):

libs/aws/langchain_aws/llms/bedrock.py
adamhadani: thanks @michaelnchin , updated here and and in the other place to use plain `str` for these APIs. good catch - I did some local testing with the converse API (which worked), not with these ones.

libs/aws/langchain_aws/llms/bedrock.py
adamhadani: addressed

PR #822, opened 2025-12-29
feat(aws): add sync multimodal embedding support to BedrockEmbeddings

### LINKED ISSUE #797: feat(aws): add multimodal embeddings support for BedrockEmbeddings
### Package

langchain-aws

### Checked other resources

### Feature Description

Add multimodal embedding support (image, audio, video) to `BedrockEmbeddings`, with model capability validation. Currently only text embeddings are supported.

Models with multimodal capabilities: 
- amazon.titan-embed-image-v1 - text, image
- amazon.nova-2-multimodal-embeddings-v1:0 - text, image, audio, video
- cohere.embed-v4:0 (and v3) - text, image
- twelvelabs.marengo-embed-3-0-v1:0 (and 2.7) - text, image, audio, video

### Use Case

Multimodal embeddings enable semantic search across images, audio, and video content - use cases across search involving media files, document images, or video libraries. Users currently cannot embed non-text content without bypassing BedrockEmbeddings entirely.

```python
Considerations
- Validate modality is supported for the given model before making API call
- Some models support text and image interleaved
- Audio/video may require async invocation (start_async_invoke) for some models
- S3 URIs required for large media files (Marengo, Nova async)
- Different dimension sizes supported per model
```

### Additional Context

Issues:
- #791 
- #796 
- #773 
- #744 
- #680  

Docs:
  - https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-titan-embed-mm.html
  - https://docs.aws.amazon.com/nova/latest/userguide/embeddings-schema.html
  - https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed-v4.html
  - https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed-v3.html
  - https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-marengo-3.html

johnkitaoka:
Add multimodal embeddings for supported Bedrock models for sync API including tests. 

Closes #797 

Includes:
- Input validation, can be file path/s3 URI/b64 encoding
- dimensions and normalization for supported models
- support for all available multimodal Bedrock embeddings models: Cohere v4 + v3, Nova Embeddings, Titan Multimodal Embeddings, Marengo models
- media format detection

Does not include:
- async API support (Marengo video embedding, Nova embedding for audio/video >= 30 seconds)
- bulk/array embedding optimization (Cohere v4 supports up to 96 in a single request)   
- interleaved text and image inputs (Cohere v4)
- kwargs support for specific parameters controlling embeddings output types, embedding purpose (embeddingPurpose for Nova, input_type for Cohere), truncation, detail (for images in Nova)
- media validation (size, duration checks)


libs/aws/langchain_aws/embeddings/bedrock.py
```diff
@@ -307,6 +343,210 @@ def _invoke_model(self, input_body: Dict[str, Any] = {}) -> Dict[str, Any]:
...
+        if data[:4] == b"OggS":
+            return "ogg"
+
+        return "jpeg"  # default fallback
+
+    def _load_media(
+        self,
+        data: MediaInput,
+        media_type: Literal["image", "audio", "video"],
+    ) -> tuple:
```
michaelnchin: Nit: should use a more specific type annotation here
```suggestion
    ) -> tuple[str, str, Literal["inline", "s3"]]:
```

libs/aws/langchain_aws/embeddings/bedrock.py
```diff
@@ -307,6 +343,210 @@ def _invoke_model(self, input_body: Dict[str, Any] = {}) -> Dict[str, Any]:
             logger.exception("Error raised by inference endpoint")
             raise e
 
+    def _detect_media_format(self, data: bytes) -> str:
+        """Detect media format from magic bytes."""
+        if len(data) < 12:
+            return "jpeg"  # default fallback
```
michaelnchin: Optional: How confident are we about the data always being JPEG in this case? I think we should log a warning here to let the user know how we handled an ambiguous media format (and maybe also before the final return statement)

libs/aws/langchain_aws/embeddings/bedrock.py
```diff
@@ -398,6 +638,102 @@ async def aembed_documents(self, texts: List[str]) -> List[List[float]]:
 
         return list(result)
 
+    def embed_image(self, image: MediaInput) -> List[float]:
```
michaelnchin, line 649: General suggestion for all of the new `embed_image(s)`, `embed_audio`, and `embed_video` methods: 

The existing `embed_query` and `embed_documents` methods have [equivalent async implementations](https://github.com/johnkitaoka/langchain-aws/blob/feat/multimodal-embeddings/libs/aws/langchain_aws/embeddings/bedrock.py#L613-L639). For consistency, can we also add async methods for the new ones in this PR?

Ex.
```python
async def aembed_image(self, image: MediaInput) -> List[float]:
    """Asynchronously embed a single image."""
    return await run_in_executor(None, self.embed_image, image)

async def aembed_images(self, images: List[MediaInput]) -> List[List[float]]:
    """Asynchronously embed multiple images."""
    result = await asyncio.gather(*[self.aembed_image(img) for img in images])
    return list(result)
...
```

libs/aws/langchain_aws/embeddings/bedrock.py
johnkitaoka: Thanks for the call out! I had this as a default since the API will return an error if the format is incompatible, but adding logs is a good idea

libs/aws/langchain_aws/embeddings/bedrock.py
johnkitaoka: Updated with new functions `embed_image`, `embed_images`, `embed_audio`, `embed_video ` . Will make a new PR for full async API support for Nova embeddings and Marengo for audio/video files!

johnkitaoka:
Hi @michaelnchin ! I tested the Cohere v4 embeddings model w/ the test image via boto3 and it failed, so I created a new test image w/ imagemagick and it worked - updated the test with new image and it works across all models!

michaelnchin:
Thanks @johnkitaoka ! Confirmed that CI is now green: https://github.com/langchain-ai/langchain-aws/actions/runs/20867904101

PR #828, opened 2026-01-04
feat(langgraph-checkpoint-aws): Add browser and code interpreter tools enhancements for Bedrock AgentCore

sundargthb:
Adds LangChain tools for AWS Bedrock AgentCore Browser and Code Interpreter services.

### Browser Toolkit
- Navigate, click, extract text/links, get elements, current page
- New tools: type_text, screenshot, scroll, wait_for_element
- Thread-aware session management via BrowserSessionManager

### Code Interpreter Toolkit  
- Execute code (Python/JS/TS), shell commands, file operations
- New tools: upload_file, install_packages
- Thread-aware session management per thread_id

## Dependencies

Requires `bedrock-agentcore` SDK: [[link to SDK PR]](https://github.com/aws/bedrock-agentcore-sdk-python/pull/202)
The SDK PR must be merged and released before this PR can be merged.

## Testing

- [x] Browser toolkit sample notebook
- [x] Code interpreter sample notebook

### DIFF
--- a/libs/aws/langchain_aws/tools/browser_tools.py
+++ b/libs/aws/langchain_aws/tools/browser_tools.py
@@ -88,10 +88,12 @@ class TypeTextInput(BaseModel):
 class ScreenshotInput(BaseModel):
     """Input for ScreenshotTool."""
 
-    full_page: bool = Field(
-        default=False,
-        description="If True, captures the entire scrollable page. "
-        "If False (default), captures only the visible viewport.",
+    capture_type: Literal["viewport", "full_page"] = Field(
+        default="viewport",
+        description=(
+            "Type of screenshot: 'viewport' (visible area, default) "
+            "or 'full_page' (entire scrollable page)"
+        ),
     )
 
 
@@ -542,7 +544,9 @@ class ThreadAwareExtractHyperlinksTool(ThreadAwareBaseTool):
             for link in soup.find_all("a", href=True):
                 text = link.get_text().strip()
                 href = link["href"]
-                if href.startswith("http") or href.startswith("https"):
+                if isinstance(href, str) and (
+                    href.startswith("http") or href.startswith("https")
+                ):
                     links.append({"text": text, "url": href})
 
             self.release_sync_browser(thread_id)
@@ -589,7 +593,9 @@ class ThreadAwareExtractHyperlinksTool(ThreadAwareBaseTool):
             for link in soup.find_all("a", href=True):
                 text = link.get_text().strip()
                 href = link["href"]
-                if href.startswith("http") or href.startswith("https"):
+                if isinstance(href, str) and (
+                    href.startswith("http") or href.startswith("https")
+                ):
                     links.append({"text": text, "url": href})
 
             await self.release_async_browser(thread_id)
@@ -839,7 +845,7 @@ the entire scrollable page, or False (default) for just the visible viewport."""
 
     def _run(
         self,
-        full_page: bool = False,
+        capture_type: str = "viewport",
         config: Optional[RunnableConfig] = None,
         **_: Any,
     ) -> str:
@@ -847,6 +853,7 @@ the entire scrollable page, or False (default) for just the visible viewport."""
         try:
             thread_id = self.get_thread_id(config)
             page = self.get_sync_page(thread_id)
+            full_page = capture_type == "full_page"
             screenshot_bytes = page.screenshot(full_page=full_page, type="png")
             screenshot_base64 = base64.b64encode(screenshot_bytes).decode("utf-8")
             self.release_sync_browser(thread_id)
@@ -869,7 +876,7 @@ the entire scrollable page, or False (default) for just the visible viewport."""
 
     async def _arun(
         self,
-        full_page: bool = False,
+        capture_type: str = "viewport",
         config: Optional[RunnableConfig] = None,
         **_: Any,
     ) -> str:
@@ -877,6 +884,7 @@ the entire scrollable page, or False (default) for just the visible viewport."""
         try:
             thread_id = self.get_thread_id(config)
             page = await self.get_async_page(thread_id)
+            full_page = capture_type == "full_page"
             screenshot_bytes = await page.screenshot(full_page=full_page, type="png")
             screenshot_base64 = base64.b64encode(screenshot_bytes).decode("utf-8")
             await self.release_async_browser(thread_id)
--- a/libs/aws/pyproject.toml
+++ b/libs/aws/pyproject.toml
@@ -36,7 +36,7 @@ test = [
     "langchain-classic>=1.0.0",
     "langchain-tests>=1.1.2",
     "langchain>=1.0.0rc2",
-    "bedrock-agentcore>=0.1.0; python_version>='3.10'",
+    "bedrock-agentcore>=1.1.4; python_version>='3.10'",
     "playwright>=1.53.0",
     "beautifulsoup4>=4.13.4",
 ]
--- a/libs/aws/tests/unit_tests/tools/test_browser_toolkit.py
+++ b/libs/aws/tests/unit_tests/tools/test_browser_toolkit.py
@@ -1,9 +1,14 @@
 """Unit tests for browser toolkit and tools."""
 
+import pytest
+
+# Skip all tests in this module if optional dependencies are not installed
+pytest.importorskip("bedrock_agentcore", reason="Requires langchain-aws[tools]")
+pytest.importorskip("playwright", reason="Requires langchain-aws[tools]")
+
 from typing import cast
 from unittest.mock import AsyncMock, MagicMock, patch
 
-import pytest
 from langchain_core.runnables.config import RunnableConfig
 from langchain_core.tools import BaseTool
 
@@ -284,7 +289,7 @@ class TestBrowserToolInputSchemas:
         from langchain_aws.tools.browser_tools import ScreenshotInput
 
         input_data = ScreenshotInput()
-        assert input_data.full_page is False
+        assert input_data.capture_type == "viewport"
 
     def test_scroll_input_defaults(self) -> None:
         """Test ScrollInput default values."""
--- a/libs/aws/tests/unit_tests/tools/test_code_interpreter_toolkit.py
+++ b/libs/aws/tests/unit_tests/tools/test_code_interpreter_toolkit.py
@@ -1,9 +1,13 @@
 """Unit tests for code interpreter toolkit and tools."""
 
+import pytest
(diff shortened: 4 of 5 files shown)


libs/aws/langchain_aws/tools/browser_tools.py
```diff
@@ -688,6 +733,344 @@ async def _arun(
...
+            if thread_id:
+                try:
+                    self.release_sync_browser(thread_id)
+                except Exception as release_error:
+                    logger.warning(f"Error releasing browser: {release_error}")
+            raise ToolException(f"Error taking screenshot: {str(e)}")
+
+    async def _arun(
+        self,
+        full_page: bool = False,
```
3coins: Browsers also have selected area capture option. Is there a possibility of other screen capture options that would deprecate this boolean approach?

libs/aws/langchain_aws/tools/browser_tools.py
```diff
@@ -688,6 +733,344 @@ async def _arun(
...
+        **_: Any,
+    ) -> str:
+        thread_id = None
+        try:
+            thread_id = self.get_thread_id(config)
+            page = await self.get_async_page(thread_id)
+            screenshot_bytes = await page.screenshot(full_page=full_page, type="png")
+            screenshot_base64 = base64.b64encode(screenshot_bytes).decode("utf-8")
+            await self.release_async_browser(thread_id)
+            page_type = "full page" if full_page else "viewport"
```
3coins, line 891: Is there a need for this transform?

libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
```diff
@@ -229,7 +403,7 @@ def _execute_code(
         code: str,
         config: RunnableConfig,
         language: str = "python",
-        clearContext: bool = False,
+        clear_context: bool = False,
```
3coins, line 406: Update the docstring to match the name change.

libs/aws/langchain_aws/tools/browser_tools.py
```diff
@@ -57,6 +58,50 @@ class CurrentWebPageToolInput(BaseModel):
...
+    )
+
+
+class ScrollInput(BaseModel):
+    """Input for ScrollTool."""
+
+    direction: str = Field(
+        default="down", description="Scroll direction: 'up', 'down', 'left', or 'right'"
+    )
+    amount: int = Field(default=500, description="Number of pixels to scroll")
```
3coins: Is this expected to be positive int? Should we add those constraints?
```python
Field(default=500, description="...", gt=0)
```

libs/aws/langchain_aws/tools/browser_tools.py
```diff
@@ -57,6 +58,50 @@ class CurrentWebPageToolInput(BaseModel):
...
+
+class WaitForElementInput(BaseModel):
+    """Input for WaitForElementTool."""
+
+    selector: str = Field(description="CSS selector for the element to wait for")
+    timeout: int = Field(
+        default=30000,
+        description="Maximum time to wait in milliseconds "
+        "(default: 30000ms = 30 seconds)",
+    )
```
3coins, line 120:
```suggestion
    timeout: int = Field(
        default=30000,
        description="Maximum time to wait in milliseconds "
        "(default: 30000ms = 30 seconds)",
        gt=0,
        le=300000
    )
```

libs/aws/langchain_aws/tools/browser_tools.py
```diff
@@ -57,6 +58,50 @@ class CurrentWebPageToolInput(BaseModel):
...
+        "If False (default), captures only the visible viewport.",
+    )
+
+
+class ScrollInput(BaseModel):
+    """Input for ScrollTool."""
+
+    direction: str = Field(
+        default="down", description="Scroll direction: 'up', 'down', 'left', or 'right'"
+    )
```
3coins: Should we make this a Literal?

```suggestion
    direction: Literal["up", "down", "left", "right"] = Field(
        default="down", description="Scroll direction"
    )
```

libs/aws/langchain_aws/tools/browser_tools.py
```diff
@@ -57,6 +58,50 @@ class CurrentWebPageToolInput(BaseModel):
...
+    selector: str = Field(description="CSS selector for the element to wait for")
+    timeout: int = Field(
+        default=30000,
+        description="Maximum time to wait in milliseconds "
+        "(default: 30000ms = 30 seconds)",
+    )
+    state: str = Field(
+        default="visible",
+        description="State to wait for: 'attached', 'detached', 'visible', or 'hidden'",
+    )
```
3coins:
```suggestion
    state: Literal["attached", "detached", "visible", "hidden"] = Field(
        default="visible",
        description="State to wait for",
    )
```

libs/aws/langchain_aws/tools/browser_tools.py
```diff
@@ -688,6 +733,344 @@ async def _arun(
...
+                delta_y = -amount
+            elif direction == "right":
+                delta_x = amount
+            elif direction == "left":
+                delta_x = -amount
+            else:
+                raise ValueError(
+                    f"Invalid direction: {direction}. "
+                    "Use 'up', 'down', 'left', or 'right'."
+                )
```
3coins: minor: There is some duplication around this in _arun, which can be extracted to a separate method.

libs/aws/langchain_aws/tools/browser_tools.py
```diff
@@ -688,6 +733,344 @@ async def _arun(
...
+- Waiting for elements to become visible after animations
+- Ensuring an element exists before interacting with it
+
+States:
+- 'visible': Element is present and visible (default)
+- 'attached': Element is in the DOM (may be hidden)
+- 'detached': Element is removed from the DOM
+- 'hidden': Element is present but not visible
+
+Timeout is in milliseconds (default: 30000 = 30 seconds)."""
```
3coins: Is this level of detail needed?

libs/aws/langchain_aws/tools/browser_tools.py
```diff
@@ -688,6 +733,344 @@ async def _arun(
...
+    description: str = """Scroll the webpage in the specified direction.
+
+Use this tool to:
+- Load more content on infinite scroll pages
+- Navigate to different sections of a long page
+- Bring elements into view before interacting with them
+- Scroll through search results or lists
+
+Directions: 'up', 'down', 'left', 'right'
+Amount: pixels to scroll (default 500)"""
```
3coins: Is this level of detail needed?

samples/tools/bedrock_agentcore_browser.ipynb
```diff
@@ -114,27 +114,41 @@
...
+    "- click_element: Click on an element\n",
+    "- navigate_back: Go back to previous page\n",
+    "- extract_text: Extract text content from page\n",
+    "- extract_hyperlinks: Extract links from page\n",
+    "- get_elements: Get elements matching a selector\n",
+    "- current_webpage: Get current page URL and title\n",
+    "- type_text: Type text into input fields (forms, search boxes)\n",
+    "- take_screenshot: Capture a screenshot of the page\n",
+    "- scroll_page: Scroll the page up/down/left/right\n",
+    "- wait_for_element: Wait for an element to appear or reach a state\n",
```
3coins: Isn't this info already added to the prompt by bind_tools?

libs/aws/langchain_aws/tools/browser_tools.py
sundargthb: Good point for future consideration. Playwright does support `clip` for area capture, but that would require width, height parameters. For v1, the boolean covers the two most common cases (viewport vs full page). We can extend to support area capture in a future iteration.

libs/aws/langchain_aws/tools/browser_tools.py
sundargthb: Yes, tool outputs are passed to the LLM as text, so binary screenshot data must be encoded. This is the standard pattern in LangChain for returning images from tools.

libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
sundargthb: yes, will update the docstring to match

libs/aws/langchain_aws/tools/browser_tools.py
sundargthb: Good catch, will add the constraint.

libs/aws/langchain_aws/tools/browser_tools.py
sundargthb: yes, that will be better. Will make the change

libs/aws/langchain_aws/tools/browser_tools.py
sundargthb: will extract to a helper method

libs/aws/langchain_aws/tools/browser_tools.py
sundargthb: Some context helps the LLM understand when to use this vs other tools.. I can reduce it a bit

libs/aws/langchain_aws/tools/browser_tools.py
sundargthb: Will condense similarly.

samples/tools/bedrock_agentcore_browser.ipynb
sundargthb: This markdown cell was intended for readers of the notebook, but happy to remove it to reduce redundancy.

libs/aws/langchain_aws/tools/browser_tools.py
3coins: Would that require changing boolean value to a string, that would be an API change, requiring a major version release. If you think it will change, it's better to make this a string with Literal values.

libs/aws/langchain_aws/tools/browser_tools.py
3coins: I was referring to this line only, which is just append to the message, it seems redundant and we can directly pass the value to the message.
```
 page_type = "full page" if full_page else "viewport"
 ```

libs/aws/langchain_aws/tools/browser_tools.py
sundargthb: Good point - changed to `Literal["viewport", "full_page"]` now to avoid breaking changes later. This also leaves room to add `"area"` in the future.

3coins (approved):
LGTM!

PR #831, opened 2026-01-14
feat(aws): Pass integration_source="langchain" to AgentCore SDK clients

sundargthb:
## Summary

Passes `integration_source="langchain"` to all AgentCore SDK client instantiations, enabling AWS to track customer acquisition from the LangChain integration.

## Motivation

AWS AgentCore team is implementing attribution telemetry to measure which framework integrations drive customer adoption. By passing `integration_source="langchain"`, every API request from langchain-aws users will be tagged, allowing AWS to attribute these customers to the LangChain ecosystem.

This supports the broader goal of demonstrating LangChain's value as an AgentCore distribution channel.

## Changes

| File | Line | Change |
|------|------|--------|
| `libs/aws/langchain_aws/tools/code_interpreter_toolkit.py` | 124 | Add `integration_source="langchain"` |
| `libs/aws/langchain_aws/tools/browser_session_manager.py` | 108 | Add `integration_source="langchain"` (async) |
| `libs/aws/langchain_aws/tools/browser_session_manager.py` | 165 | Add `integration_source="langchain"` (sync) |
| `libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/utils.py` | client creation | Add `integration_source="langchain"` |

## Example Change
```python
# Before
code_interpreter = CodeInterpreter(region=self.region)

# After
code_interpreter = CodeInterpreter(
    region=self.region,
    integration_source="langchain"
)
```

## What This Enables

When a LangChain user makes an AgentCore API call, the HTTP request will include:
```
User-Agent: ... bedrock-agentcore-python/1.0.0 (integration_source=langchain)
```

AWS can then query their logs to answer: "How many customers discovered AgentCore through LangChain?"

## Backward Compatibility

✅ **No breaking changes** to langchain-aws public API. This is an internal implementation detail.

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/utils.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/utils.py
@@ -474,7 +474,7 @@ def create_client_config(config: Config | None = None) -> Config:
     existing_user_agent = getattr(config, "user_agent_extra", "") if config else ""
     new_user_agent = (
         f"{existing_user_agent} x-client-framework:langgraph-checkpoint-aws "
-        f"md/sdk_user_agent/{SDK_USER_AGENT} (integration_source=langchain)".strip()
+        f"md/sdk_user_agent/{SDK_USER_AGENT}".strip()
     )
 
     return Config(user_agent_extra=new_user_agent, **config_kwargs)


libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/utils.py
```diff
@@ -474,7 +474,7 @@ def create_client_config(config: Config | None = None) -> Config:
     existing_user_agent = getattr(config, "user_agent_extra", "") if config else ""
     new_user_agent = (
         f"{existing_user_agent} x-client-framework:langgraph-checkpoint-aws "
-        f"md/sdk_user_agent/{SDK_USER_AGENT}".strip()
+        f"md/sdk_user_agent/{SDK_USER_AGENT} (integration_source=langchain)".strip()
```
3coins: Does this user-agent string have to conform to a defined pattern? If the integration source is already part of the API, is this even useful?

sundargthb: i thought to extend the client config to include it but not required here.. removed this

PR #833, opened 2026-01-14
fix(aws): handling of falsey numeric values for temperature and topP in ChatBedrockConverse

### LINKED ISSUE #832: ChatBedrockConverse.stream drops temperature=0 due to falsey check
### Package

langchain-aws

### Checked other resources

### Example Code

Script used to find the issue:

```python

llm = ChatBedrockConverse(
    model="amazon.nova-lite-v1:0",
    region_name="us-east-1",
)

messages = [("human", "Hello")]

# temperature=0 is a valid value, but it gets dropped from the request payload
for chunk in llm.stream(messages, temperature=0):
    print(chunk.text, end="", flush=True)
```

#### Observed Request payload

When running the script above and inspecting the outgoing HTTP request, the payload contains an empty `inferenceConfig` object (i.e., `temperature` is not included):

```
{
    "messages": [
        {
            "role": "user",
            "content": [
                {
                    "text": "Hello"
                }
            ]
        }
    ],
    "system": [],
    "inferenceConfig": {}
}
```

#### Expected Request Payload

`temperature` should be explicitly included when provided, even if set to `0`:

```
{
  "inferenceConfig": {
    "temperature": 0
  }
}
```

#### Versions

```
langchain==1.2.0
langchain-aws==1.2.0
langchain-core==1.2.5
```

```shell

```

### Description

When using the LangChain AWS `ChatBedrockConverse` client, passing a temperature value of `0` (or `0.0`) to the stream method does not behave as expected.

Although `0` is a valid float value for temperature, it is treated as a falsey value and causes the client to fall back to the internal default (`self.temperature`), which is None. As a result, the `temperature` field is omitted from the HTTP request sent to AWS Bedrock.

This behavior prevents users from explicitly setting `temperature=0`, which is a valid and commonly used configuration.

#### Observed Behavior

- Passing `temperature=0` or `temperature=0.0` results in the parameter being omitted from the request payload.
- Inspecting the outgoing HTTP requests confirms that the `temperature` field is not included.
- The client silently falls back to `self.temperature`, whose default value is `None`.

#### Root Cause

The issue is caused by a boolean check on the `temperature` parameter when building the `inferenceConfig` dictionary:

```py
inferenceConfig = {
    "maxTokens": None
    if should_unset_max_tokens
    else (maxTokens or self.max_tokens),
    "temperature": temperature or self.temperature,
    "topP": self.top_p or topP,
    "stopSequences": stop or stopSequences or self.stop_sequences,
}
```

Since `0` is falsey in Python, the expression:

```py
temperature or self.temperature
```

evaluates to `self.temperature` when `temperature == 0`, instead of preserving the explicitly provided value.

#### Expected Behavior

- Passing `temperature=0` should explicitly set the temperature to `0` in the request payload.
- Only `None` should trigger the ballback to `self.temperature`

Line 1407 is doing a boolean check on a float value of `0`, resulting in a boolean value of `False`. The default value of `None` as defined on line 412 is `None`

#### References
 
- Problematic line (temperature handling) at line 1407: https://github.com/langchain-ai/langchain-aws/blob/542e3e88e4d281058b3d22ffe8e2868067e38574/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L1407
- Default value definition for self.temperature at line 421: https://github.com/langchain-ai/langchain-aws/blob/542e3e88e4d281058b3d22ffe8e2868067e38574/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L421

abriere-coveo:
Fixes #832 

# Summary

This PR fixes an issue in `ChatBedrockConverse` where explicitly passing valid numeric values such as `0` or `0.0` for `temperature` and `topP` was incorrectly treated as falsey and replaced by the instance defaults.

As a result, these parameters were omitted from the `inferenceConfig` sent to AWS Bedrock, preventing users from explicitly setting them to zero.

# Root Cause

The issue was caused by using boolean or checks when building the `inferenceConfig` dictionary:

```py
"temperature": temperature or self.temperature,
"topP": self.top_p or topP,
```

In Python, `0` evaluates to `False`, causing the code to fall back to the default values even when a valid value was explicitly provided.

# Fix

This PR replaces the boolean checks with explicit `None` checks for both parameters:

```py
"temperature": self.temperature if temperature is None else temperature,
"topP": self.top_p if topP is None else topP,
```

This ensures:

- `None` continues to trigger the fallback to defaults
- Valid numeric values like `0` and `0.0` are preserved and correctly sent to AWS

# Impact

- Allows users to explicitly set `temperature=0` and/or `topP=0`
- Aligns behavior with common expectations for numeric model parameters
- Backwards-compatible for all non-zero values

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -1404,8 +1404,8 @@ class ChatBedrockConverse(BaseChatModel):
                 "maxTokens": None
                 if should_unset_max_tokens
                 else (maxTokens or self.max_tokens),
-                "temperature": temperature or self.temperature,
-                "topP": self.top_p or topP,
+                "temperature": self.temperature if temperature is None else temperature,
+                "topP": self.top_p if topP is None else topP,
                 "stopSequences": stop or stopSequences or self.stop_sequences,
             }
         if not toolConfig and tools:


michaelnchin (approved):
Nice catch, thanks @abriere-coveo !

abriere-coveo:
Thanks for taking the time to review and merge @michaelnchin !

Taking some time to think about this bug a bit more, do you have any testing strategies or idea that could be added to this repo in order to validate that such a bug won't happen again?

We can talk about this in the issue directly if you prefer -> https://github.com/langchain-ai/langchain-aws/issues/832

Issue #793, opened 2025-12-09
system prompt caching not working (ChatBedrock using anthropic model)


JHSeo-git:
### Package

langchain-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code

```python
from langchain_aws import ChatBedrock

model = ChatBedrock(...)

system_prompt = SystemMessage([
  { "type": "text", "text": "..." }, # over 1,024 tokens
  { "cache_control": {"type": "ephemeral"}
])
human_prompt = HumanMessage("Hello!")

messages = [system_prompt, human_prompt]

model.stream(messages, ...)

# "input_token_details": {
#   "cache_creation": 0,
#   "cache_read": 0
# },
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

`cache_control` is preserved in `_format_anthropic_messages` but dropped in `_stream` for system messages

[_format_anthropic_messages](https://github.com/langchain-ai/langchain-aws/blob/07a5f461f243adbea193a7a354daeca028f702fe/libs/aws/langchain_aws/chat_models/bedrock.py#L476) correctly preserves `cache_control` for system messages:

```python
if item.get("cache_control"):
    content_item["cache_control"] = {"type": "ephemeral"}
system_blocks.append(content_item)
```

But in the [_stream](https://github.com/langchain-ai/langchain-aws/blob/07a5f461f243adbea193a7a354daeca028f702fe/libs/aws/langchain_aws/chat_models/bedrock.py#L930) method, system messages are converted to plain text:

```python
elif isinstance(system_raw, list):
    # Convert list of dicts to string representation
    system_str = "\n".join(
        item.get("text", "") if isinstance(item, dict) else str(item)
        for item in system_raw
    )
```

It seems that all `cache_control` metadata is stripped and only the text content is extracted. I suspect this may be the reason why the functionality is not working as expected.

Thank you for taking the time to look into this.

JHSeo-git:
I was able to resolve the issue by switching from `ChatBedrock` to `ChatBedrockConverse`, which correctly preserves `cachePoint` metadata during streaming. The cache behavior now works as expected, and the `cache_creation` / `cache_read` counters are reflected properly.

Closing the issue.

Haknt:
Submitted a fix for this in #838

PR #838, opened 2026-01-18
fix(aws): preserve cache_control in system prompts for ChatBedrock

Haknt:
When using list-format system prompts with `cache_control` for prompt caching, `_generate()` and `_stream()` were converting the list to a string, dropping `cache_control`.

```python
system = SystemMessage([{
    "type": "text",
    "text": "...",
    "cache_control": {"type": "ephemeral"}
}])
# cache_control was lost after conversion
```

Now passes through as-is. When `system_prompt_with_tools` is set with a list system prompt, prepends tools as a content block instead of string concatenation.

### `cache_control` kwarg support

You can now pass `cache_control` directly to invoke, generate, and stream:

```python
# invoke
model.invoke(messages, cache_control={"type": "ephemeral", "ttl": "5m"})

# stream
for chunk in model.stream(messages, cache_control={"type": "ephemeral", "ttl": "5m"}):
    print(chunk.content)
```

This applies `cache_control` to the last message at API call time, useful for middleware integration (#839). Works with all content block types including `tool_result`, matching `ChatAnthropic` behavior.

### Changes

- **System prompt preservation**: List-format system prompts with `cache_control` are passed through without conversion to string
- **Tools prepend**: When `system_prompt_with_tools` is set with a list system prompt, tools are prepended as a content block to preserve `cache_control`
- **`cache_control` kwarg**: Supported in `invoke()`, `generate()`, `stream()`, and `astream()` — applies to the last message's last content block
- **Type safety**: Added `assert isinstance(result, tuple)` for proper mypy narrowing on `format_messages` return type

### Tests

- System prompt `cache_control` preserved (invoke, stream, astream)
- List/string system prompt formats
- Tools + list system prompt prepend
- `cache_control` kwarg applied to string and list content
- `cache_control` kwarg not leaking as top-level API body parameter
- `cache_control` kwarg in stream mode

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -787,6 +787,32 @@ _message_type_lookups = {
 }
 
 
+def _apply_cache_control_to_messages(
+    cache_control: Optional[Dict[str, Any]],
+    formatted_messages: Optional[List[Dict[str, Any]]],
+) -> None:
+    """Apply cache_control to the last content block of the last message."""
+    if not cache_control or not formatted_messages:
+        return
+    for fmt_msg in reversed(formatted_messages):
+        content = fmt_msg.get("content")
+        if isinstance(content, list) and content:
+            for block in reversed(content):
+                if isinstance(block, dict):
+                    block["cache_control"] = cache_control
+                    break
+            break
+        elif isinstance(content, str):
+            fmt_msg["content"] = [
+                {
+                    "type": "text",
+                    "text": content,
+                    "cache_control": cache_control,
+                }
+            ]
+            break
+
+
 class ChatBedrock(BaseChatModel, BedrockBase):
     """A chat model that uses the Bedrock API."""
 
@@ -955,25 +981,9 @@ class ChatBedrock(BaseChatModel, BedrockBase):
                     ] + list(system)
 
             # Apply cache_control to last message if provided via kwargs
-            cache_control = kwargs.pop("cache_control", None)
-            if cache_control and formatted_messages:
-                for fmt_msg in reversed(formatted_messages):
-                    content = fmt_msg.get("content")
-                    if isinstance(content, list) and content:
-                        for block in reversed(content):
-                            if isinstance(block, dict):
-                                block["cache_control"] = cache_control
-                                break
-                        break
-                    elif isinstance(content, str):
-                        fmt_msg["content"] = [
-                            {
-                                "type": "text",
-                                "text": content,
-                                "cache_control": cache_control,
-                            }
-                        ]
-                        break
+            _apply_cache_control_to_messages(
+                kwargs.pop("cache_control", None), formatted_messages
+            )
         elif provider in ("openai", "qwen"):
             formatted_messages = cast(
                 List[Dict[str, Any]],
@@ -1116,25 +1126,9 @@ class ChatBedrock(BaseChatModel, BedrockBase):
                 citations_enabled = _citations_enabled(formatted_messages)
 
                 # Apply cache_control to last message if provided via kwargs
-                cache_control = params.pop("cache_control", None)
-                if cache_control and formatted_messages:
-                    for fmt_msg in reversed(formatted_messages):
-                        content = fmt_msg.get("content")
-                        if isinstance(content, list) and content:
-                            for block in reversed(content):
-                                if isinstance(block, dict):
-                                    block["cache_control"] = cache_control
-                                    break
-                            break
-                        elif isinstance(content, str):
-                            fmt_msg["content"] = [
-                                {
-                                    "type": "text",
-                                    "text": content,
-                                    "cache_control": cache_control,
-                                }
-                            ]
-                            break
+                _apply_cache_control_to_messages(
+                    params.pop("cache_control", None), formatted_messages
+                )
 
             elif provider in ("openai", "qwen"):
                 formatted_messages = cast(
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
@@ -874,7 +874,7 @@ def test_beta_use_converse_api_with_inference_profile_as_nova_model(
             "us-west-2",
         ),
         (
-            "eu.anthropic.claude-3-haiku-20240307-v1:0",
+            "eu.anthropic.claude-haiku-4-5-20251001-v1:0",
             None,
             "anthropic",
             nullcontext(),
@@ -1860,7 +1860,7 @@ def test_system_prompt_cache_control_preserved() -> None:
 
     llm = ChatBedrock(
         client=mock_client,
-        model_id="anthropic.claude-3-haiku-20240307-v1:0",
+        model_id="anthropic.claude-haiku-4-5-20251001-v1:0",
         region_name="us-west-2",
     )
     system = SystemMessage(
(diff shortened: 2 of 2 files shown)


Haknt:
Hey @ccurme @michaelnchin — friendly ping. This PR has been open since January and CI is green. It fixes a real issue we hit in production: `cache_control` gets silently stripped from system prompts and non-text content blocks in ChatBedrock, breaking Bedrock prompt caching. Would appreciate a review when you get a chance. Thanks!

rowan-walshe:
Feel free to ignore me, as I'm likely wrong, but isn't `cache_control` an Anthropic API specific parameter? For models served by AWS Bedrock, including for Anthropic models, don't you need to use `cachePoint` instead ([docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-get-started))?

Haknt:
> Feel free to ignore me, as I'm likely wrong, but isn't cache_control an Anthropic API specific parameter? For models served by AWS Bedrock, including for Anthropic models, don't you need to use cachePoint instead?

Great point! So yes, Bedrock Converse API uses `cachePoint` — but `langchain-aws` handles that conversion internally. Users mark content with `cache_control` (LangChain's standard way), and `_messages_to_bedrock()` converts it to `cachePoint` before it hits the API.

The problem is: for system prompts and non-text content blocks, `cache_control` gets dropped *before* that conversion ever runs. So `cachePoint` is never generated for those blocks, and caching silently doesn't work.

This PR just makes sure `cache_control` survives long enough to be converted.

michaelnchin:
Hi @Haknt, thank you for your work on this! We appreciate your patience - will take a look at this soon given that this is a production blocker for you. 

For some context, updates to ChatBedrock/BedrockLLM are generally lower on our priority list, as we're actively pushing users to migrate to ChatBedrockConverse, given that Converse API is much more feature-complete and maintainable for newer models than the legacy InvokeModel API. 

For this specific PR, the same issue also does not affect ChatBedrockConverse [per the linked issue](https://github.com/langchain-ai/langchain-aws/issues/793#issuecomment-3635953692). If you are able to make the switch, we would definitely encourage you to do so. OTOH, if you do still have a specific need for ChatBedrock / InvokeModel API, we would love to know more about your use case.

Haknt:
> If you are able to make the switch, we would definitely encourage you to do so. OTOH, if you do still have a specific need for ChatBedrock / InvokeModel API, we would love to know more about your use case.

Thanks for the context! We're on `ChatBedrock` (InvokeModel) for most of our pipeline — our caching middleware and message formatting were built around it. `ChatBedrockConverse` is only used for document parsing.

We actually have a follow-up PR (#839) that adds a native prompt caching middleware for Bedrock — also depends on this one. Right now we maintain a local `ChatBedrock` subclass and a custom middleware as workarounds, so merging these two would let us drop ~600 lines of patch code on our side.

If `ChatBedrock` is being wound down though, that's good to know — we can plan the migration. But it would still be great to get these in for the transition period.

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -929,34 +929,51 @@ def _stream(
             return
         provider = self._get_provider()
         prompt: Optional[str] = None
-        system: Optional[str] = None
+        system: Optional[Union[str, List[Dict[str, Any]]]] = None
         formatted_messages: Optional[List[Dict[str, Any]]] = None
 
         if provider == "anthropic":
             result = ChatPromptAdapter.format_messages(provider, messages)
+            assert isinstance(result, tuple)
```
michaelnchin, line 963: This assert is redundant - for Anthropic, `format_messages` calls `_format_anthropic_messages` which always returns tuple: https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock.py#L434

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -1071,38 +1088,54 @@ def _generate(
...
         else:
             prompt: Optional[str] = None
-            system: Optional[str] = None
+            system: Optional[Union[str, List[Dict[str, Any]]]] = None
             formatted_messages: Optional[List[Dict[str, Any]]] = None
             params: Dict[str, Any] = {**kwargs}
 
             if provider == "anthropic":
                 result = ChatPromptAdapter.format_messages(provider, messages)
+                assert isinstance(result, tuple)
```
michaelnchin, line 1107: Same as in `_stream`, remove this

libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -1071,38 +1088,54 @@ def _generate(
...
+                            break
+                        elif isinstance(content, str):
+                            fmt_msg["content"] = [
+                                {
+                                    "type": "text",
+                                    "text": content,
+                                    "cache_control": cache_control,
+                                }
+                            ]
+                            break
```
michaelnchin: This block duplicates the same logic in `_stream` - let's move this out into a helper function (`_apply_cache_control_to_messages` ?)

michaelnchin (approved):
Thank you for the contribution @Haknt ! Changes look great overall and work well per manual testing. Added some non-blocking comments.

michaelnchin (commented):
Existing integ tests all pass on branch: https://github.com/langchain-ai/langchain-aws/actions/runs/22652362716.

libs/aws/tests/unit_tests/chat_models/test_bedrock.py
```diff
@@ -1839,3 +1839,428 @@ def test_service_tier_passed_to_as_converse() -> None:
...
+    ).encode()
+    mock_response.get.side_effect = lambda key, default=None: {
+        "body": mock_response.get.return_value,
+        "ResponseMetadata": {"HTTPHeaders": {}},
+    }.get(key, default)
+    mock_client.invoke_model.return_value = mock_response
+
+    llm = ChatBedrock(
+        client=mock_client,
+        model_id="anthropic.claude-3-haiku-20240307-v1:0",
```
michaelnchin: Nit: Haiku 3 is being retired soon, update this and any other references to `anthropic.claude-haiku-4-5-20251001-v1:0`

libs/aws/langchain_aws/chat_models/bedrock.py
Haknt: Actually, this isn't redundant — it's needed for **mypy type narrowing**. `format_messages()` returns `Union[Tuple[...], List[Dict]]`, so without the assert, mypy can't narrow `result[0]` to the tuple branch. Removing it gives two `[assignment]` errors:

```
Incompatible types in assignment (expression has type
"str | list[dict[str, Any]] | dict[str, Any]",
variable has type "str | list[dict[str, Any]] | None")
```

Verified locally with `mypy --ignore-missing-imports --disallow-untyped-defs` (matching the repo's pyproject.toml config). Happy to switch to an `if not isinstance(...): raise TypeError(...)` if you'd prefer an explicit error over AssertionError, but removing it entirely would break the type checker.

libs/aws/langchain_aws/chat_models/bedrock.py
Haknt: Same reasoning as above — needed for mypy type narrowing. See [my reply](https://github.com/langchain-ai/langchain-aws/pull/838#discussion_r2882637071).

libs/aws/langchain_aws/chat_models/bedrock.py
Haknt: Done — extracted into `_apply_cache_control_to_messages()` at module level, consistent with other helpers like `_format_anthropic_messages`. ef020c8

libs/aws/tests/unit_tests/chat_models/test_bedrock.py
Haknt: Updated all references (including the existing one at line 877). ef020c8

PR #839, opened 2026-01-18
feat(aws): add prompt caching middleware for Bedrock

### LINKED ISSUE #723: Feature Request: Add Prompt Caching Middleware for ChatBedrockConverse (similar to Anthropic)
### Current Situation
- LangChain 1.0 introduces the middleware pattern for agent customization ([blog post](https://blog.langchain.com/agent-middleware/))
- **Anthropic integration already includes a prompt caching middleware** that uses `modify_model_request` to automatically add `cache_control` tags to messages
- AWS Bedrock supports prompt caching via `cachePoint` objects in the Converse API ([AWS docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html))
- Currently, users must **manually add cachePoint objects** to message content, which is error-prone and verbose

### Problem
Without a dedicated middleware:
- Users need to manually construct messages with `cachePoint` objects:
```python
# Current manual approach (cumbersome)
system_message = SystemMessage(
    content=[
        {"type": "text", "text": router_prompt},
        {"cachePoint": {"type": "default"}},  # Manual cache point
    ]
)
```
- No automatic cache point placement based on best practices
- Inconsistent caching patterns across different use cases
- Harder to maintain and update caching logic

### Requested Feature
Add a **prompt caching middleware for ChatBedrockConverse** that:
- Automatically adds `cachePoint` objects to appropriate message positions
- Follows AWS Bedrock best practices for cache placement
- Provides configurable caching strategies (e.g., cache system prompts, cache after N messages)
- Works seamlessly with LangChain 1.0 middleware pattern

### Expected API (Following Anthropic Pattern)

**Reference: Anthropic Implementation (Already Exists)**
```python
from langchain_anthropic import ChatAnthropic
from langchain_anthropic.middleware import AnthropicPromptCachingMiddleware
from langchain.agents import create_agent

agent = create_agent(
    model=ChatAnthropic(model="claude-sonnet-4-latest"),
    system_prompt=LONG_PROMPT,
    middleware=[AnthropicPromptCachingMiddleware(ttl="5m")],  # ✅ Already available
)
```

**Requested: AWS Bedrock Implementation (Same Pattern)**
```python
from langchain_aws.middleware import BedrockPromptCachingMiddleware  # ❌ Doesn't exist yet
from langchain.agents import create_agent

agent = create_agent(
    model=ChatBedrockConverse(
        model="anthropic.claude-sonnet-4-20250514-v1:0",
        region_name="us-west-2"
    ),
    system_prompt=LONG_PROMPT,
    middleware=[BedrockPromptCachingMiddleware(ttl="5m")],  # Requested feature
)
```

Haknt:
Closes #723

> **Note:** Depends on #838 for `cache_control` kwarg support in ChatBedrock.

Adds prompt caching middleware for Anthropic models on AWS Bedrock, following the same pattern as `langchain-anthropic`'s middleware.

### Changes

Two middleware classes:
- `BedrockPromptCachingMiddleware` for `ChatBedrock` (InvokeModel API) — passes `cache_control` via `model_settings`
- `BedrockConversePromptCachingMiddleware` for `ChatBedrockConverse` (Converse API) — adds `cachePoint` to system prompt via immutable `request.override()`

Both avoid direct message mutation to prevent accumulation issues with LangGraph checkpointing.

#### Immutable pattern (Converse middleware)

`_apply_cache_point` uses `request.override(system_message=SystemMessage(...))` instead of mutating `request.system_prompt` directly (which is a read-only property on `ModelRequest`). The `cachePoint` block passes through `_lc_content_to_bedrock()` as-is since it has no `"type"` key.

### Usage

```python
from langchain_aws.middleware import BedrockConversePromptCachingMiddleware

model = ChatBedrockConverse(model="anthropic.claude-3-haiku-20240307-v1:0")
middleware = BedrockConversePromptCachingMiddleware()

agent = create_react_agent(model, tools, middleware=[middleware])
```

For `ChatBedrock`:
```python
from langchain_aws.middleware import BedrockPromptCachingMiddleware

model = ChatBedrock(model_id="anthropic.claude-3-haiku-20240307-v1:0")
middleware = BedrockPromptCachingMiddleware(ttl="5m")
```

### Notes
- No message mutation — prevents checkpoint accumulation across turns
- Skips non-Anthropic models gracefully (configurable: ignore/warn/raise)
- Configurable `min_messages_to_cache` threshold
- 11 unit tests covering both middleware classes

### DIFF
--- a/libs/aws/langchain_aws/middleware/prompt_caching.py
+++ b/libs/aws/langchain_aws/middleware/prompt_caching.py
@@ -2,17 +2,14 @@
 
 Requires:
     - langchain: For agent middleware framework
-    - langchain-aws: For ChatBedrock/ChatBedrockConverse models (already a dependency)
+    - langchain-aws: For ChatBedrock model (already a dependency)
 """
 
 from collections.abc import Awaitable, Callable
-from typing import Any, Literal, Union
+from typing import Literal
 from warnings import warn
 
-from langchain_core.messages import SystemMessage
-
 from langchain_aws.chat_models.bedrock import ChatBedrock
-from langchain_aws.chat_models.bedrock_converse import ChatBedrockConverse
 
 try:
     from langchain.agents.middleware.types import (
@@ -30,7 +27,7 @@ except ImportError as e:
     raise ImportError(msg) from e
 
 
-def _is_anthropic_model(model: Union[ChatBedrock, ChatBedrockConverse]) -> bool:
+def _is_anthropic_model(model: ChatBedrock) -> bool:
     """Check if the model is an Anthropic model on Bedrock."""
     model_id = getattr(model, "model_id", "") or getattr(model, "model", "")
     return "anthropic" in model_id.lower()
@@ -82,7 +79,7 @@ class BedrockPromptCachingMiddleware(AgentMiddleware):
             msg = (
                 f"BedrockPromptCachingMiddleware only supports ChatBedrock, "
                 f"not {type(request.model).__name__}. "
-                f"For ChatBedrockConverse, use BedrockConversePromptCachingMiddleware."
+                f"Converse API prompt caching support is planned for a future release."
             )
             if self.unsupported_model_behavior == "raise":
                 raise ValueError(msg)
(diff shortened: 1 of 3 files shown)


libs/aws/langchain_aws/middleware/prompt_caching.py
```diff
@@ -0,0 +1,250 @@
...
+    raise ImportError(msg) from e
+
+
+def _is_anthropic_model(model: Union[ChatBedrock, ChatBedrockConverse]) -> bool:
+    """Check if the model is an Anthropic model on Bedrock."""
+    model_id = getattr(model, "model_id", "") or getattr(model, "model", "")
+    return "anthropic" in model_id.lower()
+
+
+class BedrockPromptCachingMiddleware(AgentMiddleware):
```
michaelnchin, line 36: Looks good! I see that this closely follows the reference [Anthropic prompt caching middleware](https://github.com/langchain-ai/langchain/blob/e50625e7c3cac04be68ae095bee030e0c321ef34/libs/partners/anthropic/langchain_anthropic/middleware/prompt_caching.py#L30), as we generally try to do for classes in this package.

One side note - while ChatBedrock middleware is currently limited to Anthropic models, it should also eventually allow Nova models since they also support prompt caching (see [AWS docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html)). But, this isn’t possible to add as-is since ChatBedrock always [delegates Nova requests to ChatBedrockConverse](https://github.com/langchain-ai/langchain-aws/blob/464b1b64631647a7e30baaa0d25d23ca707cf032/libs/aws/langchain_aws/chat_models/bedrock.py#L837-L838), which  doesn’t handle raw `cache_control`yet. Can look to add this in a separate PR with the Converse changes mentioned below.

libs/aws/langchain_aws/middleware/prompt_caching.py
```diff
@@ -0,0 +1,250 @@
...
+            # ChatBedrock will apply this at API call time.
+            new_model_settings = {
+                **request.model_settings,
+                "cache_control": self._get_cache_control_settings(),
+            }
+            return await handler(request.override(model_settings=new_model_settings))
+        return await handler(request)
+
+
+class BedrockConversePromptCachingMiddleware(AgentMiddleware):
```
michaelnchin: I see that the Converse middleware diverges significantly from the Anthropic implementation, as a byproduct of the limitation that ChatBedrockConverse doesn’t have its own `model_settings` -> `cachePoint` injection mechanism like in ChatBedrock. So the middleware has to manually inject the `cachePoint` blocks into the messages instead. 

I understand the approach in the context of this constraint, but there are a few gaps directly resulting from the `model_settings` bypass:
- We’re only updating the `request.system_prompt` and not the other messages, so there’s no real benefit for multi-turn conversations beyond the system prompt. I’m assuming this is related to the extra complexity of iterating->mutating all of request.messages to place `cachePoint`. The InvokeModel middleware is able to avoid this by delegating to ChatBedrock where the marker just gets applied to the last message at runtime to cache everything.
- In `_apply_cache_point`, there’s a condition to ostensibly avoid inserting duplicate `cachePoint` blocks for list type system prompts, but this will never actually get executed because the base `ModelRequest.system_prompt` type [can only return string](https://reference.langchain.com/python/langchain/agents/middleware/types/ModelRequest#member-system_prompt-9). So, list-format system prompts will always end up with duplicate `cachePoint`s. In the ChatBedrock implementation, we defer to the chat model to build the system and other messages and append `cachePoint` at call time, instead of having to update `ModelRequest` directly, so duplicates aren’t ever an issue.

Couple other notes:
- The `cachePoint` dict is hardcoded to only pass `type`, but Converse does also allow for `"ttl" : "5m | 1h” (see [docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html)), so this should be supported.
- Like the ChatBedrock middleware, we currently only allow Anthropic models, but should also eventually add support for prompt caching with Nova.

Given the above, we should probably look to update ChatBedrockConverse directly with `cache_control` model setting support and use this to simplify the Converse middleware. But I can cover that in a separate PR, it’s ok if this just covers InvokeModel.

Haknt: Good points. Removed the Converse middleware in b949206 — this PR now only covers InvokeModel. Let me know if I can help with the Converse `cache_control` support.

Haknt:
> Would you prefer to remove the `BedrockConversePromptCachingMiddleware` class in favor of merging everything else immediately?

Yes, removed in b949206. If you'd like, I can take a first pass at the Converse `cache_control` model_settings support in a separate PR. Any pointers on how you'd approach it?

PR #840, opened 2026-01-19
feat(aws): Be aware of "au" region prefix in Bedrock model_id

tgargiani:
Using a Bedrock model_id with the "au" prefix (e.g. `au.anthropic.claude-sonnet-4-5-20250929-v1:0`) fails on `NotImplementedError: Provider au model does not support chat.` in `chat_models/bedrock.py`.
Fixed the `_get_provider` function to be aware of this prefix and added a test parameter.

### DIFF
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -1678,7 +1678,10 @@ def test__get_base_model() -> None:
         model="us.anthropic.claude-3-5-haiku-20241022-v1:0", region_name="us-west-2"
     )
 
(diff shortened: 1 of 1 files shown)


michaelnchin (commented):
@tgargiani thank you for the submission!

The `BedrockLLM` change looks good. Could you also update the equivalent allowlists for [ChatBedrockConverse](https://github.com/langchain-ai/langchain-aws/blob/2971616124582ee5abbc62332f74e0502fc7036d/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L919) and [BedrockEmbeddings](https://github.com/langchain-ai/langchain-aws/blob/2971616124582ee5abbc62332f74e0502fc7036d/libs/aws/langchain_aws/embeddings/bedrock.py#L156)?

tgargiani:
@michaelnchin Sure thing, thanks for your guidance! I also added a test case to `test_bedrock_converse.py`.

PR #725, opened 2025-10-24
fix(langgraph-checkpoint-aws): clean orphan tool_calls in AgentCoreMemorySaver to prevent Bedrock ValidationException

murilosimao:
Fixed [#726](https://github.com/langchain-ai/langchain-aws/issues/726)

PR #842, opened 2026-01-20
fix(langgraph-checkpoint-aws): Update `AgentCoreMemorySaver` to patch instead of remove dangling tool_calls

### LINKED ISSUE #837: When using AgentCoreMemorySaver with Interruption I cannot resume the execution because when loading state the checkpointer executes: Removed 1 orphaned tool_call(s) from AIMessage during checkpoint load
### Package

langchain-aws

### Checked other resources

### Example Code

```python
import asyncio
import os

from dotenv import load_dotenv
from langgraph.pregel.debug import RunnableConfig
from langgraph.types import Command, Interrupt

load_dotenv()

CHECKPOINTER_MEMORY_ID = os.environ.get("CHECKPOINTER_MEMORY_ID")

if not CHECKPOINTER_MEMORY_ID:
    raise ValueError("CHECKPOINTER_MEMORY_ID environment variable is not set")

# from langgraph.checkpoint.memory import InMemorySaver

from src.agent import TicketManagerAgent
from langgraph_checkpoint_aws import AgentCoreMemorySaver
from src.model.mcp_endpoint_config import MCPEndpointConfig

ticket_manager_mcp_config = MCPEndpointConfig(
    name="ticket_manager",
    url="mi_api",
    transport="http",
)


agent_checkpointer = AgentCoreMemorySaver(
    CHECKPOINTER_MEMORY_ID, region_name='us-east-1'
)
# agent_checkpointer = InMemorySaver()

ticket_manager_agent = TicketManagerAgent(
    agent_checkpointer, [ticket_manager_mcp_config]
)

if __name__ == "__main__":

    async def main():
        agent = await ticket_manager_agent.create_agent()
        config = RunnableConfig(
            configurable={"thread_id": "11301010214", "actor_id": "actor_123"}
        )
        if agent:
            reply = await agent.ainvoke(
                {
                    "messages": [
                        {
                            "role": "user",
                            "content": "Hola, dame detalles del ticket xyz?",
                        }
                    ]
                },
                config,
            )
            print(reply)
            current_agent_state = agent.get_state(config)
            if current_agent_state.interrupts:
                approvals: list[str] = []
                if current_agent_state.interrupts:
                    print("Interrupted!")
                    for index, interrupts in enumerate(current_agent_state.interrupts):
                        approvals.append("approve")
                        interrupts = interrupts.value
                        print(f"Interrupt {index + 1}:")
                        action_requests = interrupts["action_requests"]
                        print(
                            f"Action Requests: {action_requests} for interrupt: {index + 1}"
                        )
                        review_configs = interrupts["review_configs"]
                        config_map = {cfg["action_name"]: cfg for cfg in review_configs}
                        for action in action_requests:
                            review_config = config_map[action["name"]]
                            print(f"Tool: {action['name']}")
                            print(f"Arguments: {action['args']}")
                            print(
                                f"Allowed decisions: {review_config['allowed_decisions']}"
                            )
                breakpoint()
                for approval in approvals:
                    reply = await agent.ainvoke(
                        Command(resume={"decisions": [{"type": approval}]}),
                        config=config,
                    )
            else:
                print("No interrupts detected")

    asyncio.run(main())
```

```shell
Removed 1 orphaned tool_call(s) from AIMessage during checkpoint load
```

### Description

Seem to be there is a problem when loading interrupts from agent core memory.  For the next feature I want to expose this agent as an API, this is the reason I am loading the state             
current_agent_state = agent.get_state(config). 
The client invokes the api with the actor_id and thread_id provided. 

When I use the InMemorySaver() to run this thing locally it works perfectly because it restores the interruption.

michaelnchin:
Fixes #837

### Context:

Addressing the follow up item listed in [this comment](https://github.com/langchain-ai/langchain-aws/pull/725#discussion_r2465582743) from PR #725, the context for which can be found [here](https://github.com/langchain-ai/langchain-aws/issues/837#issuecomment-3758107353).

When a checkpoint is saved during tool execution, this may result in an `AIMessage` with `tool_calls` and no corresponding `ToolMessage`, which would normally cause Bedrock to throw a `ValidationException` upon resumption from the checkpoint. 

Currently, `AgentCoreMemorySaver` safeguards against these Bedrock `ValidationException`s related to dangling `tool_calls` by removing them from the `AIMessage`. However, this is not ideal, given that we lose awareness of the tool call attempt from the previous conversation state.

### Solution:

This PR updates the default behavior of `AgentCoreMemorySaver` to instead append a dummy `ToolMessage` for each dangling `tool_calls`, and preserve the original tool calls. 

For example, with the following dangling tool calls:
```python
messages = [
    HumanMessage(content="What's the weather in SF?"),
    AIMessage(
        content="",
        tool_calls=[
            {"id": "call_abc123", "name": "get_weather", "args": {"city": "SF"}}
        ]
    ),
]
```

The old handler (removal) would produce the following output:
```python
result = [
    HumanMessage(content="What's the weather in SF?"),
    AIMessage(
        content="",
        tool_calls=[]  # tool_call removed!
    ),
]
```

While the new handler (patching) will instead generate:
```python
result = [
    HumanMessage(content="What's the weather in SF?"),
    AIMessage(
        content="",
        tool_calls=[
            {"id": "call_abc123", "name": "get_weather", "args": {"city": "SF"}}
        ]
    ),
    ToolMessage(
        content="Tool call 'get_weather' with id 'call_abc123' was interrupted before completion.",
        name="get_weather",
        tool_call_id="call_abc123",
        status="error"
    ),
]
```

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
@@ -429,8 +429,8 @@ class EventProcessor:
 
         return checkpoints, writes_by_checkpoint, channel_data_by_version
 
+    @staticmethod
     def build_checkpoint_tuple(
-        self,
         checkpoint_event: CheckpointEvent,
         writes: list[WriteItem],
         channel_data: dict[tuple[str, str], Any],


libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
```diff
@@ -489,27 +493,88 @@ def build_checkpoint_tuple(
...
+                    )
+                    patched_messages.append(
+                        ToolMessage(
+                            content=(
+                                f"Tool call '{tc_name}' with id '{tc_id}' was "
+                                f"interrupted before completion."
+                            ),
+                            name=tc_name,
+                            tool_call_id=tc_id,
+                        )
```
3coins, line 542: What's the behavior for dangling tool calls, when the session is resumed? Is the LangChain application able to recognize the interrupted tool call (status only has success/error), and resumes the tool call or does the model take over and retries the tool call?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
```diff
@@ -70,13 +73,15 @@ def __init__(
         max_retries: int = DEFAULT_MAX_RETRIES,
         initial_backoff: float = DEFAULT_INITIAL_BACKOFF,
         max_backoff: float = DEFAULT_MAX_BACKOFF,
+        remove_dangling_tool_calls: bool = False,
```
3coins: We are introducing a new option with `False` as default, this seems to me as a breaking change for existing implementations that might behave a certain way with an expectation that an interrupted tool call will not exist when session is resumed. If we are correcting a supposedly buggy behavior, why even have this option at all?

We should call this out explicitly in the release notes, so users can handle this appropriately.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/helpers.py
michaelnchin: Great question. Currently, AgentCoreMemorySaver only recognizes tool call interrupts for the purpose of patching the dangling tool call. The new patching behavior adding `ToolMessage` now makes the tool call functionally "complete" on checkpoint restore, but LangChain doesn't have a mechanism to retry/resume or do anything else differently based on the specific `ToolMessage` status. So, it still fully defers the retry decision to the model.

I'm glad that you brought up the `ToolMessage` status attribute though, because I now see that it [defaults to `success`](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/messages/tool.py#L82). This could potentially inhibit the model itself from retrying the tool call, depending on how it interprets our patch content message, so I'll update the snippet above to explicitly set `status="error"`.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
michaelnchin: > If we are correcting a supposedly buggy behavior, why even have this option at all?

Yeah, I was mulling this over as well. It's a good point that the new default already results in a breaking change, so we might as well fully commit to the new solution, which is also consistent with DeepAgents' handling of dangling tool_calls.

I'll make a note regarding the callout in the next release. Thanks!

Issue #827, opened 2026-01-02
Streaming returns stringified `tool_use.input` instead of dict


austinmw:
### Package

langchain-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code

```python
# langchain-aws==1.0.0
# langchain-core==1.2.5
# langchain==1.2.0
# boto3>=1.35.0

import asyncio
from langchain_core.tools import tool
from langchain.chat_models import init_chat_model

@tool
def get_weather(city: str) -> str:
    """Get weather for a city."""
    return f"Sunny in {city}"

async def main():
    model = init_chat_model(
        "us.anthropic.claude-3-5-haiku-20241022-v1:0",
        model_provider="bedrock_converse",
        region_name="us-east-1",
    ).bind_tools([get_weather])

    # STREAMING - Bug: input is STRING
    chunks = [chunk async for chunk in model.astream("What's the weather in Paris?")]
    streamed = chunks[0]
    for chunk in chunks[1:]:
        streamed = streamed + chunk

    # NON-STREAMING - Correct: input is DICT
    invoked = await model.ainvoke("What's the weather in Paris?")

    def get_input_type(msg):
        for block in msg.content:
            if isinstance(block, dict) and block.get("type") == "tool_use":
                return type(block.get("input")).__name__
        return None

    print(f"Streaming tool_use.input type:     {get_input_type(streamed)}")  # str (BUG)
    print(f"Non-streaming tool_use.input type: {get_input_type(invoked)}")   # dict (correct)

asyncio.run(main())
```

### Error Message and Stack Trace (if applicable)

```shell
When the streamed message is later sent back to Bedrock (e.g., in a multi-turn conversation), it causes:


botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the ConverseStream operation: The format of the value at messages.1.content.2.toolUse.input is invalid. Provide a json object for the field and try again.
```

### Description

When using `astream()` with Bedrock Converse, the `content[].tool_use.input` field in the returned `AIMessageChunk` is a JSON **string** (e.g., `'{"city": "Paris"}'`) instead of a **dict** (e.g., `{"city": "Paris"}`).

This happens because:
1. Bedrock's streaming API returns tool inputs as string chunks that get concatenated
[...]

PR #843, opened 2026-01-22
fix(aws): stringified tool_use.input to dict

RohanDisa:
In streaming mode, Bedrock Converse API returns tool_use.input as a JSON string in delta events, while non-streaming mode returns it as a dict. This fix parses the stringified input back to a dict in _bedrock_to_lc function to ensure consistent behavior between streaming and non-streaming modes.

Fixes: #827

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -1728,11 +1728,12 @@ def _parse_stream_event(event: Dict[str, Any]) -> Optional[BaseMessageChunk]:
         }
         tool_call_chunks = []
         if block["type"] == "tool_use":
+            raw_input = block.get("input")
             tool_call_chunks.append(
                 tool_call_chunk(
                     name=block.get("name"),
                     id=block.get("id"),
-                    args=block.get("input"),
+                    args=json.dumps(raw_input) if raw_input is not None else None,
                     index=event["contentBlockStart"]["contentBlockIndex"],
                 )
             )
@@ -1750,11 +1751,12 @@ def _parse_stream_event(event: Dict[str, Any]) -> Optional[BaseMessageChunk]:
         }
         tool_call_chunks = []
         if block["type"] == "tool_use":
+            raw_input = block.get("input")
             tool_call_chunks.append(
                 tool_call_chunk(
                     name=block.get("name"),
                     id=block.get("id"),
-                    args=block.get("input"),
+                    args=json.dumps(raw_input) if raw_input is not None else None,
                     index=event["contentBlockDelta"]["contentBlockIndex"],
                 )
             )


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -2061,12 +2061,32 @@ def _bedrock_to_lc(content: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
             lc_content.append({"type": "text", "text": block["text"]})
         elif "tool_use" in block:
             block["tool_use"]["id"] = block["tool_use"].pop("tool_use_id", None)
+            # Fix: Parse stringified input from streaming mode to dict
+            tool_use_input = block["tool_use"].get("input")
+            if tool_use_input is not None and isinstance(tool_use_input, str):
+                try:
+                    block["tool_use"]["input"] = json.loads(tool_use_input)
```
3coins: Is the input guaranteed to be a full json string, or there could be instances where partial json could be returned? If that's the case, then [parse_partial_json](https://github.com/langchain-ai/langchain/blob/91bb474dbc7427728bc2296881c9f03d5bf029b5/libs/core/langchain_core/utils/json.py#L58) would be better.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -2061,12 +2061,32 @@ def _bedrock_to_lc(content: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
...
+            if server_tool_use_input is not None and isinstance(
+                server_tool_use_input, str
+            ):
+                try:
+                    block["server_tool_use"]["input"] = json.loads(
+                        server_tool_use_input
+                    )
+                except (json.JSONDecodeError, TypeError):
+                    # If parsing fails, keep the original value
+                    pass
```
3coins: Can avoid some duplication here and above by extracting to a `_parse_tool_input` function.

3coins (commented):
@RohanDisa 
Thanks for submitting this fix. Left some suggestions. Would also be great to have unit/integration tests to validate these changes.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
RohanDisa: Yes, that does make sense.

libs/aws/tests/integration_tests/agents/test_bedrock_agents.py
michaelnchin, line 1: @RohanDisa, what is the reasoning for removing `test_bedrock_agents.py`?

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -2051,6 +2052,17 @@ def _lc_content_to_bedrock(
     return [block for block in bedrock_content if block.get("text", True)]
 
 
+def _parse_tool_input(tool_block: Dict[str, Any]) -> None:
+    """Normalize Bedrock tool-use block: tool_use_id -> id, parse string input."""
+    tool_block["id"] = tool_block.pop("tool_use_id", None)
+    raw_input = tool_block.get("input")
+    if raw_input is not None and isinstance(raw_input, str):
```
michaelnchin: Nit: first condition is redundant here since `isinstance(None, str)` already returns False 
```suggestion
    if isinstance(raw_input, str):
```

libs/aws/tests/integration_tests/agents/test_bedrock_agents.py
RohanDisa: Hey @michaelnchin, that was my bad, I was using an AI-assisted refactor and it incorrectly suggested removing the file. I should have reviewed that change more carefully before committing. I have restored the file and will be more careful going forward.

michaelnchin: Got it, thanks for updating.

michaelnchin:
Closing in favor of #880, see above comment.

Issue #841, opened 2026-01-19
Support Bedrock API Key as client/instance parameter (not only environment variable) for multi-tenant use cases


nathan-vm:
### Package

langchain-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a feature request and not a bug report or question

### Feature Description

Based on this issue:
[https://github.com/langchain-ai/langchain-aws/issues/582](https://github.com/langchain-ai/langchain-aws/issues/582?utm_source=chatgpt.com)

AWS_BEARER_TOKEN_BEDROCK currently works only via environment variables.
This creates a limitation for multi-tenant systems.

Because environment variables are global they cannot be safely changed per request

**Request**: 

Add support for passing the Bedrock API Key at the instance or client level, instead of relying on a global environment variable.
This would allow safe, per-request credential isolation while continuing to use Bedrock API Keys.

### Use Case

Use case: 

- The application is multi-tenant
- Each tenant has its own Bedrock API Key
- API Keys are stored in a secret manager (e.g. AWS Secrets Manager, Vault)
- The API Key is retrieved and injected per request

### Proposed Implementation (optional)

```python
def handle_request(request):
    tenant_id = request.tenant_id

    # 1. Retrieve tenant-specific API key
    bedrock_api_key = secret_manager.get(
        key=f"bedrock/api-key/{tenant_id}"
    )

    # 2. Create Bedrock LLM instance with scoped credentials
    llm = ChatBedrockConverse(
        model_id="mistral.mistral-large-2402-v1:0",
        region_name="us-east-1",
        bedrock_api_key=bedrock_api_key,   # NEW PARAM
    )

    # 3. Execute request
    response = llm.invoke(request.messages)

    return response
```

leloscheidt:
I also need this feature!

PR #849, opened 2026-01-29
feat(aws): Add API key parameter to Bedrock chat models

michaelnchin:
Related: #841 

This PR adds a new `bedrock_api_key`(or just `api_key`) parameter to ChatBedrock and ChatBedrockConverse, allowing for [Bedrock API keys](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys.html) to be passed explicitly during model instantiation. This parameter follows the equivalent `api_key` implementations in [ChatAnthropic](https://github.com/langchain-ai/langchain/blob/72571185a8f51e353a2fe9143855f310c4d31e08/libs/partners/anthropic/langchain_anthropic/chat_models.py#L804) and [ChatOpenAI](https://reference.langchain.com/python/integrations/langchain_openai/ChatOpenAI/#langchain_openai.ChatOpenAI.openai_api_key). 

Usage example:

```python
API_KEY = "your_api_key_here"

llm = ChatBedrockConverse(
    model="global.anthropic.claude-sonnet-4-5-20250929-v1:0",
    region_name="us-west-2",
    bedrock_api_key=API_KEY
)
```

A few notes:
- If `bedrock_api_key` is not specified, it will pick up the value of the `AWS_BEARER_TOKEN_BEDROCK` environmental variable, if present. (This is the existing default behavior.)
- If `bedrock_api_key` is specified AND `AWS_BEARER_TOKEN_BEDROCK` exist, the env variable will be overwritten with the value of `bedrock_api_key`.
- If both `bedrock_api_key` and AWS credentials are passed during model instantiation, the API key will take precedence.

### DIFF
--- a/libs/aws/langchain_aws/utils.py
+++ b/libs/aws/langchain_aws/utils.py
@@ -1,6 +1,6 @@
+import logging
 import os
 import re
-import warnings
 from abc import abstractmethod
 from typing import Any, Dict, Generic, Iterator, List, Literal, Optional, TypeVar, Union
 
@@ -8,6 +8,8 @@ from botocore.exceptions import BotoCoreError, UnknownServiceError
 from langchain_core.messages import AIMessage
 from packaging import version
 from pydantic import SecretStr
+
+logger = logging.getLogger(__name__)
 
 MESSAGE_ROLES = Literal["system", "user", "assistant"]
 MESSAGE_FORMAT = Dict[Literal["role", "content"], Union[MESSAGE_ROLES, str]]
@@ -136,7 +138,7 @@ def create_aws_client(
     aws_session_token: Optional[SecretStr] = None,
     endpoint_url: Optional[str] = None,
     config: Any = None,
-    bedrock_api_key: Optional[SecretStr] = None,
+    api_key: Optional[SecretStr] = None,
 ) -> Any:
     """Helper function to validate AWS credentials and create an AWS client.
 
@@ -149,7 +151,7 @@ def create_aws_client(
         aws_session_token: AWS session token.
         endpoint_url: The complete URL to use for the constructed client.
         config: Advanced client configuration options.
-        bedrock_api_key: Bedrock API key for bearer token authentication.
+        api_key: Bedrock API key for bearer token authentication.
             If provided, sets AWS_BEARER_TOKEN_BEDROCK env variable.
     Returns:
         boto3.client: An AWS service client instance.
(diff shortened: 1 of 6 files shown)


libs/aws/langchain_aws/utils.py
```diff
@@ -159,6 +163,26 @@ def create_aws_client(
...
+        has_api_key = bool(bedrock_api_key and bedrock_api_key.get_secret_value())
+
+        if has_api_key and has_aws_credentials:
+            warnings.warn(
+                "Both bedrock_api_key and AWS credentials were provided. "
+                "Using bedrock_api_key for authentication; AWS credentials "
+                "will be ignored.",
+                UserWarning,
+                stacklevel=2,
+            )
```
f4roukb: Hi there.
I'd say it's safer to crash than to just warn because of auth ambiguity.
Either way, I suggest you use `logging.warn` instead of `warnings.warn` because warnings can be silently ignored, filtered out, or disabled, while logs are more reliably captured in production environments.

f4roukb (commented):
Would appreciate it if you could make this small change. Thanks.

libs/aws/langchain_aws/utils.py
3coins: If the purpose here is to warn the developer building agents to allow either AWS credentials or API key, but not both, then `warning.warn` seems appropriate. Otherwise, agree with @f4roukb, that `logging.warn` would be best to use here.

libs/aws/langchain_aws/utils.py
```diff
@@ -135,6 +136,7 @@ def create_aws_client(
     aws_session_token: Optional[SecretStr] = None,
     endpoint_url: Optional[str] = None,
     config: Any = None,
+    bedrock_api_key: Optional[SecretStr] = None,
```
3coins: This is starting to add props specific to bedrock, than universally applicable to other services. Would using a generic `api_key` work here?

libs/aws/langchain_aws/utils.py
```diff
@@ -159,6 +163,26 @@ def create_aws_client(
...
+            warnings.warn(
+                "Both bedrock_api_key and AWS credentials were provided. "
+                "Using bedrock_api_key for authentication; AWS credentials "
+                "will be ignored.",
+                UserWarning,
+                stacklevel=2,
+            )
+
+        if has_api_key:
+            os.environ["AWS_BEARER_TOKEN_BEDROCK"] = bedrock_api_key.get_secret_value()  # type: ignore[union-attr]
```
nathan-vm: Hi @michaelnchin, thanks a lot for the implementation — it looks great overall 👍
I have one concern/question about a potential concurrency issue.

What happens if two `ChatBedrockConverse` instances are created at the same time (e.g. in a concurrent request scenario), each one with a different API key?

Since the implementation sets the API key via a global environment variable:

```python
os.environ["AWS_BEARER_TOKEN_BEDROCK"] = bedrock_api_key.get_secret_value()
```

this could lead to a race condition where `llm1` ends up using the API key from `llm2`, depending on timing. For example:

```python
API_KEY_1 = "your_api_key_here_1"
API_KEY_2 = "your_api_key_here_2"

llm1 = ChatBedrockConverse(
    model="global.anthropic.claude-sonnet-4-5-20250929-v1:0",
    region_name="us-west-2",
    bedrock_api_key=API_KEY_1
)

llm2 = ChatBedrockConverse(
    model="global.anthropic.claude-sonnet-4-5-20250929-v1:0",
    region_name="us-west-2",
    bedrock_api_key=API_KEY_2
)
```

In this scenario, is it possible for one instance to override the credentials of the other due to the shared `os.environ` state?
Just want to double-check if this is expected behavior or if there’s a plan to isolate credentials per instance.

michaelnchin: Thanks @nathan-vm, this is a valid concern. I don't believe there's a simple way to avoid this potential race condition with multiple instances, as there isn't a way to pass API keys directly to boto3 client and set them per instance (unlike other AWS credentials fields); they have to be set via the env variable. 

Alternatively - if we execute HTTP requests directly against Bedrock instead of Boto3 client, it's possible to set the API directly in headers per request. But this would require fleshing out a completely separate requests generation and response processing code path, and that level of extra complexity probably isn't worth it just for API keys support.

libs/aws/langchain_aws/utils.py
michaelnchin: Agreed - will update to `api_key`

libs/aws/langchain_aws/utils.py
michaelnchin: I concur with @3coins on the first point. For the second, will update `warnings.warn` -> `logging.warn`.

PR #784, opened 2025-12-02
feat(aws): Add Amazon Nova 2.0 support with system tools and reasoning configuration

dkovvuri:
Add comprehensive support for Amazon Nova 2.0 models including system tools
(nova_grounding and nova_code_interpreter) and reasoning configuration with
string-based validation.

Features:
- Add NovaSystemTool, NovaGroundingTool, and NovaCodeInterpreterTool helpers
- Support system tools via bind_tools() with both helper classes and strings
- Add reasoning configuration validation (accepts "low", "medium", "high")
- Support reasoning in both streaming and non-streaming modes
- Parse Nova's reasoningContent format to standard LangChain content blocks
- Add IAM permission error handling for bedrock:InvokeTool

PR #854, opened 2026-01-31
fix(aws): Ensure OpenAI tool format for ChatBedrockConverse bind_tools

### LINKED ISSUE #853: Bug: ChatBedrockConverse `bind_tools` may break LangGraph `create_agent` in langchain-aws>=v1.2.1
In #784, ChatBedrockConverse's `bind_tools` method was updated to support Nova 2.0 built-in tools, which also affected the way that regular tools were being stored in `model.kwargs`.

In `1.2.0` -> tools are converted to dictionaries (openai format) before storing them in model.kwargs
In `1.2.1` -> tools are stored as they are passed i.e `StructuredTool` in model.kwargs

This breaks the `create_react_agent` in langgraph.prebuilt using `ChatBedrockCoverse` model at https://github.com/langchain-ai/langgraph/blob/main/libs/prebuilt/langgraph/prebuilt/chat_agent_executor.py#L203

with the following error:
```
AttributeError: 'StructuredTool' object has no attribute 'get'
```

Although `create_react_agent` is deprecated.. its still being used a lot.

_Originally posted by @Prudhvi0717 in https://github.com/langchain-ai/langchain-aws/pull/784#discussion_r2711026423_

michaelnchin:
Fixes #853

Updates ChatBedrockConverse to address a regression present in langchain-aws>=1.2.1. 

#784 included a change to `bind_tools()` removing formatting of base/custom tools from `StructuredTool` to OpenAI (dictionary) format when no Nova 2.0 system tools were present, which broke compatibility with `create_agent` and the older LangGraph `create_react_agent`.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -1208,20 +1208,17 @@ class ChatBedrockConverse(BaseChatModel):
         if system_tools:
             kwargs["disable_streaming"] = True
 
-        # If we have system tools, we need to use toolConfig format
-        # Otherwise, use the old format for backward compatibility
-        if system_tools:
-            # Format custom tools using existing logic
-            formatted_custom_tools: List[Any] = []
-            for tool in custom_tools:
-                if _is_cache_point(tool):
-                    formatted_custom_tools.append(tool)
-                else:
-                    try:
-                        formatted_custom_tools.append(convert_to_openai_tool(tool))
-                    except Exception:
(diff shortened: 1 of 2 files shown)


3coins (approved):
Looks good!

Do our integration tests cover the `create_agent` from langchain agents? If so, this problem would not have surfaced with previous code, is that correct?

michaelnchin:
@3coins, thank you! At the moment, no chat model integ tests for `create_agent` (only a few for BedrockAgents classes) - will add a few in a separate PR.

Issue #867, opened 2026-02-09
`BedrockEmbeddings.aembed_documents` embeds documents as queries when embedding model is Cohere


RisaKirisu:
### Package

langchain-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code

```python
embedder = BedrockEmbeddings(model_id="us.cohere.embed-v4:0", region_name="us-west-2")
docs = ["Doc1...", "Doc2...", "Doc3...",]
embeddings = await embedder.aembed_documents(docs)
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

In main `BedrockEmbeddings`, `aembed_documents` call directly to `aembed_query` on each documents:
```python
    async def aembed_documents(self, texts: List[str]) -> List[List[float]]:
        """Asynchronous compute doc embeddings using a Bedrock model.

        Args:
            texts: The list of texts to embed

        Returns:
            List of embeddings, one for each text.

        """

        result = await asyncio.gather(*[self.aembed_query(text) for text in texts])

        return list(result)
```

However, when the embedding endpoint is Cohere, queries and documents should be handled differently, as `embed_query` and `embed_documents` do. So this is likely an oversight when the async version was written.

Specifically, Cohere embedding models has an `input_type` param that differs queries and documents. Treating both queries and documents hinders retrieval performance.

PR #881, opened 2026-02-12
fix: Bedrock Cohere uses query embedding for documents instead of document embedding

f4roukb:
Issue: https://github.com/langchain-ai/langchain-aws/issues/867

Changes:
**No breaking changes.**
- One-liner fix.
- Add unit test insuring the model invocation is performed with the right arguments.

### DIFF
--- a/libs/aws/langchain_aws/embeddings/bedrock.py
+++ b/libs/aws/langchain_aws/embeddings/bedrock.py
@@ -641,7 +641,12 @@ class BedrockEmbeddings(BaseModel, Embeddings):
             List of embeddings, one for each text.
 
         """
-        return await run_in_executor(None, self.embed_documents, texts)
+        if self._inferred_provider == "cohere":
+            return await run_in_executor(None, self.embed_documents, texts)
+
+        result = await asyncio.gather(*[self.aembed_query(text) for text in texts])
+
+        return list(result)
 
     def embed_image(self, image: MediaInput) -> List[float]:
         """Embed a single image.


libs/aws/langchain_aws/embeddings/bedrock.py
```diff
@@ -641,10 +641,7 @@ async def aembed_documents(self, texts: List[str]) -> List[List[float]]:
             List of embeddings, one for each text.
 
         """
-
-        result = await asyncio.gather(*[self.aembed_query(text) for text in texts])
-
-        return list(result)
+        return await run_in_executor(None, self.embed_documents, texts)
```
michaelnchin: One critical note - we really need to be careful about switching to `run_in_executor(None, self.embed_documents, texts)` in all cases. 

The blanket code change here is advantageous specifically for Cohere models because `embed_documents` routes inputs for this model through `_cohere_multi_embedding`, which takes advantage of the fact that Cohere [accepts an array of multiple texts in a single request](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed-v4.html) (see changes in #350 for more details). So delegating the full list to the sync embed_documents path is optimal here as it batches them into fewer API calls. 

HOWEVER, for other models like [Titan G1](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-titan-embed-text.html), only one text is allowed per request. This is why `embed_documents` instead calls `_iteratively_embed_documents`, which runs a for loop to sequentially process each text. Wrapping that in a single `run_in_executor` call would remove the concurrency that the original implementation was intentionally providing by sending up a separate request for each text in its own executor thread via `asyncio.gather`. 

TLDR; we should add a conditional in `aembed_documents` to run this only for Cohere models, and retain the old method for everything else.

```suggestion
        if self._inferred_provider == "cohere":
            return await run_in_executor(None, self.embed_documents, texts)

        result = await asyncio.gather(*[self.aembed_query(text) for text in texts])

        return list(result)
```

michaelnchin (approved):
Thanks @f4roukb !

libs/aws/langchain_aws/embeddings/bedrock.py
f4roukb: Will keep it in mind for future contributions, thanks!

Issue #806, opened 2025-12-16
AgentCoreMemorySaver: Add "end-of-workflow" checkpoint mode to reduce API call overhead


kamrul-finlexgmbh:
### Package

langgraph-checkpoint-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a feature request and not a bug report or question

### Feature Description

Add a configuration option to AgentCoreMemorySaver that allows checkpointing only at the end of workflow execution, instead of after every graph node (super-step).

Currently, LangGraph checkpoints after every node, which results in a large number of API calls to AgentCore Memory. 

For a typical 6-node conversational workflow with tool calling, this generates 62 API calls (49 createEvent + 13 listEvents) adding ~8.7 seconds of latency per user request.

### Use Case

We're building a real-time conversational agent using LangGraph with AgentCoreMemorySaver for session persistence. Our workflow:

User Message → LLM → Tool Call → LLM → Tool Call → LLM → Response
Current behavior: Checkpoint saved after each of the 6 nodes = 62 API calls = 8.7s overhead Desired behavior: Checkpoint saved once at the end = 2 API calls = ~300ms overhead For our use case:

- We don't need mid-workflow fault tolerance or recovery
- We only need the final conversation state persisted for session continuity
- Response latency is critical for user experience

### Proposed Implementation (optional)

```python
Add a checkpoint_mode parameter to AgentCoreMemorySaver:

from langgraph_checkpoint_aws import AgentCoreMemorySaver

# Current behavior (default)
checkpointer = AgentCoreMemorySaver(
    MEMORY_ID,
    region_name="us-east-1"
)

# New: checkpoint only at end of workflow
checkpointer = AgentCoreMemorySaver(
    MEMORY_ID,
    region_name="us-east-1",
    checkpoint_mode="end_of_workflow"  # or "deferred" / "batch"
)

Implementation options:
1. Buffer writes internally - Override put() and put_writes() to buffer checkpoint data, then flush only when the graph execution completes
2. Expose LangGraph's checkpoint hooks - If LangGraph supports conditional checkpointing, expose that configuration
3. Add a flush() method - Let users manually control when checkpoints are written
```

### Additional Context

Additional Context
Performance impact:

- Current (every node): 62 API calls, ~8.7s latency overhead
[...]

PR #892, opened 2026-02-21
feat(aws): Enable Buffering and Flushing Writes to Checkpoint Savers

f4roukb:
**Issue:** https://github.com/langchain-ai/langchain-aws/issues/806

**Description:** Enable buffering the write ops of **any** checkpoint saver and flushing them manually where needed (e.g., inside the graph execution), and automatically at the end of graph execution.

**Changes:**
- New `BufferedCheckpointSaver` class (decorator pattern with both sync and async API). It buffers writes and flushes them:
a) After the context is exit when a context manager is used: `aflush_on_exit` (async) or `flush_on_exit` (sync).
b) When the `aflush` (async) or `flush` (sync) method is called (potentially inside the graph). 
- Unit tests on top of Langgraph's `MemorySaver`, including tests for edge cases.
- Component tests for buffering on top of AgentCoreMemorySaver, AgentCoreValkeySaver, and DynamoDBSaver.
I essentially picked the existing component tests back up and adapted them by wrapping the instances in the buffering decorator and adapting the assertions.
I also added more fundamental component tests in each of these test suites to ensure behavior is tested on different kinds of graphs, with streaming, with parallelism, with error mid-execution, and with flush mid-execution.
- There is some redundancy across component tests. I tried to minimize it by adding conftest.py and utils.py files. I haven't modified the existing component tests. A clean-up could be scheduled if this PR is merged.
- **No breaking changes.**

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
@@ -12,7 +12,6 @@ from langgraph.checkpoint.base import (
     CheckpointTuple,
 )
 
-
 logger = logging.getLogger(__name__)
 
 
@@ -96,7 +95,7 @@ class BufferedCheckpointSaver(BaseCheckpointSaver):
         """
         if isinstance(saver, BufferedCheckpointSaver):
             raise ValueError(f"{type(self).__name__} cannot be nested")
-        
+
         super().__init__(serde=saver.serde)
 
         self._saver = saver
@@ -145,6 +144,7 @@ class BufferedCheckpointSaver(BaseCheckpointSaver):
             # Checkpoint and writes are automatically flushed on exit
             ```
         """
+
         def try_flush():
             try:
                 self.flush()
@@ -153,7 +153,7 @@ class BufferedCheckpointSaver(BaseCheckpointSaver):
                     "Failed to auto-flush buffered checkpoint and writes on exit."
                 )
                 raise
-        
+
         try:
             yield self
         except Exception:
@@ -180,6 +180,7 @@ class BufferedCheckpointSaver(BaseCheckpointSaver):
             # Checkpoint and writes are automatically flushed on exit
             ```
         """
+
         async def try_aflush():
             try:
                 await self.aflush()
@@ -188,7 +189,7 @@ class BufferedCheckpointSaver(BaseCheckpointSaver):
                     "Failed to auto-flush buffered checkpoint and writes on exit."
                 )
                 raise
-        
+
         try:
             yield self
         except Exception:
--- a/libs/langgraph-checkpoint-aws/tests/integration_tests/agentcore/conftest.py
+++ b/libs/langgraph-checkpoint-aws/tests/integration_tests/agentcore/conftest.py
@@ -32,12 +32,11 @@ def agentcore_actor_id() -> str:
 
 @pytest.fixture(scope="function")
 def agentcore_memory_saver(
-        agentcore_memory_id,
-        aws_region,
-    ) -> AgentCoreMemorySaver:
+    agentcore_memory_id,
+    aws_region,
+) -> AgentCoreMemorySaver:
     """Create AgentCoreMemorySaver instance."""
     return AgentCoreMemorySaver(
         memory_id=agentcore_memory_id,
         region_name=aws_region,
     )
-
--- a/libs/langgraph-checkpoint-aws/tests/integration_tests/agentcore/test_buffered_agentcore_memory_integration.py
+++ b/libs/langgraph-checkpoint-aws/tests/integration_tests/agentcore/test_buffered_agentcore_memory_integration.py
@@ -1,15 +1,16 @@
 """Integration tests for BufferedCheckpointSaver wrapping AgentCoreMemorySaver."""
+
 import datetime
 import operator
 from typing import Annotated, TypedDict
 
 import pytest
-from langgraph.checkpoint.base import Checkpoint, uuid6
-from langgraph.graph import END, START, StateGraph
-from langgraph.graph.state import CompiledStateGraph
 from langchain.agents import create_agent
-from langchain_core.tools import Tool
 from langchain_aws import ChatBedrock
+from langchain_core.runnables import RunnableConfig
+from langchain_core.tools import Tool
+from langgraph.checkpoint.base import Checkpoint, CheckpointMetadata, uuid6
+from langgraph.graph import END, START, StateGraph
 
 from langgraph_checkpoint_aws import AgentCoreMemorySaver, BufferedCheckpointSaver
 from tests.integration_tests.agentcore.utils import (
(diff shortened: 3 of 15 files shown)


f4roukb:
@michaelnchin, since you addressed the corresponding issue, could you please have a look? Happy to address your comments. Thank you.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
```diff
@@ -0,0 +1,397 @@
...
+                    parent_config=self._update_parent_runnable_config(buffered_config),
+                    pending_writes=pending_writes,
+                )
+
+        return self._saver.get_tuple(config)
+
+    async def aget_tuple(self, config: RunnableConfig) -> CheckpointTuple | None:
+        """Async version of get_tuple."""
+        if self._pending_checkpoint:
+            result = self.get_tuple(config)
```
michaelnchin: This doesn't seem to account for checkpoint savers doesn't implement `get_tuple` (like in `AsyncBedrockSessionSaver`). In these cases, it would fall down to `BaseCheckpointSaver.get_tuple` and [crash with NotImplementedError](https://github.com/langchain-ai/langgraph/blob/8087e6a42c62c2049a5fb3f99372a8c601d07e08/libs/checkpoint/langgraph/checkpoint/base/__init__.py#L193).

michaelnchin: Also I don't quite understand why we're calling sync `get_tuple` here. If the buffered checkpoint match condition in `get_tuple` fails, then the sync `return self._saver.get_tuple(config)` call on L307 will get executed in addition to the `self._saver.aget_tuple` call on L315.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
```diff
@@ -0,0 +1,397 @@
...
+                and (
+                    not requested_checkpoint_id
+                    or requested_checkpoint_id == buffered_checkpoint_id
+                )
+            ):
+                # Collect pending writes for this checkpoint
+                pending_writes: list[tuple[str, str, Any]] = []
+                for _, writes, task_id, _ in self._pending_writes:
+                    for channel, value in writes:
+                        pending_writes.append((task_id, channel, value))
```
michaelnchin: So, the condition here checks that the checkpoint match is the thread_id and checkpoint namespace+id. But the pending writes collected aren't getting filtered for the same matching attributes. 

What happens if the writes get buffered for multiple different threads and namespaces? They'll get bundled into the same returned CheckpointTuple despite not matching. 

We need to add some filter here to ensure that we only take the writes whose attributes match the checkpoint.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
```diff
@@ -0,0 +1,397 @@
...
+
+            with checkpointer.flush_on_exit():
+                response = graph.invoke({"messages": [...]}, config)
+            # Checkpoint and writes are automatically flushed on exit
+            ```
+        """
+        try:
+            yield self
+        finally:
+            self.flush()
```
michaelnchin: Even if there's no danger to flushing the existing buffer on exceptions, we should still log them for visibility given that the resumed checkpoint may not behave as expected due to the ignored failure.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
```diff
@@ -0,0 +1,397 @@
...
+        result = None
+
+        if self._pending_checkpoint is not None:
+            config, checkpoint, metadata, new_versions = self._pending_checkpoint
+            result = self._saver.put(config, checkpoint, metadata, new_versions)
+            self._pending_checkpoint = None
+
+        for config, writes, task_id, task_path in self._pending_writes:
+            self._saver.put_writes(config, writes, task_id, task_path)
+        self._pending_writes = []
```
michaelnchin: Clearing the `_pending_writes` after the loop concludes seems dangerous. If one of these `put_writes` calls fails midway through the loop, the successful writes for the previous iterations are persisted, but the corresponding buffer items are persisted in `_pending_writes`. Subsequent retries of `flush()` would start over from the beginning of the buffer and end up performing duplicate writes. 

So, we should probably instead be clearing the items from the buffer within the loop after the corresponding write succeeds.

Extra note: It looks like compensatory handling for existing writes exists for some of the checkpointer saver classes, but not others, so it's safer to just handle it in our code:
- AgentCoreMemorySaver: duplicates created. saver.py:282 calls `store_blob_event`, which creates a new event unconditionally.                                                                         
- AgentCoreValkeySaver: duplicates created. saver.py:609 uses `rpush` (list append), so repeated calls accumulate duplicate entries.                                                                  
- DynamoDBSaver: safe. unified_repository.py:729-740 uses ConditionExpression: attribute_not_exists(PK) and silently ignores ConditionalCheckFailedException.                                       
- AsyncBedrockSessionSaver: partially safe. async_saver.py:457-475 skips existing writes via `process_write_operations`, but still creates a redundant invocation each time.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
```diff
@@ -0,0 +1,397 @@
...
+        """Update the runnable config with the checkpoint id."""
+        configurable = config.get("configurable", {})
+        result_configurable: dict[str, Any] = {
+            "thread_id": configurable.get("thread_id"),
+            "checkpoint_ns": configurable.get("checkpoint_ns", ""),
+            "checkpoint_id": checkpoint["id"],
+        }
+        # Preserve actor_id if present (for AgentCoreMemorySaver compatibility)
+        if "actor_id" in configurable:
+            result_configurable["actor_id"] = configurable["actor_id"]
```
michaelnchin: If existing or future checkpoint saver class adds a new `configurable` key, it would get silently dropped with the hardcoded construction of `result_configurable`here. 

Should we have this preserve any unrecognized key/values as well?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
```diff
@@ -0,0 +1,397 @@
...
+            return None
+
+        result_configurable: dict[str, Any] = {
+            "thread_id": configurable.get("thread_id"),
+            "checkpoint_ns": configurable.get("checkpoint_ns", ""),
+            "checkpoint_id": parent_id,
+        }
+        # Preserve actor_id if present
+        if "actor_id" in configurable:
+            result_configurable["actor_id"] = configurable["actor_id"]
```
michaelnchin: ^ same as comment from `_update_runnable_config`

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
f4roukb: The changes now reflect my intended approach. They also address your first comment.

f4roukb:
I missed quite a few things working on this one, huh 😅 Thanks a lot for the spot-on observations. I addressed your comments and will be working on additional unit/integration tests. Should be re-reviewable by EOW.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/buffered_saver.py
f4roukb: I added some log messages there. Let me know what you think.

f4roukb:
@michaelnchin, I think it's a good milestone to get some more feedback please. I updated the PR description with more details. I can add component tests for the `AsyncBedrockSessionSaver` afterwards.

Please note that I don't have access to the necessary infrastructure to actually run the component tests. I'd really appreciate it if you could help push this. Thank you.

janukobytsch:
Any update on this? Eagerly waiting for this feature.

f4roukb:
@michaelnchin, any thoughts on this PR please? Looking forward to wrapping this up.

janukobytsch:
@michaelnchin could you take another look at this please? 🙏🙏

michaelnchin:
Superseded by #970

PR #893, opened 2026-02-23
fix(aws): Preserve metadata from ChatSagemakerEndpoint streaming

### LINKED ISSUE #888: ChatSagemakerEndpoint._stream() silently drops usage_metadata from chunks with empty content
### Description

`ChatSagemakerEndpoint._stream()` silently discards streaming chunks that contain `usage_metadata` but have empty content. This causes token usage to never be reported for streaming SageMaker calls, breaking cost tracking and usage monitoring.

### Root Cause

There are two bugs in the `_stream()` method in [`libs/aws/langchain_aws/chat_models/sagemaker_endpoint.py`](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/sagemaker_endpoint.py):

**Bug 1 — Empty-content filter discards the usage chunk (line ~408):**

```python
if message.content:  # ← falsy for empty string, drops the usage chunk
```

OpenAI-compatible inference servers (vLLM, TGI, etc.) return token usage in a **final SSE chunk with empty content**:

```json
data: {"id":"cmpl-123","choices":[],"usage":{"prompt_tokens":21,"completion_tokens":14,"total_tokens":35}}
```

A `content_handler.transform_output()` implementation can correctly parse this into an `AIMessageChunk` with `usage_metadata` and `content=""`. But since `""` is falsy, `if message.content:` prevents the chunk from ever being yielded.

**Bug 2 — `usage_metadata` is never forwarded (line ~410):**

Even if Bug 1 were fixed, the `usage_metadata` would still be lost because the `AIMessageChunk` is reconstructed without it:

```python
chunk = AIMessageChunk(content=message.content)  # ← usage_metadata not passed through
```

### Impact

Any user relying on `usage_metadata` from streaming SageMaker endpoints will see **zero token counts** on virtually all streaming calls. In our case, this caused ~95% of streaming calls to report 0 tokens, resulting in ~20x cost underreporting.

This also means `AIMessageChunk.usage_metadata` — a standard LangChain field used for cost tracking by callbacks and tracing tools — is silently broken for the SageMaker streaming path.

### Suggested Fix

Replace the filter and chunk construction to preserve `usage_metadata`:

```python
# Before (current code):
if message.content:
    if isinstance(message, AIMessage):
        chunk = AIMessageChunk(content=message.content)
        generation_chunk = ChatGenerationChunk(message=chunk)
    # ...
    yield generation_chunk

# After (proposed fix):
has_content = bool(message.content)
has_usage = getattr(message, "usage_metadata", None) is not None

if has_content or has_usage:
    if isinstance(message, AIMessage):
        chunk = AIMessageChunk(
            content=message.content or "",
            usage_metadata=getattr(message, "usage_metadata", None),
        )
        generation_chunk = ChatGenerationChunk(message=chunk)
    else:
        base_chunk = BaseMessageChunk(
            content=message.content, type=message.type
        )
        generation_chunk = ChatGenerationChunk(message=base_chunk)
    if run_manager:
        run_manager.on_llm_new_token(
            generation_chunk.text, chunk=generation_chunk
        )
    yield generation_chunk
```

The key changes:
1. Also yield chunks where `usage_metadata` is present, even if `content` is empty
2. Forward `usage_metadata` from the original message to the yielded `AIMessageChunk`

### Reproduction

```python
from langchain_aws.chat_models.sagemaker_endpoint import ChatSagemakerEndpoint

# Any OpenAI-compatible SageMaker endpoint (vLLM, TGI, etc.)
llm = ChatSagemakerEndpoint(
    endpoint_name="your-vllm-endpoint",
    streaming=True,
    content_handler=YourContentHandler(streaming=True),  # must set usage_metadata on output
    model_kwargs={"temperature": 0.0, "max_tokens": 100},
)

chunks = list(llm.stream([HumanMessage(content="What is 2+2?")]))
chunks_with_usage = [c for c in chunks if c.usage_metadata]

print(f"Chunks with usage: {len(chunks_with_usage)}")
# Expected: 1
# Actual: 0  ← usage_metadata silently dropped
```

### Environment

- `langchain-aws` version: 1.2.0 (bug also present on `main` as of 2026-02-19)
- Inference server: vLLM (OpenAI-compatible), but applies to any server returning usage in a separate final chunk
- Python 3.11

michaelnchin:
Fixes #888.

This applies the fixes to `ChatSagemakerEndpoint._stream()` suggested in the original issue:
- Relaxes the early exit condition on empty `message.content` to account for possible metadata-only messages, and
- Includes the original `usage_metadata` in the constructed `AIMessageChunk` so that streaming usage chunks are no longer silently dropped.

In addition to the above:
- The preceding conditional for stop-token enforcement will also read `usage_metadata`
- Message `response_metadata` and `id` will also be preserved
- If content handler returns `AIMessageChunk`, we'll pass it through as is (this is consistent with how ChatBedrockConverse [handles stream chunks](https://github.com/langchain-ai/langchain-aws/blob/5ab89a8b2959f2173af6f419cd3f8e22bddff116/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L1147-L1167))

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -941,7 +941,7 @@ class ChatBedrockConverse(BaseChatModel):
         if self.base_model_id:
             return self.base_model_id
 
-        # For regional model IDs (e.g., us.anthropic.claude-3-5-haiku-20241022-v1:0),
+        # For regional model IDs (e.g., us.anthropic.claude-haiku-4-5-20251001-v1:0),
         # get the base model ID by removing the regional prefix
         if self.model_id.startswith(
             ("eu.", "us.", "us-gov.", "apac.", "sa.", "amer.", "global.", "jp.", "au.")
--- a/libs/aws/langchain_aws/tools/browser_toolkit.py
+++ b/libs/aws/langchain_aws/tools/browser_toolkit.py
@@ -30,7 +30,7 @@ class BrowserToolkit:
 
             # Create a ReAct agent using the browser tools
             agent = create_agent(
-                "bedrock_converse:us.anthropic.claude-3-5-haiku-20241022-v1:0",
+                "bedrock_converse:us.anthropic.claude-haiku-4-5-20251001-v1:0",
                 tools=browser_tools
             )
 
--- a/libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
+++ b/libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
@@ -134,7 +134,7 @@ class CodeInterpreterToolkit:
 
             # Create a ReAct agent using the code interpreter tools
             agent = create_agent(
-                "bedrock_converse:us.anthropic.claude-3-5-haiku-20241022-v1:0",
+                "bedrock_converse:us.anthropic.claude-haiku-4-5-20251001-v1:0",
                 tools=code_tools
             )
 
--- a/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
@@ -430,7 +430,6 @@ def test_guardrails() -> None:
 @pytest.mark.parametrize(
     "thinking_model",
     [
-        "us.anthropic.claude-3-7-sonnet-20250219-v1:0",
         "us.anthropic.claude-sonnet-4-20250514-v1:0",
         "us.anthropic.claude-opus-4-20250514-v1:0",
         "us.anthropic.claude-opus-4-1-20250805-v1:0",
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
@@ -568,7 +568,6 @@ def test_anthropic_bind_tools_tool_choice() -> None:
 @pytest.mark.parametrize(
     "model_id",
     [
-        "anthropic.claude-3-7-sonnet-20250219-v1:0",
         "anthropic.claude-sonnet-4-20250514-v1:0",
         "anthropic.claude-sonnet-4-5-20250929-v1:0",
         "anthropic.claude-haiku-4-5-20251001-v1:0",
@@ -642,7 +641,7 @@ def test_claude_thinking_with_structured_output_ok(mock_create_aws_client) -> No
 @pytest.mark.parametrize(
     "model_id",
     [
-        "anthropic.claude-3-7-sonnet-20250219-v1:0",
+        "anthropic.claude-sonnet-4-20250514-v1:0",
         "anthropic.claude-sonnet-4-5-20250929-v1:0",
         "anthropic.claude-haiku-4-5-20251001-v1:0",
     ],
@@ -846,7 +845,7 @@ def test_beta_use_converse_api_with_inference_profile_as_nova_model(
             "us-gov-west-1",
         ),
         (
-            "us.anthropic.claude-3-7-sonnet-20250219-v1:0",
+            "us.anthropic.claude-sonnet-4-20250514-v1:0",
             None,
             "anthropic",
             nullcontext(),
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -144,7 +144,6 @@ def test_anthropic_bind_tools_tool_choice() -> None:
     "thinking_model",
     [
         "anthropic.claude-sonnet-4-5-20250929-v1:0",
-        "anthropic.claude-3-7-sonnet-20250219-v1:0",
         "anthropic.claude-sonnet-4-20250514-v1:0",
         "anthropic.claude-opus-4-20250514-v1:0",
         "anthropic.claude-haiku-4-5-20251001-v1:0",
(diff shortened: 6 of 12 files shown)


libs/aws/langchain_aws/chat_models/sagemaker_endpoint.py
```diff
@@ -403,17 +403,36 @@ def _stream(
                     and isinstance(message.content, str)
                 ):
```
3coins, line 408: Could we extract metadata once before stop token enforcement?

```python
usage_metadata = getattr(message, "usage_metadata", None)
response_metadata = message.response_metadata
msg_id = message.id

if stop and isinstance(message, AIMessage) and isinstance(message.content, str):
    text = enforce_stop_tokens(message.content, stop)
    message = AIMessage(
        content=text,
        usage_metadata=usage_metadata,
        response_metadata=response_metadata,
        id=msg_id,
    )
```

libs/aws/tests/unit_tests/chat_models/test_sagemaker_endpoint.py
3coins, line 1: All tests here assume string content, but `message.content` can be a list (e.g., multi-modal content). Consider adding a test case where message.content is a list to ensure bool(message.content) behaves correctly:

```python
  def test_stream_preserves_metadata_with_list_content() -> None:
      responses: List[BaseMessage] = [
          AIMessage(
              content=[{"type": "text", "text": "Hello"}],
              usage_metadata={"input_tokens": 5, "output_tokens": 3, "total_tokens": 8},
          ),
      ]
```

libs/aws/tests/integration_tests/chat_models/test_sagemaker_endpoint.py
```diff
@@ -55,10 +56,27 @@ def transform_output(self, output: bytes) -> AIMessageChunk:
...
-                content = response_json["choices"][0]["delta"]["content"]
 
-            return AIMessageChunk(content=content)
+            usage_metadata: UsageMetadata | None = None
+            if usage := response_json.get("usage"):
+                usage_metadata = UsageMetadata(
+                    input_tokens=usage.get("prompt_tokens", 0),
+                    output_tokens=usage.get("completion_tokens", 0),
+                    total_tokens=usage.get("total_tokens", 0),
+                )
```
3coins: Would be cleaner to extract this into a helper method.

```python
def _extract_usage(self, response_json: dict) -> UsageMetadata | None:
    if usage := response_json.get("usage"):
        return UsageMetadata(
            input_tokens=usage.get("prompt_tokens", 0),
            output_tokens=usage.get("completion_tokens", 0),
            total_tokens=usage.get("total_tokens", 0),
        )
    return None
        
```

libs/aws/tests/integration_tests/chat_models/test_sagemaker_endpoint.py
```diff
@@ -55,10 +56,27 @@ def transform_output(self, output: bytes) -> AIMessageChunk:
...
+
+            content = ""
+            choices = response_json.get("choices", [])
+            if choices and "delta" in choices[0]:
+                delta_content = choices[0]["delta"].get("content", "")
+                if delta_content and delta_content != stop_token:
+                    content = delta_content
```
3coins: Would be cleaner to extract into a helper method.

```python
def _extract_content(self, response_json: dict, stop_token: str) -> str:
    choices = response_json.get("choices", [])
    if choices and "delta" in choices[0]:
        delta_content = choices[0]["delta"].get("content", "")
        if delta_content and delta_content != stop_token:
            return delta_content
    return ""

```

libs/aws/tests/integration_tests/chat_models/test_sagemaker_endpoint.py
```diff
@@ -70,33 +88,39 @@ def __init__(self) -> None:
             + b'"model":"Qwen/Qwen2.5-1.5B-Instruct",'
             + b'"system_fingerprint":"3.0.1-native",'
             + b'"choices":[{"index":0,"delta":{"role":"assistant","content":"An"},'
-            + b'"logprobs":null,"finish_reason":null}],"usage":null}',
+            + b'"logprobs":null,"finish_reason":null}],"usage":null}\n',
```
3coins: **minor**: Keeping this in raw byte string is hard to read and maintain. Would be nice to have a helper function to simplify this. Maybe something like this:

```python
def _make_chunk(content: str, **kwargs) -> bytes:
    chunk_data = {
        "object": "chat.completion.chunk",
        "id": "",
        "created": 1742696407,
        "model": "Qwen/Qwen2.5-1.5B-Instruct",
        "system_fingerprint": "3.0.1-native",
        "choices": [{"index": 0, "delta": {"role": "assistant", "content": content}, ...}],
        **kwargs
    }
    return b'data: ' + json.dumps(chunk_data).encode() + b'\n'
```

libs/aws/tests/unit_tests/chat_models/test_sagemaker_endpoint.py
michaelnchin: Thanks - added extra test per suggestion

libs/aws/langchain_aws/chat_models/sagemaker_endpoint.py
michaelnchin: Done, also revised any other usages in `_stream`

libs/aws/tests/integration_tests/chat_models/test_sagemaker_endpoint.py
michaelnchin: Done, updated with new `_extract_usage`, `_extract_content`, and `_make_chunk` methods as suggested.

Issue #802, opened 2025-12-12
AgentCoreMemorySaver causing ValidationException on sessionId length(exceeding 100 chars)


vermaankit484:
### Package

langgraph-checkpoint-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code

```python
from langgraph_supervisor import create_supervisor
from langgraph_checkpoint_aws import AgentCoreMemorySaver
from bedrock_agentcore import BedrockAgentCoreApp

app = BedrockAgentCoreApp()

memory_id = "memory-abcdefgh"
agentcore_checkpointer = AgentCoreMemorySaver(
    memory_id=memory_id,
    region_name='us-west-2'
)

# agentcore_checkpointer = InMemorySaver(), this works well

def get_supervisor_agent():
    supervisor = create_supervisor(
            agents=agents, # Subagents passed
            model=llm_ORCHESTRATION,
            prompt=SUPERVISOR_INSTRUCTIONS
    )
  
    supervisor.compile(checkpointer=agentcore_checkpointer)


@app.entrypoint
async def endpoint(payload, context):
    # Passing session_id of length 35 while calling API endpoint
    session_id = context.session_id or payload.get("sessionId", uuid.uuid4().hex[:8])
    
    config = {"configurable": {"thread_id": session_id, "actor_id": "testmulti-agent"}}
    
    agent = get_supervisor_agent()
    async for event in agent.astream_events(
            {"messages": [{"role": "user", "content": agent_query}]}, 
            config,
            version="v2"):
        if event.get("event") == "on_chat_model_stream":
            chunk = event.get("data", {}).get("chunk")
            yield {
                "type": "chunkBlock",
                "delta": chunk,
                "sessionId": session_id
            }
```

### Error Message and Stack Trace (if applicable)

```shell
[...]

PR #895, opened 2026-02-23
fix(langgraph-checkpoint-aws): shorten session_id when checkpoint_ns is set to satisfy AWS 100-char limit

### LINKED ISSUE #782: Exception for sessionId in nested graphs in langgraph
### Package

langgraph-checkpoint-aws

### Checked other resources

### Example Code

```python
@hook_config(can_jump_to=["end"])
def before_model(self, state: AgentState, runtime: Runtime) -> dict[str, str]:
    graph_state = runtime.context
    recent_conversation_history = "\n".join([f"{m.type}: {m.content}" for m in graph_state.messages[:-1]])
    user_question = graph_state.messages[-1].content
    user_prompt = INPUT_GUARDRAIL_USER_PROMPT.format(user_question=user_question, recent_conversation_history=recent_conversation_history)
    agent = create_agent(model=self.llm, response_format=InputGuardrailResponse, system_prompt=INPUT_GUARDRAIL_SYSTEM_PROMPT)
    response = agent.invoke({"messages": [{"role": "user", "content": user_prompt}]})
    response = response.get("structured_response", InputGuardrailResponse(validity="REJECTED", reason="Failed to parse response"))
    if response.validity == "REJECTED":
        return {"messages": state.messages + [AIMessage(content=response.reason)], "jump_to": "end"}
    return None
```

```shell
File "/Users/admintemp/Desktop/code/llm/repos/llmntal/.venv/lib/python3.13/site-packages/langgraph_checkpoint_aws/agentcore/saver.py", line 96, in get_tuple
    events = self.checkpoint_event_client.get_events(
        checkpoint_config.session_id,
    ...<2 lines>...
        self.max_results,
    )
  File "/Users/admintemp/Desktop/code/llm/repos/llmntal/.venv/lib/python3.13/site-packages/langgraph_checkpoint_aws/agentcore/helpers.py", line 225, in get_events
    response = self.client.list_events(**params)
  File "/Users/admintemp/Desktop/code/llm/repos/llmntal/.venv/lib/python3.13/site-packages/botocore/client.py", line 602, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/admintemp/Desktop/code/llm/repos/llmntal/.venv/lib/python3.13/site-packages/botocore/context.py", line 123, in wrapper
    return func(*args, **kwargs)
  File "/Users/admintemp/Desktop/code/llm/repos/llmntal/.venv/lib/python3.13/site-packages/botocore/client.py", line 1078, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the ListEvents operation: 2 validation errors detected: Value at 'sessionId' failed to satisfy constraint: Member must have length less than or equal to 100; Value at 'sessionId' failed to satisfy constraint: Member must satisfy regular expression pattern: [a-zA-Z0-9][a-zA-Z0-9-_]*
```

### Description

Similar to https://github.com/langchain-ai/langchain-aws/issues/740

Since Langchain added middleware, we can call functions before or after agent calls with a hook signature.
The checkpoint then produces a sessionId that looks like this for example:
'brain_873b857b-9daf-96a1-0931-73720fa21504_InputGuardrail.before_model_c6185429-e015-b70e-6887-328ed79ca85c'

We then get a validation error in boto for the point (...Guardrail *.* before_model...)

Adding in agentcore->models.py:
checkpoint = self.checkpoint_ns.replace(":", "_").replace("|", "_").replace(".", "_")

Fixes that issue, but the sessionId then becomes too long (>100 len), resulting in yet another validation error

jameel-hussain:
Fixes #782

## Summary
When using subgraphs, `checkpoint_ns` can grow (e.g. nested namespaces), so `session_id` (`thread_id` + `_` + `checkpoint_ns`) can exceed AWS’s 100-character limit and trigger validation errors. This change shortens `session_id` only when `checkpoint_ns` is set, using a deterministic SHA-256 hex digest (64 chars), so the same logical session always maps to the same id and stays within the limit.

## Why
- AWS rejects session IDs longer than 100 characters.
- With subgraphs, the namespace in `session_id` can make the string too long.
- Shortening only when `checkpoint_ns` is non-empty keeps the existing behavior for the common case (no namespace) and fixes the subgraph case without changing the public API.

## Changes
- **`CheckpointerConfig`** (agentcore): Added private `_shorten_session_id()`; `session_id` property returns the shortened id only when `checkpoint_ns` is set, otherwise returns `thread_id` unchanged.
- Unit tests updated to expect the shortened `session_id` when `checkpoint_ns` is set (deterministic hash).

## Review focus
- Shortening is applied only when `checkpoint_ns` is truthy; no change when it’s empty.
- SHA-256 hex is deterministic and 64 chars; no new dependencies.

---
*This contribution was prepared with assistance from AI tooling.*

### DIFF
--- a/libs/langgraph-checkpoint-aws/pyproject.toml
+++ b/libs/langgraph-checkpoint-aws/pyproject.toml
@@ -39,6 +39,7 @@ test = [
     "pytest-socket>=0.7.0",
     "pytest-asyncio>=0.26.0",
     "pytest-mock>=3.15.1",
+    "pytest-xdist>=3.8.0,<4.0.0",
     "fakeredis>=2.25.1"
 ]
 test_integration = [
--- a/libs/langgraph-checkpoint-aws/uv.lock
+++ b/libs/langgraph-checkpoint-aws/uv.lock
@@ -52,30 +52,29 @@ wheels = [
 
 [[package]]
 name = "boto3"
-version = "1.42.52"
+version = "1.42.56"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "botocore" },
     { name = "jmespath" },
     { name = "s3transfer" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/41/ed/8eacb8ec7bf264079608be5f9a2a57e31e7fed7a791bb3b15500ca9274a5/boto3-1.42.52.tar.gz", hash = "sha256:ff4a4afb832f63a1358e11fe6eb321da0f4767979c6721dd32fb02e6eabcebf5", size = 112811, upload-time = "2026-02-18T21:54:57.804Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/2d/2a/de34ad6c43c56fe6dd5824bff2cd7fdef5edd9de0617cbd217040318ba97/boto3-1.42.52-py3-none-any.whl", hash = "sha256:7b3e0c4bfd8815a3df64fbe98fc9f87dfb12bd7a783cf63dfc2f166c66798c9d", size = 140556, upload-time = "2026-02-18T21:54:56.609Z" },
+    { url = "https://files.pythonhosted.org/packages/59/bb/a7fd6f4f38cf0f6cd91c152442cbe5b5f270275a60a5c7147ae5560f2f25/boto3-1.42.56-py3-none-any.whl", hash = "sha256:e09f91170b19e0ecd5e514321d99323be3e91a58dfa73fd58cf73b7b8b179c84", size = 140556, upload-time = "2026-02-24T20:28:49.815Z" },
 ]
 
 [[package]]
 name = "boto3-stubs"
-version = "1.42.51"
+version = "1.42.56"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "botocore-stubs" },
     { name = "types-s3transfer" },
     { name = "typing-extensions", marker = "python_full_version < '3.12'" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/1c/69/41759e8fd0b256da1c46a8ef8336265dd6f3c40e974ef9413bcded8e2603/boto3_stubs-1.42.51.tar.gz", hash = "sha256:0fdc9d7a43ec5790c0067cfa037f4018f9927c287fdf2f9cd5c9849173656ec0", size = 100532, upload-time = "2026-02-17T21:34:51.45Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/3d/a1/23cdd844dc940ef9c77ee1fc95cbf4c0a3f4ee53ecb3c2bbf7ba9932d2f7/boto3_stubs-1.42.56.tar.gz", hash = "sha256:95e4eab435b1b8e05e0d977c76f706efcded2b9a375cbc45b031c7458d100b25", size = 100972, upload-time = "2026-02-24T21:06:18.112Z" }
 wheels = [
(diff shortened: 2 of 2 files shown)


jameel-hussain:
fixes #782 and #802

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/models.py
```diff
@@ -15,13 +16,24 @@ class CheckpointerConfig(BaseModel):
...
+        is set to stay under AWS session_id length limit (100 characters).
+        """
+        return hashlib.sha256(full_id.encode()).hexdigest()
+
     @property
     def session_id(self) -> str:
         """Generate session ID from thread_id and checkpoint_ns."""
         if self.checkpoint_ns:
             # Use underscore separator to ensure valid session ID pattern
             checkpoint = self.checkpoint_ns.replace(":", "_").replace("|", "_")
```
michaelnchin: We can remove the old `.replace` sanitization here, since the new SHA-256 conversion will only return alphanumeric chars

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/models.py
```diff
@@ -15,13 +16,24 @@ class CheckpointerConfig(BaseModel):
...
-            return f"{self.thread_id}_{checkpoint}"
+            full_id = f"{self.thread_id}_{checkpoint}"
+            return self._shorten_session_id(full_id)
         return self.thread_id
```
michaelnchin, line 39: Curious - any need to also validate/shorten `thread_id`? Can it also exceed 100 chars by itself if `checkpoint_ns` isn't present?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/models.py
```diff
@@ -15,13 +16,24 @@ class CheckpointerConfig(BaseModel):
...
+
     @property
     def session_id(self) -> str:
         """Generate session ID from thread_id and checkpoint_ns."""
         if self.checkpoint_ns:
             # Use underscore separator to ensure valid session ID pattern
             checkpoint = self.checkpoint_ns.replace(":", "_").replace("|", "_")
-            return f"{self.thread_id}_{checkpoint}"
+            full_id = f"{self.thread_id}_{checkpoint}"
+            return self._shorten_session_id(full_id)
```
michaelnchin: Let's add a logger statement here in case users need to trace the original `checkpoint_ns` + `thread_id`.

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/models.py
jameel-hussain: the reason I am keeping it as is because that will be minimum change in the existing flow, if I hash the thread id then I will have to update the delete methods as well, and if someone pass thread id that fails validation error, the user can still update, fix the validation and get it working, you cannot control the namespace though.

jameel-hussain:
@michaelnchin I have resolved the linter errors, added log for the full id requested, and removed the sanitize step, please take a look, let me know if there is anything else you want me to do, Thanks!

michaelnchin (commented):
Integ tests passing after merging fix from #899: https://github.com/langchain-ai/langchain-aws/actions/runs/22382748063

ersksumit1:
Hey @michaelnchin, when we can expect release of this fix?

michaelnchin:
@ersksumit1 we can probably get one out by Friday this week - will be in `1.0.6`

PR #904, opened 2026-02-28
fix: add optional sanitization for Bedrock tool_call serialization bug

AmineIzanami:
## Summary

- Adds optional checkpoint sanitization to `AgentCoreMemorySaver` that fixes Bedrock Converse API tool_call serialization bug
- When checkpoints containing `tool_use` blocks are saved, the `input`/`args` fields may be serialized as JSON strings instead of dicts, causing `ValidationException` on resume
- New `sanitize_tool_calls` parameter (default `True`) sanitizes tool_call args on checkpoint load

## Problem

When using `AgentCoreMemorySaver` with AWS Bedrock Converse API, checkpoints containing `tool_use` blocks may have their `input` field serialized as a JSON string instead of a dict. When these checkpoints are loaded and the conversation is resumed, Bedrock rejects the request with:

```
ValidationException: 1 validation error detected: Value '{"name":"test"}' at
'messages.2.member.content.1.member.toolUse.input' failed to satisfy constraint:
Member must be a map (JSON object)
```

### Reproduction

1. Create an agent with `AgentCoreMemorySaver` and Bedrock LLM (`ChatBedrockConverse`)
2. Invoke agent with a request that triggers a tool call
3. Use HITL (`interrupt()`) to pause execution — checkpoint saved automatically
4. Resume with `Command(resume=...)` — **fails** with `ValidationException`

## Solution

Defense-in-depth fix that sanitizes checkpoints on load:

1. **New module `sanitization.py`** — standalone sanitization functions that:
   - Parse JSON string tool args back to dicts (`_parse_tool_args`)
   - Sanitize `tool_calls[].args`, `content[].input`, and `additional_kwargs` formats
   - Handle both LangChain and Bedrock tool call formats
2. **Modified `AgentCoreMemorySaver`**:
   - Added `sanitize_tool_calls: bool = True` parameter
   - `get_tuple()` and `list()` apply sanitization before returning
   - `aget_tuple()` covered via delegation to `get_tuple()`

## Changes

- `langgraph_checkpoint_aws/sanitization.py` — new sanitization module
- `langgraph_checkpoint_aws/agentcore/saver.py` — added sanitization support
- `tests/unit_tests/test_sanitization.py` — 22 unit tests including integration scenario

## Testing

- [x] Unit tests for all sanitization functions
- [x] Integration test simulating real Bedrock HITL checkpoint roundtrip with the bug
- [x] Test that already-correct checkpoints pass through unchanged (no performance impact)
- [x] Test that sanitization is non-destructive (original checkpoint not modified)

## Related

- Bedrock Converse API tool_use serialization (root cause is in Bedrock's serialization layer)
- LangGraph HITL with nested agents

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
@@ -100,7 +100,7 @@ class AgentCoreMemorySaver(BaseCheckpointSaver[str]):
             return result
         sanitized = sanitize_checkpoint(result.checkpoint)
         if sanitized is not result.checkpoint:
-            return result._replace(checkpoint=sanitized)
+            return result._replace(checkpoint=cast(Checkpoint, sanitized))
         return result
 
     def get_tuple(
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/agentcore/saver.py
@@ -40,6 +40,7 @@ from langgraph_checkpoint_aws.agentcore.models import (
     WriteItem,
     WritesEvent,
 )
+from langgraph_checkpoint_aws.sanitization import sanitize_checkpoint
 
 RunnableConfigDict: TypeAlias = dict[str, Any]
 
@@ -53,6 +54,9 @@ class AgentCoreMemorySaver(BaseCheckpointSaver[str]):
     Args:
         memory_id: the ID of the memory resource created in AgentCore Memory
         serde: serialization protocol to be used. Defaults to JSONPlusSerializer
+        sanitize_tool_calls: if True, sanitize tool_call args on checkpoint load.
+            Fixes Bedrock Converse API serialization bug where tool_call args are
+            stored as JSON strings instead of dicts. Default: True
         limit: maximum number of events to parse from ListEvents.
         max_results: maximum number of results to retrieve from AgentCore Memory.
         max_retries: maximum number of retry attempts for retryable errors.
@@ -65,6 +69,7 @@ class AgentCoreMemorySaver(BaseCheckpointSaver[str]):
         memory_id: str,
         *,
         serde: SerializerProtocol | None = None,
+        sanitize_tool_calls: bool = True,
         limit: int | None = None,
         max_results: int | None = 100,
         max_retries: int = DEFAULT_MAX_RETRIES,
@@ -75,6 +80,7 @@ class AgentCoreMemorySaver(BaseCheckpointSaver[str]):
         super().__init__(serde=serde)
 
         self.memory_id = memory_id
+        self._sanitize_tool_calls = sanitize_tool_calls
         self.limit = limit
         self.max_results = max_results
         self.serializer = EventSerializer(self.serde)
@@ -87,6 +93,15 @@ class AgentCoreMemorySaver(BaseCheckpointSaver[str]):
             **boto3_kwargs,
         )
         self.processor = EventProcessor()
+
+    def _maybe_sanitize(self, result: CheckpointTuple) -> CheckpointTuple:
+        """Apply checkpoint sanitization if enabled."""
+        if not self._sanitize_tool_calls:
+            return result
+        sanitized = sanitize_checkpoint(result.checkpoint)
+        if sanitized is not result.checkpoint:
+            return result._replace(checkpoint=sanitized)
+        return result
 
     def get_tuple(
         self,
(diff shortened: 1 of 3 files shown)


3coins (commented):
@AmineIzanami 
Thanks for submitting this PR. It seems like the actual root cause of this has been fixed in #880 which has been merged. The PR here will only attempt to fix existing tool call loads which has been incorrectly serialized before the fix in #880, these will be a rather small edge cases and does not warrant a sanitization logic for all future loads.

Can you confirm that you are able to replicate the issue with latest updates in main? So we can handle appropriately, would also help to know where exactly do these string args appear?

- During streaming accumulation? (Expected, fixed by #880)
- After checkpoint load? (Need to investigate why #880 isn't working)
- In old checkpoints from before #880? (Migration issue, not ongoing bug)

Here is the investigation I did so far:

#### 1. Is AgentCoreMemorySaver Actually Causing Double Serialization?

**NO**. I tested the full serialization roundtrip:

```python
# Original message with dict args
msg = AIMessage(
    content='Creating product',
    tool_calls=[{'name': 'create_product', 'args': {'name': 'Latte', 'price': 25}, 'id': 'call_123'}]
)

# After EventSerializer roundtrip
# Result: {'name': 'Latte', 'price': 25} - still a dict
```
The EventSerializer correctly preserves dict types. **No double serialization is occurring.**

#### 2. Where Do String Args Actually Come From?

**From Bedrock's streaming API**, not from checkpoint serialization:

1. User sends message to agent
2. ChatBedrockConverse streams response from Bedrock
3. **Bedrock streams tool input as string fragments** (this is Bedrock's design)
4. LangChain accumulates fragments into `content[].tool_use.input` as a **string**
5. This AIMessage (with string input) gets checkpointed
6. Checkpoint is loaded later
7. Message is sent to ChatBedrockConverse
8. **PR #880's fix converts string → dict** when sending to Bedrock

kimnamu (commented):
Thanks for this, @AmineIzanami — and for the detailed repro in the description. I'm not a maintainer, just another Bedrock user who ran your branch locally and wanted to share what I found, in case it's useful for the discussion @3coins started.

**On the scoping question @3coins raised:** I confirmed #880 (merged 2026-02-18, "convert string tool_use input to dict in `_lc_content_to_bedrock`") fixes the root cause on the *send* path, so newly-streamed checkpoints round-trip fine. That makes this PR a defense-in-depth / migration layer for checkpoints written *before* #880 — so it's a narrower scope than "all future loads." That's the maintainer's call, and I'd defer to it.

**If this does move forward, two things I'd flag from running the branch:**

**1. `sanitize_message` silently drops `name`, `usage_metadata`, and `invalid_tool_calls`.** The rebuild constructs a fresh `AIMessage` with only `content / tool_calls / additional_kwargs / response_metadata / id`, so any other field is lost — and this only fires on the exact messages the fix targets. Reproduced on your branch:
```
BEFORE: name='agent_a'  usage_metadata={'input_tokens':10,'output_tokens':4,'total_tokens':14}  invalid_tool_calls=[{...}]
AFTER : name=None       usage_metadata=None                                                     invalid_tool_calls=[]
```
`usage_metadata` loss breaks token/cost tracking on resume; `name` loss breaks multi-agent routing. `model_copy(update=...)` preserves everything and stays non-destructive (verified the original message is untouched):
```python
if needs_sanitization:
    return msg.model_copy(update={
        "content": sanitized_content,
        "tool_calls": sanitized_tool_calls or msg.tool_calls,
        "additional_kwargs": sanitized_kwargs,
    })
```
I ran this against the field-loss case above: `args` becomes a dict ✅, and `name` / `usage_metadata` / `invalid_tool_calls` are all preserved ✅, original unmutated ✅. A regression test asserting `name`/`usage_metadata` survive sanitization would lock it in — the current 26 tests don't set those fields, so the loss passes silently.

**2. (not blocking, just flagging) `_parse_tool_args(None) -> {}`.** A `None` arg becomes `{}`. For real Bedrock tool calls that's the right normalization, but it's a behavior change worth a one-line comment so a future reader doesn't read it as accidental.

**Rebase note:** this branch predates #1067 ("Reorganize for logical package structure per service"), which moved `agentcore/saver.py` → `checkpoint/agentcore/saver.py`. The diff won't apply to current `main` as-is — the `AgentCoreMemorySaver` edits now belong under `libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore/saver.py`, and the new module would live in that tree too. Worth a rebase + re-running `make test` before re-review.

Verified locally on `9a0a491`: `26 passed`, clean tests. The field-preservation fix above is the one thing I'd want before it lands.

*This review was prepared with the help of an AI agent (Claude Code); I reviewed the branch, ran the code, and verified the outputs above myself before posting.*

PR #907, opened 2026-03-02
fix(aws): correct Cohere embedding batch size calculation

rowan-walshe:
In https://github.com/langchain-ai/langchain-aws/pull/376, `_batch_cohere_embedding_texts` was added to account for request size limits. However, it treated the per text character limit (2048 for V3, 512k for V4) as a cumulative limit for a batch. This causes unnecessary batch splits, resulting in reduced performance due to excess API calls. The exact performance difference users will see depends on many factors, but from my testing, I saw a ~8x speedup for `BedrockEmbeddings.embed_documents`. We were embedding many documents as part of a larger task. The whole task previously took ~80 minutes. Just this change brought it down to ~20.

The AWS docs ([1](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed-v3.html), [2](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-embed-v4.html)) in this area are not particularly clear, so it's understandable that you might interpret the character limits as per request rather than per text. However, a simple test verifies that the API is happy to accept 96 2048-character texts for Cohere Embed V3. It's harder to verify the max limit for V4, since a max-size request can easily exceed the default AWS Service Quota for this model; however, it is possible to verify that the API will accept more than 512_000 characters, suggesting that, like for the V3 model, the character limits are per text rather than per request.

This change validates per text limits upfront, then batches by item count only (max 96 texts per batch).

### DIFF
--- a/libs/aws/tests/integration_tests/embeddings/test_bedrock_embeddings.py
+++ b/libs/aws/tests/integration_tests/embeddings/test_bedrock_embeddings.py
@@ -166,8 +166,7 @@ def test_bedrock_cohere_batching() -> None:
     # Test large character batch
     large_char_batch = ["foo", "bar", "a" * 2045, "baz"]
     assert list(bedrock._batch_cohere_embedding_texts(large_char_batch)) == [
-        ["foo", "bar"],
-        ["a" * 2045, "baz"],
+        ["foo", "bar", "a" * 2045, "baz"],
     ]
 
     # Should be fine with exactly 2048 characters


michaelnchin (commented):
Thanks for the contribution @rowan-walshe, changes look good!

Just need one adjustment to the Cohere batching integration test [here](https://github.com/rowan-walshe/langchain-aws/blob/6488f79b0cd9eb18b1dc83db27dbcaebe9fb03d6/libs/aws/tests/integration_tests/embeddings/test_bedrock_embeddings.py#L166-L171) to ensure it follows the new behavior.

PR #909, opened 2026-03-03
feat(aws): Support bind_tools in ChatSagemakerEndpoint

f4roukb:
Issue: https://github.com/langchain-ai/langchain-aws/issues/617
NOTE: The issue requests this feature in `SagemakerEndpoint`, but I think it makes much more sense to have it in `ChatSagemakerEndpoint`. 

Motivation: Enable developing agents using custom models deployed behind Sagemaker Endpoints.

Description: Support `bind_tools` in `ChatSagemakerEndpoint`. Uses OpenAI-compatible format for tool calls and tool responses since this format is already quite popular in custom models.

Changes:
- Implemented `bind_tools` inherited from `BaseChatModel`.
- Implemented an OpenAI-compatible ContentHandler, and use it by default. This is not a breaking change since the field was previously mandatory. Can change if needed.
- Enabled proper conversions for both streaming and non-streaming:
1. LangChain tool output <-> Sagemaker Endpoint API payload
2. Sagemaker Endpoint API payload <-> LangChain tool call (or chunks)
- **No breaking changes**


f4roukb:
@michaelnchin, would you mind having a look at your convenience? Happy to address your comments. Thank you.

JiwaniZakir (commented):
In `_parse_openai_style_tool_calls`, the `json.loads(args)` call on line ~196 has no error handling for malformed JSON strings. The guard `if args else {}` only covers the empty-string case, so a non-empty but invalid JSON string (e.g., `"{"`) will raise an uncaught `json.JSONDecodeError` that bubbles up through `transform_output` with a confusing traceback. A `try/except json.JSONDecodeError` with a logged warning and a fallback (e.g., keeping `args` as a raw string or raising a more descriptive `ValueError`) would make failures easier to diagnose.

Additionally, `content_handler: ChatModelContentHandler = Field(default=OpenAICompatibleChatModelContentHandler())` uses a module-level instance as the default. Even though the comment notes it's stateless today, Pydantic convention is `default_factory=OpenAICompatibleChatModelContentHandler` to avoid subtle issues if the class ever acquires mutable state and to align with how other mutable-typed fields are handled elsewhere in the codebase. Making `content_handler` optional with a default is also a behavioral change for existing users who previously received a validation error when omitting it — worth a note in the changelog or docstring.

libs/aws/langchain_aws/chat_models/sagemaker_endpoint.py
```diff
@@ -108,6 +127,138 @@ class ChatModelContentHandler(ContentHandlerBase[Any, Any]):
...
+    ) -> List[ToolCall]:
+        """Parse tool calls from OpenAI-compatible response format."""
+        # docs: https://developers.openai.com/api/reference/resources/chat/?utm_source=chatgpt.com
+        if not tool_calls_data:
+            return []
+
+        tool_calls: List[ToolCall] = []
+        for tc in tool_calls_data:
+            func = tc.get("function") or {}
+            args = func.get("arguments", {})
```
kimnamu, line 215: Malformed/non-JSON `arguments` (common from custom/quantized models) crashes here with an uncaught `JSONDecodeError` — the `if args` guard only handles the empty string. Fail-soft keeps the raw string instead of killing the turn:
```suggestion
            args = func.get("arguments", {})
            if isinstance(args, str):
                if args:
                    try:
                        args = json.loads(args)
                    except json.JSONDecodeError:
                        logger.warning(
                            "Tool call arguments are not valid JSON; keeping raw string: %s",
                            args,
                        )
                        args = {"__raw_arguments__": args}
                else:
                    args = {}
```

libs/aws/langchain_aws/chat_models/sagemaker_endpoint.py
```diff
@@ -492,9 +695,28 @@ def _messages_to_sagemaker(
...
+        elif isinstance(msg, ToolMessage):
+            sagemaker_messages.append(
+                {
+                    "role": "tool",
+                    "tool_call_id": msg.tool_call_id,
+                    "content": content,
+                    "artifact": msg.artifact,
+                    "status": msg.status,
+                }
+            )
```
kimnamu, line 730: `artifact`/`status` aren't part of the OpenAI tool-role schema this handler targets, and `artifact` is `Any` — a non-serializable value (datetime, numpy array) makes the whole request die in `transform_input`'s `json.dumps` with an opaque `TypeError`. Emitting only the OpenAI-compatible fields (like the other roles do) avoids both:
```suggestion
        elif isinstance(msg, ToolMessage):
            sagemaker_messages.append(
                {
                    "role": "tool",
                    "tool_call_id": msg.tool_call_id,
                    "content": content,
                }
            )
```

kimnamu (commented):
Thanks for adding this, @f4roukb — `bind_tools` on `ChatSagemakerEndpoint` is genuinely useful for custom-model agents, and the "default OpenAI-compatible handler is non-breaking since the field was mandatory" framing is correct. I'm not a maintainer, just another Bedrock/SageMaker user — **I only reviewed the tool-serialization/streaming paths** (`function_calling.py`, `_parse_openai_style_*`, `_messages_to_sagemaker`), not the full diff. I checked out the branch, ran the 68 unit tests (all green), plus the cases below.

**1. Malformed tool-call `arguments` raises an uncaught `JSONDecodeError`** (confirming @JiwaniZakir). Custom/quantized models emitting OpenAI-style tool calls frequently return truncated or non-JSON `arguments`. On this branch:
```python
resp = {"choices":[{"message":{"tool_calls":[{"id":"c","function":{"name":"foo","arguments":"{"}}]}}]}
handler._parse_openai_style_response(resp)
# JSONDecodeError: Expecting property name enclosed in double quotes (char 1)
```
The `if args else {}` guard only covers the empty-string case. A fail-soft keeps the raw string for debugging instead of crashing the whole turn (inline suggestion below).

**2. `ToolMessage` serialization leaks non-OpenAI keys and can crash `transform_input`.** `_messages_to_sagemaker` emits `artifact` and `status` on the tool message:
```python
_messages_to_sagemaker([ToolMessage(content="ok", tool_call_id="c2")])
# [{'role':'tool','tool_call_id':'c2','content':'ok','artifact': None, 'status':'success'}]
```
The OpenAI tool-role schema this handler targets only has `role`/`tool_call_id`/`content`, so `artifact`/`status` are extraneous for the endpoint. More importantly, `ToolMessage.artifact` is `Any` — a non-serializable artifact (a `datetime`, a numpy array) makes the *whole request* die in `transform_input`'s `json.dumps` with an opaque `TypeError: Object of type datetime is not JSON serializable`, not a domain error. Every other role here emits only OpenAI-compatible fields; this is the one place that doesn't. Inline suggestion below. The existing `test_tool_message_conversion` only asserts the three keys are present, so it passes either way — worth adding `assert "artifact" not in result[0]` to pin the OpenAI-compat contract.

Both are input-robustness edges that won't show up on the happy path — not blocking. Lovely, well-scoped feature otherwise.

*This review was prepared with the help of an AI agent (Claude Code); I reviewed and reproduced every claim locally before posting.*

f4roukb:
Thank you @JiwaniZakir and @kimnamu. I've addressed the concerns.

  1. `_parse_openai_style_tool_calls` now catches `json.JSONDecodeError` on malformed tool-call arguments, logs a warning, and falls back to `{"__raw_arguments__": <raw string>}` instead of raising.
  2. `_messages_to_sagemaker` no longer emits `artifact`/`status` on tool messages. Only the OpenAI-compatible `role`/`tool_call_id`/`content` fields, matching every other role.

f4roukb:
@michaelnchin, could you please have a look at your convenience? Thank you.

kimnamu (commented):
Thanks @f4roukb for the fast turnaround on both points, and for pinning them with tests: the exact-dict assertion in `test_tool_message_conversion` and the `datetime` artifact case are exactly the guards I had in mind.

Re-verified on `eabcf2a`: both fixes land as described and the new tests bite. Nothing further from my side.

| Check (from my review) | `b37d8c5` | `eabcf2a` |
|---|---|---|
| `arguments: "{"` in a tool call | `JSONDecodeError` from `_parse_openai_style_response` | `args={"__raw_arguments__": "{"}` + one `logger.warning` |
| `ToolMessage(artifact=datetime(...))` through `transform_input` | `TypeError: Object of type datetime is not JSON serializable` | `{"role": "tool", "tool_call_id": "c2", "content": "ok"}` |
| New tests with the two source hunks reverted | 3 failed | 3 passed |
| `tests/unit_tests` (`--disable-socket`) | | 1210 passed, 3 skipped, 1 xpassed |
| `ruff check` / `ruff format --diff` / `mypy` | | clean |

<details><summary>Red/green output</summary>

```
# source hunks reverted to b37d8c5 behaviour, tests from eabcf2a
FAILED ...::test_parse_openai_style_tool_calls_malformed_json_arguments
  json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
FAILED ...::TestMessagesToSagemaker::test_tool_message_conversion
FAILED ...::TestMessagesToSagemaker::test_tool_message_conversion_drops_non_openai_fields
  AssertionError: assert 'artifact' not in {'role': 'tool', 'tool_call_id': 'call_456', 'content': 'ok', 'artifact': datetime.datetime(2024, 1, 1, 0, 0), ...}

# eabcf2a, same selection
3 passed, 67 deselected in 0.31s
```
</details>

One item from @JiwaniZakir's review still looks open, not blocking: `content_handler` keeps `Field(default=OpenAICompatibleChatModelContentHandler())`, and that instance is shared (`a.content_handler is b.content_handler` is `True` across two `ChatSagemakerEndpoint`s). Harmless while the handler is stateless; `default_factory=OpenAICompatibleChatModelContentHandler` would make that explicit.

PR #911, opened 2026-03-04
docs(aws): clarify ChatBedrock Anthropic max_tokens default

carlosan1708:
## Summary

Adds a clarifying note to `ChatBedrock` documentation that explains Anthropic behavior on the Bedrock `InvokeModel` path.

## Why

Users may assume `max_tokens` is omitted when they do not pass it explicitly. In the legacy `InvokeModel` path (`beta_use_converse_api=False`) for Anthropic models, LangChain injects `max_tokens=1024` by default. This can affect generation length, cost, and debugging expectations.

Making this behavior explicit in the class docstring improves discoverability and helps avoid confusion when comparing request payloads vs Converse paths which uses the maximum allowed value for the model..

## What changed

- Updated `ChatBedrock` class docstring in:
  - `libs/aws/langchain_aws/chat_models/bedrock.py`
- No runtime logic changes.

### DIFF
--- a/libs/aws/langchain_aws/embeddings/bedrock.py
+++ b/libs/aws/langchain_aws/embeddings/bedrock.py
@@ -767,12 +767,12 @@ def _batch_cohere_embedding_texts(
 ) -> Generator[List[str], None, None]:
     """Batches a set of texts into chunks acceptable for the Cohere embedding API.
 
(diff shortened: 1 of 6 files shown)


carlosan1708:
@michaelnchin Great call, I have moved the docstring to the suggested place.

michaelnchin (approved):
Thanks @carlosan1708 !

PR #918, opened 2026-03-11
feat(aws): add ChatBedrockNovaSonic using Bidirectional Streaming API  (#785)

### LINKED ISSUE #785: Using the Bidirectional Streaming API
### Package

langchain-aws

### Checked other resources

### Feature Description

I would like to use langchain-aws to communicate with Nova 2 Sonic Model. 

From the official doc: https://docs.aws.amazon.com/nova/latest/userguide/speech-bidirection.html, it requires to call InvokeModelWithBidirectionalStream API. Wondering if it's supported in langchain-aws.

Thanks

### Use Case

for using Nova 2 Sonic model

```python

```

### Additional Context

https://docs.aws.amazon.com/nova/latest/userguide/speech-bidirection.html

Keyur-S-Patel:
Closes #785

### Summary

Adds `ChatBedrockNovaSonic`, a LangChain chat model for Amazon Nova Sonic's bidirectional streaming speech-to-speech API. Also adds `create_aws_bedrock_runtime_client` utility for the smithy-based `aws-sdk-bedrock-runtime` SDK, following the same patterns as `create_aws_client` for boto3.

### `create_aws_bedrock_runtime_client` utility

New utility in `utils.py` that mirrors `create_aws_client` but targets the smithy-based SDK required for bidirectional streaming. Supports the same credential options:

```python
from langchain_aws.utils import create_aws_bedrock_runtime_client

# Explicit keys
client = create_aws_bedrock_runtime_client(
    region_name="us-east-1",
    aws_access_key_id=SecretStr("AKIA..."),
    aws_secret_access_key=SecretStr("..."),
)

# Profile-based (resolves via boto3.Session)
client = create_aws_bedrock_runtime_client(
    credentials_profile_name="my-profile",
)

# API key (bearer token)
client = create_aws_bedrock_runtime_client(
    region_name="us-east-1",
    api_key=SecretStr("bedrock-api-key"),
)
```

### `ChatBedrockNovaSonic` chat model

Follows `ChatBedrockConverse` patterns for AWS credential fields and `validate_environment`. Aligned with `BaseChatModel` contract.

```python
from langchain_aws import ChatBedrockNovaSonic

model = ChatBedrockNovaSonic(
    model_id="amazon.nova-sonic-v1:0",
    region_name="us-east-1",
    voice_id="matthew",
)

# Text conversation via LangChain interface
response = await model.ainvoke("Hello!")

# Full audio streaming via session API
async with model.create_session() as session:
    await session.send_audio_chunk(audio_bytes)
    async for event in session.receive_events():
        handle(event)
```

### Changes

- **`create_aws_bedrock_runtime_client`**: New utility in `utils.py` with same credential validation as `create_aws_client` (explicit keys, profile, api_key, env var fallback, warning on conflict)
- **AWS credential fields**: `credentials_profile_name`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token`, `endpoint_url`, `bedrock_api_key` — matching `ChatBedrockConverse`
- **Voice validation**: Validates `voice_id` against known v1/v2 voices. V2-only voices (`olivia`, `kiara`, `arjun`, `tina`, `carolina`, `leo`) rejected when using v1 model
- **Endpointing sensitivity**: `endpointing_sensitivity` field (`HIGH`/`MEDIUM`/`LOW`) for turn detection — Nova 2 Sonic only, raises `ValueError` on v1
- **BaseChatModel alignment**: `_get_ls_params`, `_identifying_params`, `is_lc_serializable`, `get_lc_namespace`, `lc_secrets`, `ConfigDict(populate_by_name=True)`
- **Sync `_generate`**: Properly handles running/non-running event loops instead of deprecated `get_event_loop().run_until_complete()`

### Tests

- Voice validation: v1 voices on v1, v2 voices on v2, v2-on-v1 rejection, unknown voice rejection
- Endpointing sensitivity: v2-only enforcement, all valid values
- LangChain interface: `_get_ls_params`, `_identifying_params`, `is_lc_serializable`, `get_lc_namespace`, `lc_secrets`
- Model version detection: `_is_v2_model` helper
- Session protocol: event ordering, system prompt, text/audio sending, end sequence
- `create_aws_bedrock_runtime_client`: default, explicit keys, session token, profile, invalid creds, custom endpoint, env var region, api_key, api_key+creds warning, session region fallback


libs/aws/langchain_aws/chat_models/bedrock_nova_sonic.py
```diff
@@ -0,0 +1,990 @@
+"""Amazon Nova Sonic bidirectional streaming chat model.
```
michaelnchin: Remove this module-level docstring. The install instructions and quick start example can be moved to the `ChatBedrockNovaSonic` class docstring intesad.

libs/aws/langchain_aws/chat_models/bedrock_nova_sonic.py
```diff
@@ -0,0 +1,990 @@
...
+persistent, full-duplex streaming connection.
+
+!!! warning "Experimental"
+    This integration is experimental and requires the separate
+    ``aws-sdk-bedrock-runtime`` Python package, which is still under
+    active development.
+
+Installation::
+
+    pip install "langchain-aws[nova-sonic]"
```
michaelnchin: To get this working, we need to add a `nova-sonic` optional dependencies group in pyproject.toml: https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/pyproject.toml#L25

Something like:
```python
nova-sonic = ["aws-sdk-bedrock-runtime"]
```

libs/aws/langchain_aws/chat_models/bedrock_nova_sonic.py
michaelnchin, line 1: General comment - we should export ChatBedrockNovaSonic from `__init__.py` so users can import it via `langchain_aws import ChatBedrockNovaSonic`.

Please update both of these init files:
- https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/__init__.py
- https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/__init__.py

libs/aws/langchain_aws/chat_models/bedrock_nova_sonic.py
```diff
@@ -0,0 +1,990 @@
...
+        self._output_sample_rate = output_sample_rate
+        self._audio_media_type = audio_media_type
+        self._endpointing_sensitivity = endpointing_sensitivity
+
+        self._stream: Any = None
+        self._is_active = False
+        self._prompt_name = str(uuid.uuid4())
+        self._content_name = str(uuid.uuid4())
+        self._audio_content_name = str(uuid.uuid4())
+        self._response_task: Optional[asyncio.Task[None]] = None
```
michaelnchin: Nit: `_response_task` attribute is unused after assignment - remove this?

libs/aws/langchain_aws/chat_models/bedrock_nova_sonic.py
```diff
@@ -0,0 +1,990 @@
...
+        **kwargs: Any,
+    ) -> ChatResult:
+        """Generate a response from text messages via Nova Sonic.
+
+        Extracts system and user messages, sends them through a
+        bidirectional session, and collects the text response.
+
+        Args:
+            messages: Input messages.
+            stop: Stop sequences (not supported by Nova Sonic).
```
michaelnchin, line 924: If `stop` isn't supported by Sonic models, should we drop this parameter from the function signature? Also may apply to `_generate`

libs/aws/langchain_aws/chat_models/bedrock_nova_sonic.py
```diff
@@ -0,0 +1,990 @@
...
+            A ChatResult containing the model response.
+        """
+        system_prompt = self.system_prompt
+        user_texts: List[str] = []
+
+        for msg in messages:
+            if isinstance(msg, SystemMessage):
+                system_prompt = str(msg.content)
+            elif isinstance(msg, HumanMessage):
+                user_texts.append(str(msg.content))
```
michaelnchin, line 943: Optional, but may want to add a note in the docstring about why we ignore `AIMessage`s here (for users who may expect the standard multi-turn pattern for chat models)

```
Note:                                                                                                                                                                                                            
      Only ``SystemMessage`` and ``HumanMessage`` are processed from input.
      ``AIMessage`` and other message types are ignored because Nova                                                                                                                                               
      Sonic maintains conversation state within the bidirectional 
      stream. Previous assistant responses do not need to be sent
      back as context.
```

libs/aws/langchain_aws/chat_models/bedrock_nova_sonic.py
Keyur-S-Patel: We can't drop `stop` from the signature. [BaseChatModel._generate](https://github.com/langchain-ai/langchain/blob/4a632cf6a97487df4b79c3119e91e154dc987a4c/libs/core/langchain_core/language_models/chat_models.py#L1394) is an `@abstractmethod` that includes stop: list[str] | None = None, so all subclasses must accept it. The same applies to _agenerate and _astream, which have default implementations that forward stop to the subclass override.

michaelnchin: Let's set a logger warning to notify the user that `stop` has no effect for the supported models, if they attempt to set a value for it.

michaelnchin:
New integ tests passing in CI: https://github.com/langchain-ai/langchain-aws/actions/runs/23169256064

PR #919, opened 2026-03-11
feat(aws): add configurable code interpreter identifier to CodeInterpreter

harrytran001:
## Disclaimer

This contribution was developed with assistance from AI coding tools. The changes have been reviewed and tested locally (format, lint, unit tests).

---

## Summary

Adds an optional **code interpreter identifier** to `CodeInterpreterToolkit` and `create_code_interpreter_toolkit`, so callers can use a custom Bedrock code interpreter (e.g. one created via `create_code_interpreter` with VPC or other settings) instead of the default system interpreter (`aws.codeinterpreter.v1`).

Previously there was no way to configure the identifier; the toolkit always called `CodeInterpreter.start()` with no arguments.

---

## Why

- **Use case:** Users who provision custom code interpreters (e.g. for VPC, compliance, or isolation) need to pass that interpreter’s ID when starting sessions. The bedrock_agentcore SDK’s `CodeInterpreter.start(identifier=...)` already supports this; the toolkit did not expose it.
- **Approach:** Optional, keyword-only `code_interpreter_identifier` on both the toolkit and the factory. When `None` (default), behavior is unchanged: `start()` is called with no args and the SDK uses its default. When set, we call `start(identifier=code_interpreter_identifier)`.

---

## Changes

- **`CodeInterpreterToolkit`:** New `__init__(..., *, code_interpreter_identifier: Optional[str] = None)`. Stored and passed to `CodeInterpreter.start(identifier=...)` in `_get_or_create_interpreter` when creating a new session.
- **`create_code_interpreter_toolkit`:** New optional kwarg `code_interpreter_identifier`, passed through to the toolkit.
- **Tests:** New unit tests for init with identifier, `start(identifier=...)` when set, and factory passthrough. Existing tests unchanged (default behavior still asserts `start()` called with no args).

---

## Review focus

- **Backward compatibility:** New parameter is keyword-only and optional; existing callers are unaffected.
- **Thread vs identifier:** Identifier is per-toolkit (same for all threads); session isolation remains per `thread_id` via `_code_interpreters` dict. Confirm this matches expectations.
- **API surface:** Only identifier is added; no `session_timeout_seconds` or session `name` in this PR to keep the change minimal.

### DIFF
--- a/libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
+++ b/libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
@@ -122,7 +122,7 @@ class CodeInterpreterToolkit:
     The toolkit lazily initializes the code interpreter session on first use.
     It supports multiple threads by maintaining separate code interpreter sessions for each thread ID.
 
-        Example:
+    Example:
         ```python
         import asyncio
         from langchain.agents import create_agent


libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
```diff
@@ -122,16 +122,22 @@ class CodeInterpreterToolkit:
     The toolkit lazily initializes the code interpreter session on first use.
     It supports multiple threads by maintaining separate code interpreter sessions for each thread ID.
 
-    Example:
+        Example:
```
michaelnchin: Nit: retain the original formatting here

```suggestion
    Example:
```

PR #920, opened 2026-03-11
fix(aws): fix race condition when there are parallel tool calls

harrytran001:
## Disclaimer

This contribution was developed with assistance from AI coding tools. The changes have been reviewed and tested locally (format, lint, unit tests).

---

## Summary

Fixes a race in `CodeInterpreterToolkit`: when the first AI call triggers multiple code interpreter tools **concurrently** for the same `thread_id`, each tool called `_get_or_create_interpreter`, saw no interpreter yet, and created a new one. All wrote into `_code_interpreters[thread_id]`, so only the last instance was stored and the rest were orphaned (multiple sessions for one thread, possible resource leaks and inconsistent state).

This PR serializes get-or-create and cleanup with a **threading lock** so only one interpreter is created per `thread_id` and all concurrent callers receive the same instance.

---

## Why

- **Bug:** With parallel tool invocations (e.g. agent runs `execute_code`, `read_files`, `list_files` at once), the check-then-create in `_get_or_create_interpreter` was not atomic. Several interpreters were created and overwrote each other in `_code_interpreters[thread_id]`.
- **Approach:** Use a single `threading.Lock()` for the toolkit. In `_get_or_create_interpreter`, hold the lock for the full check → create → start → store sequence so only one creation happens per `thread_id`. In `cleanup()`, hold the lock only while mutating the map; call `stop()` outside the lock to avoid holding it during I/O.

---

## Changes

- **`CodeInterpreterToolkit`**
  - `import threading`; in `__init__`, add `self._interpreter_lock = threading.Lock()`.
  - `_get_or_create_interpreter`: wrap the entire "if thread_id in map → return; else create, start, store, return" block in `with self._interpreter_lock`.
  - `cleanup()`: under the lock, pop the interpreter(s) from `_code_interpreters` (or copy and clear); then call `interpreter.stop()` outside the lock.
- **Tests**
  - New unit test `test_concurrent_get_or_create_same_thread_returns_single_interpreter`: 3 threads call `_get_or_create_interpreter` for the same `thread_id` with a slow `start()` mock; assert `CodeInterpreter` is constructed once and all 3 get the same instance.

---

## Review focus

- **Lock scope:** Creation and start run under the lock (so first concurrent call may block briefly). Cleanup only holds the lock while updating the map; stop runs outside to avoid lock during I/O.
- **Compatibility:** Sync tool methods only; no API changes. Existing single-threaded and sequential usage is unchanged.

### DIFF
--- a/libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
+++ b/libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
@@ -130,8 +130,14 @@ class CodeInterpreterToolkit:
         from langchain_aws.tools import create_code_interpreter_toolkit
 
         async def main():
-            # Create and setup the code interpreter toolkit
+            # Create and setup the code interpreter toolkit (default interpreter)
             toolkit, code_tools = await create_code_interpreter_toolkit(region="us-west-2")
+
+            # Or use a custom code interpreter (e.g. VPC-configured)
+            # toolkit, code_tools = await create_code_interpreter_toolkit(
+            #     region="us-west-2",
+            #     code_interpreter_identifier="my-interpreter-abc123",
+            # )
 
             # Create a ReAct agent using the code interpreter tools
             agent = create_agent(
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/tools/code_interpreter_toolkit.py
```diff
@@ -681,17 +687,14 @@ async def cleanup(self, thread_id: Optional[str] = None) -> None:
...
                 try:
-                    self._code_interpreters[tid].stop()
+                    interpreter.stop()
                 except Exception as e:
-                    logger.warning(
-                        f"Error stopping code interpreter for thread {tid}: {e}"
-                    )
-
-            self._code_interpreters = {}
+                    logger.warning(f"Error stopping code interpreter: {e}")
```
michaelnchin: The new version drops thread id logging in the exception, IMO we should preserve this context:
```suggestion
            with self._interpreter_lock:
                interpreters = list(self._code_interpreters.items())
                self._code_interpreters = {}
            for interpreter in interpreters:
                try:
                    interpreter.stop()
                except Exception as e:
                    logger.warning(f"Error stopping code interpreter: {tid}: {e}")
```

harrytran001: Hi @michaelnchin, I made the change to preserve the thread_id in error logging

michaelnchin (approved):
Thanks @harrytran001 !

Issue #933, opened 2026-03-17
Add `response_format` support to ChatBedrockConverse for structured output via ProviderStrategy


michaelnchin:
> This is probably related to this ticket, but maybe not 100%. I am currently using 1.4.0 and am using the`init_chat_model` and `create_agent` factories. I am getting an error when trying to pass the `response_format` parameter in to `create_agent`, listed below the code below. I looked, and sure enough, `_converse_params` does not support responseFormat. Will this be supported at some point since 1.4.0 supports structured output, is this something that I can work around by using `with_structured_output` on the response from `init_chat_model`?
> 
> ```
> model = init_chat_model(
>    provider="bedrock_converse",
>    model="openai.gpt-oss-120b-1:0",
>    region_name = "us-east-1",
>    ...
> )
> agent = create_agent(
>    model=model,
>    tools=[],
>    system_prompt=MY_PROMPT,
>    response_format=MyPydanticModel,
> )
> ```
> 
> ```
> Error: Failed to extract: ChatBedrockConverse._converse_params() got an unexpected keyword argument 'responseFormat'
> ```  

 _Originally posted by @bryanpizzillo in [#571](https://github.com/langchain-ai/langchain-aws/issues/571#issuecomment-4064856804)_

bryanpizzillo:
Hi @michaelnchin,
Just following up here from the comment in https://github.com/langchain-ai/langchain-aws/issues/571#issuecomment-4072153914 that there does not seem to be a way to **just** *"Set [output_config](https://reference.langchain.com/python/langchain-aws/chat_models/bedrock_converse/ChatBedrockConverse/output_config) directly in the ChatBedrockConverse model instantiation before passing it to create_agent."* 

What I am finding is that langchain is not expecting to get the parsed object back without `response_format`. Additionally, the parameter name from the return is not what is expected. The best I have figured to work around it is to create Middleware and look at the last AIMessage and get the raw json back in order to populate my model. But this is all very messy.

I don't know if you had found that while implementing the PR or not. Additionally if you have any suggestions, or something I missed around create_agent that would be appreciated.

michaelnchin:
@bryanpizzillo pretty much yeah, the workaround can't directly solve the underlying issue with `response_format` in the latest release, since that's hard blocked by the arg always getting passed to `_converse_params`. 

[...]

PR #934, opened 2026-03-17
feat(aws): Update ChatBedrockConverse to support `create_agent` `response_format`

michaelnchin:
Fixes: #933 / https://github.com/langchain-ai/langchain-aws/issues/571#issuecomment-4064856804

Updates ChatBedrockConverse to be able to handle `response_format` JSON schema input passed by `create_agent`. The schema will be converted to the format expected by [`output_config`](https://reference.langchain.com/python/langchain-aws/chat_models/bedrock_converse/ChatBedrockConverse/output_config) before being passed to Converse API.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -1133,13 +1133,7 @@ class ChatBedrockConverse(BaseChatModel):
         # Remove disable_streaming from kwargs as it's not a valid API parameter
         filtered_kwargs = {k: v for k, v in kwargs.items() if k != "disable_streaming"}
         additional_fields = filtered_kwargs.pop("additional_model_request_fields", None)
-
-        response_format = filtered_kwargs.pop("response_format", None)
-        if response_format and "output_config" not in filtered_kwargs:
-            filtered_kwargs["output_config"] = _response_format_to_output_config(
-                response_format
-            )
-
+        _apply_response_format(filtered_kwargs)
         cache_control = filtered_kwargs.pop("cache_control", None)
         params = self._converse_params(
             stop=stop,
@@ -1201,13 +1195,7 @@ class ChatBedrockConverse(BaseChatModel):
         # Remove disable_streaming from kwargs as it's not a valid API parameter
         filtered_kwargs = {k: v for k, v in kwargs.items() if k != "disable_streaming"}
         additional_fields = filtered_kwargs.pop("additional_model_request_fields", None)
-
-        response_format = filtered_kwargs.pop("response_format", None)
-        if response_format and "output_config" not in filtered_kwargs:
-            filtered_kwargs["output_config"] = _response_format_to_output_config(
-                response_format
-            )
-
+        _apply_response_format(filtered_kwargs)
         cache_control = filtered_kwargs.pop("cache_control", None)
         params = self._converse_params(
             stop=stop,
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -1194,6 +1201,13 @@ def _stream(
         # Remove disable_streaming from kwargs as it's not a valid API parameter
         filtered_kwargs = {k: v for k, v in kwargs.items() if k != "disable_streaming"}
         additional_fields = filtered_kwargs.pop("additional_model_request_fields", None)
+
+        response_format = filtered_kwargs.pop("response_format", None)
+        if response_format and "output_config" not in filtered_kwargs:
+            filtered_kwargs["output_config"] = _response_format_to_output_config(
+                response_format
+            )
```
3coins: **minor**:  This exact block is duplicated above. Moving to a helper would avoid that.

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
3coins, line 1: **minor**: Tests are not covering the `_stream` path.

3coins (commented):
@michaelnchin 
Thanks for making these updates. We should merge this change with some caveats.

My understanding is that `create_agent` won't actually trigger this code path for most Bedrock models. For Claude models that Bedrock now supports for structured output (Claude Haiku 4.5, Sonnet 4.5, Opus 4.5, Opus 4.6 per the docs), the _PROFILES data in [_profiles.py](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/data/_profiles.py) should be updated to include `"structured_output": True`. Without that, `create_agent` will never select ProviderStrategy for Claude, and this code path won't be exercised for those models. This could be a follow-up PR, but it's worth checking this theory out.

libs/aws/langchain_aws/chat_models/bedrock_converse.py
michaelnchin: Done, moved to `_apply_response_format` helper

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
michaelnchin: Added tests for streaming.

michaelnchin:
@3coins thanks, updated for the feedback 👍 

Good catch on the missing structured output support indicators in `_profiles.py`. 

I initially considered adding the overrides directly into our [profile_augmentations.toml](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/data/profile_augmentations.toml) (ex. like [in langchain-anthropic](https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/langchain_anthropic/data/profile_augmentations.toml#L11)) file, but I think there would be maintainability concerns to be had given the amount of models we would have to update per: https://docs.aws.amazon.com/bedrock/latest/userguide/structured-output.html

For a more permanent fix, filed a PR to fix the Bedrock provider model files in the upstream models.dev repo: https://github.com/anomalyco/models.dev/pull/1326. 

Once merged, this will get pulled automatically into our `_profiles.py` files via the [`refresh_model_profiles`](https://github.com/langchain-ai/langchain-aws/blob/main/.github/workflows/refresh_model_profiles.yml) weekly cron job.

3coins (approved):

PR #932, opened 2026-03-16
fix(aws): make NovaSystemTool extend BaseTool for create_react_agent compatibility

AlonNaor22:
## Summary

`NovaSystemTool` was a plain Python class, so `ToolNode` and `create_react_agent` crashed with `ValueError` when passed tools like `NovaGroundingTool()`. This PR makes `NovaSystemTool` extend `BaseTool` to fix that.

The fix works because Nova system tools execute server-side — their results come back as `server_tool_use` blocks that never appear in `AIMessage.tool_calls`, so `ToolNode` never tries to run them locally. It just needs to accept them during init.

Changes:
- `NovaSystemTool` now extends `BaseTool` with a no-op `_run()`, empty `args_schema`, and `description` fields
- `bind_tools()` still routes them correctly via the existing `isinstance(tool, NovaSystemTool)` check
- Added `TestToolNodeCompatibility` tests that reproduce the original crash

Backward compatible — `NovaGroundingTool()` and `NovaCodeInterpreterTool()` constructors unchanged, `to_bedrock_format()` output unchanged, all 660 existing unit tests pass.

## Test plan

- All 21 nova tool tests pass (6 new)
- All 27 bind_tools tests pass
- Full suite: 660 passed, 0 failures

## AI disclosure

This contribution was developed with the assistance of Claude (AI), with human review and validation of the approach, code, and test results.

Closes #921

Issue #921, opened 2026-03-12
NovaGroundingTool raises Exception when using `create_agent`


jacob-swe-sp:
### Package

langchain-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code

```python
# From the docs and works as described: https://reference.langchain.com/python/langchain-aws/tools/nova_tools

from langchain_aws import ChatBedrockConverse
from langchain_aws.tools import NovaGroundingTool

model = ChatBedrockConverse(model="us.amazon.nova-premier-v1:0")
model_with_search = model.bind_tools([NovaGroundingTool()])
response = model_with_search.invoke(
    "What are the latest developments in quantum computing?"
)

# A slight alteration that raises an exception:

from langchain_aws import ChatBedrockConverse
from langchain_aws.tools import NovaGroundingTool
from langchain.agents import create_agent

model = ChatBedrockConverse(model="us.amazon.nova-premier-v1:0")
agent = create_agent(
    model=model,
    tools=[NovaGroundingTool()],
)
response = agent.invoke(
    "What are the latest developments in quantum computing?"
)

# Another slight alteration that also does not work

from langchain_aws import ChatBedrockConverse
from langchain_aws.tools import NovaGroundingTool
from langchain.agents import create_agent

model = ChatBedrockConverse(model="us.amazon.nova-premier-v1:0")
model_with_tools = model.bind_tools([NovaGroundingTool()])
agent = create_agent(
    model=model_with_tools,
)
response = agent.invoke(
    "What are the latest developments in quantum computing?"
)
```

### Error Message and Stack Trace (if applicable)

```shell
For the first one above:
Traceback (most recent call last):
  File "...", line 6, in <module>
    agent = create_agent(
        model=model,
        tools=[NovaGroundingTool()],  # ty:ignore[invalid-argument-type]
    )
  File "/.../.venv/lib/python3.13/site-packages/langchain/agents/factory.py", line 921, in create_agent
    ToolNode(
    ~~~~~~~~^
        tools=available_tools,
        ^^^^^^^^^^^^^^^^^^^^^^
        wrap_tool_call=wrap_tool_call_wrapper,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #956, opened 2026-03-27
fix(aws): Edge case fix for NovaSystemTool and auto/any tool calling

jacob-swe-sp:
Hello kind LangChain AWS folks.

In continuing to develop on top of Nova and the system tools, I have found one more edge case that seems to be a bug that should be patched. This work builds on top of the issue in https://github.com/langchain-ai/langchain-aws/issues/921 that was partially solved in https://github.com/langchain-ai/langchain-aws/pull/945 and https://github.com/langchain-ai/langchain-aws/pull/932.

The issue this PR solves is that when you use a Nova model with Bedrock Converse and bind both Bedrock system tools (for example Nova grounding, which is my use case) and normal custom tools (Pydantic / function tools, in my instance a structured output via the `response_format` field), the request often ends up with toolChoice set to “any”, which is incompatible with Bedrock Converse (the feeling of this is that this behavior may benefit from being loosened? but definitely out of scope for this patch). Anyways, this adds a small guard that only applies when: the base model id looks like Nova;
toolConfig includes at least one systemTool and at least one toolSpec; and toolChoice is exactly {"any": {}}. At least for my specific setup, this patches out the issue.

Thank you for your efforts!


jacob-swe-sp:
That works for me, I suspect the wonkiness is arising out of the `response_format` field in `create_agent` (which is where I got stuck on this). Let me try to write up a minimal reproduction of the error and see if I can get the logic moved over to the binding step (agree that makes more sense than per-invocation). Thanks for the rapid review.

jacob-swe-sp:
Alright, I have a repro that seems to accurately capture the issue, and your suggestion to move things around was massively simplifying. Thank you for the guidance!

This code fails for me under the current, latest release of the package:
```python
from typing import Any

from langchain.agents import create_agent
from langchain_aws.tools import NovaGroundingTool
from langgraph.graph.state import CompiledStateGraph
from pydantic import BaseModel, Field


class GroundedAnswer(BaseModel):
    answer: str = Field(description="Short answer to the user.")
    sources: list[str] = Field(default_factory=list, description="Citations.")


def main() -> None:
    model_id: str = "us.amazon.nova-2-lite-v1:0"
    region_name: str = "us-east-1"
    llm = ChatBedrockConverse(model=model_id, region_name=region_name)
    agent: CompiledStateGraph = create_agent(
        model=llm, tools=[NovaGroundingTool()], response_format=GroundedAnswer
    )
    result: Any = agent.invoke(
        input={"messages": [HumanMessage("Summarize the latest AWS announcements.")]},
    )
    print(result)


if __name__ == "__main__":
    main()
```

Truncated error trace that occurs:
```
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid number of parameters set for tagged union structure toolConfig.tools[0]. Can only set one of the following keys: toolSpec, systemTool, cachePoint.
Unknown parameter in toolConfig.tools[0]: "type", must be one of: toolSpec, systemTool, cachePoint
Unknown parameter in toolConfig.tools[0]: "function", must be one of: toolSpec, systemTool, cachePoint
```

When running with the patch I've updated into the PR (more, copy-pasted your code lol) things work smoothly. I was also able to more fully validate the change in the codebase utilizing this functionality and can confirm this would resolve the issues I am encountering.

PR #966, opened 2026-04-02
feat(aws): isolate browser and ci sessions for parallel subagents

sundargthb:
## Summary

Fixes session collisions when parallel LangGraph subgraphs (e.g., Deep Agents
subagents) share a single `BrowserToolkit` or `CodeInterpreterToolkit` instance.

## Problem

Both `BrowserSessionManager` and `CodeInterpreterToolkit` key sessions by
`thread_id` alone. When LangGraph runs parallel subagents, each subagent
inherits the parent's `thread_id` but receives a unique `checkpoint_ns`:

```python
config["configurable"] = {
    "thread_id": "parent-thread-001",         # SAME for all subagents
    "checkpoint_ns": "research-acme:abc123",   # UNIQUE per subagent
}
```

This causes two issues:
1. **BrowserSessionManager** raises `RuntimeError("session already in use")`
   when a second parallel subagent tries to acquire the same session
2. **CodeInterpreterToolkit** silently shares one interpreter across
   parallel subagents, causing state contamination

Additionally, when a single LLM turn emits multiple browser tool calls
(e.g., `navigate_browser` + `extract_text`), LangGraph runs them concurrently
via `asyncio.gather`. The session manager's "in use" flag causes the second
tool call to crash.

## Solution

1. **`utils.get_session_key()`** — 
   session keys: `"{thread_id}:{checkpoint_ns}"` when `checkpoint_ns` is present,
   plain `thread_id` otherwise. Both `browser_tools.get_thread_id()` and
   `code_interpreter_toolkit._get_thread_id()` now delegate to this function.

2. **`BrowserSessionManager.get_async_browser()`** — now waits (with 10s timeout)
   instead of raising `RuntimeError` when a session is temporarily in use. This
   handles the concurrent tool call case gracefully.

3. **`BrowserToolkit.get_active_sessions()`** — new introspection method that
   returns session metadata (session_id, identifier) for all active sessions.
   Useful for debugging and UI display.

## Backward Compatibility

Backward compatible — top-level agents without checkpoint_ns continue to use thread_id as the session key, producing identical behavior to the previous implementation

- Top-level agents without `checkpoint_ns` continue to use `thread_id` as the
  session key — no behavior change
- `get_thread_id()` method name preserved to avoid breaking internal callers
- Sync `get_sync_browser()` unchanged (raises immediately, as before)

## Testing

```python
from langchain_aws.tools.utils import get_session_key

# Top-level agent
config = {"configurable": {"thread_id": "t1"}}
assert get_session_key(config) == "t1"

# Parallel subagent with checkpoint_ns
config = {"configurable": {"thread_id": "t1", "checkpoint_ns": "sub:abc"}}
assert get_session_key(config) == "t1:sub:abc"

# Different subagents get different keys
c1 = {"configurable": {"thread_id": "t1", "checkpoint_ns": "sub-a:1"}}
c2 = {"configurable": {"thread_id": "t1", "checkpoint_ns": "sub-b:2"}}
assert get_session_key(c1) != get_session_key(c2)

# No config
assert get_session_key(None) == "default"
```

## Files Changed

| File | Change |
|---|---|
| `libs/aws/langchain_aws/tools/utils.py` |  — `get_session_key()` utility |
| `libs/aws/langchain_aws/tools/browser_tools.py` | Import + delegate `get_thread_id()` |
| `libs/aws/langchain_aws/tools/code_interpreter_toolkit.py` | Import + delegate `_get_thread_id()` |
| `libs/aws/langchain_aws/tools/browser_session_manager.py` | Import asyncio + wait-with-timeout in `get_async_browser()` |
| `libs/aws/langchain_aws/tools/browser_toolkit.py` | Add `get_active_sessions()` method |

## Context

This change enables the LangChain Deep Agents + AWS Bedrock AgentCore
integration pattern where a single shared toolkit serves multiple parallel
subagents, each getting its own isolated AgentCore MicroVM session automatically.

### DIFF
--- a/libs/aws/langchain_aws/tools/browser_session_manager.py
+++ b/libs/aws/langchain_aws/tools/browser_session_manager.py
@@ -42,6 +42,7 @@ class BrowserSessionManager:
         profile_configuration: Optional[
             Union[ProfileConfiguration, Dict[str, Any]]
         ] = None,
+        session_wait_timeout: float = 10.0,
     ):
         """
         Initialize the browser session manager.
@@ -56,6 +57,7 @@ class BrowserSessionManager:
         self.proxy_configuration = proxy_configuration
         self.extensions = extensions
         self.profile_configuration = profile_configuration
+        self.session_wait_timeout = session_wait_timeout
         self._async_sessions: Dict[str, Tuple[BrowserClient, AsyncBrowser, bool]] = {}
         self._sync_sessions: Dict[str, Tuple[BrowserClient, SyncBrowser, bool]] = {}
 
@@ -99,7 +101,8 @@ class BrowserSessionManager:
                 # This happens when LangGraph runs multiple browser tool
                 # calls concurrently (e.g., navigate + extract_text in
                 # the same LLM turn via asyncio.gather).
-                for _ in range(100):  # 10 second timeout
+                max_attempts = int(self.session_wait_timeout / 0.1)
+                for _ in range(max_attempts):  # 10 second timeout
                     await asyncio.sleep(0.1)
                     _, _, still_in_use = self._async_sessions.get(
                         thread_id, (None, None, False)
@@ -109,7 +112,8 @@ class BrowserSessionManager:
                 else:
                     raise RuntimeError(
                         f"Browser session for thread {thread_id} timed out "
-                        "waiting for previous operation to complete (10s). "
+                        "waiting for previous operation to complete "
+                        f"({self.session_wait_timeout}s). "
                         "Use a different thread_id for truly concurrent operations."
                     )
                 client, browser, _ = self._async_sessions[thread_id]
--- a/libs/aws/langchain_aws/tools/browser_toolkit.py
+++ b/libs/aws/langchain_aws/tools/browser_toolkit.py
@@ -200,26 +200,16 @@ class BrowserToolkit:
 
         """
         sessions: Dict[str, Dict[str, str]] = {}
-        for key, (
-            client,
-            browser,
-            in_use,
-        ) in self.session_manager._async_sessions.items():
-            if client.session_id:
-                sessions[key] = {
-                    "session_id": client.session_id,
-                    "identifier": client.identifier or "",
-                }
-        for key, (
-            sync_client,
-            sync_browser,
-            sync_in_use,
-        ) in self.session_manager._sync_sessions.items():
-            if sync_client.session_id:
-                sessions[key] = {
-                    "session_id": sync_client.session_id,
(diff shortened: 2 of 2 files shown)


libs/aws/langchain_aws/tools/browser_toolkit.py
```diff
@@ -181,6 +181,47 @@ def get_tools_by_name(self) -> Dict[str, BaseTool]:
...
+        for key, (
+            sync_client,
+            sync_browser,
+            sync_in_use,
+        ) in self.session_manager._sync_sessions.items():
+            if sync_client.session_id:
+                sessions[key] = {
+                    "session_id": sync_client.session_id,
+                    "identifier": sync_client.identifier or "",
+                }
```
michaelnchin: Nit - can simplify these using one outer loop:
```python
for store in (
    self.session_manager._async_sessions,
    self.session_manager._sync_sessions,
):
    for key, (client, browser, in_use) in store.items():
        if client.session_id:
            sessions[key] = {
                "session_id": client.session_id,
                "identifier": client.identifier or "",
            }
```

libs/aws/langchain_aws/tools/browser_session_manager.py
```diff
@@ -94,10 +95,24 @@ async def get_async_browser(self, thread_id: str) -> AsyncBrowser:
...
             if in_use:
-                raise RuntimeError(
-                    f"Browser session for thread {thread_id} is already in use. "
-                    "Use a different thread_id for concurrent operations."
-                )
+                # Wait for the current operation to release the session.
+                # This happens when LangGraph runs multiple browser tool
+                # calls concurrently (e.g., navigate + extract_text in
+                # the same LLM turn via asyncio.gather).
+                for _ in range(100):  # 10 second timeout
```
michaelnchin: Do we want to make this timeout value configurable?

michaelnchin (commented):
@sundargthb thanks for finding this and making this fix, looks great overall!

libs/aws/langchain_aws/tools/browser_toolkit.py
sundargthb: thanks, good call. updated

libs/aws/langchain_aws/tools/browser_session_manager.py
sundargthb: Good idea — made it configurable via a session_wait_timeout parameter on `BrowserSessionManager.__init__()`, defaulting to 10 seconds. This way you can tune it without needing to subclass, but the default just works for the common case.

PR #990, opened 2026-04-10
feat(langgraph-checkpoint-aws): add S3 key prefix support to S3OffloadConfigDict #986

Keyur-S-Patel:
Closes #986

### Summary

Adds an optional `key_prefix` field to `S3OffloadConfigDict` for the DynamoDB checkpointer's S3 offloading, allowing users to share a single S3 bucket across multiple applications with prefix-based isolation.

### Usage

```python
from langgraph_checkpoint_aws import DynamoDBSaver

saver = DynamoDBSaver(
    table_name="my-checkpoints",
    s3_offload_config={
        "bucket_name": "shared-data-bucket",
        "key_prefix": "my-app/langgraph",
    },
)
```

Without `key_prefix`, S3 keys are generated as:
```
{thread_id}/checkpoints/{checkpoint_ns}/{checkpoint_id}
```

With `key_prefix`, they become:
```
{key_prefix}/{thread_id}/checkpoints/{checkpoint_ns}/{checkpoint_id}
```

### Changes

- **`saver.py`**: Added `key_prefix` to `S3OffloadConfigDict`, extracted and passed through to `StorageStrategy`
- **`storage_strategy.py`**:
  - New `s3_key_prefix` parameter on `StorageStrategy.__init__` (leading/trailing slashes stripped automatically)
  - `_prefixed_s3_key()` helper prepends the prefix when configured
  - Applied in `_store_to_s3`, `_retrieve_from_s3`, and `_batch_delete_from_s3`
- **`__init__.py`**: No changes — `S3OffloadConfigDict` is a `TypedDict` with `total=False`, so the new field is fully backward compatible

### Tests

- **8 new unit tests** across `test_saver.py` and `test_storage_strategy.py`: init normalization (slash stripping, empty string), prefix application, store/retrieve/delete with prefix, saver passthrough with and without `key_prefix`

### Areas requiring careful review

- **Prefix applied at storage layer, not key generation**: The prefix is prepended in `StorageStrategy._store_to_s3` / `_retrieve_from_s3` / `_batch_delete_from_s3` rather than in the `_checkpoint_s3_key` / `_writes_s3_key` functions. This means the `ref_key` stored in DynamoDB remains unprefixed. This is intentional — it keeps the DynamoDB metadata portable and avoids breaking existing data if a user later changes or removes the prefix. Reviewers should verify this design choice aligns with the project's expectations.

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/saver.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/saver.py
@@ -32,13 +32,15 @@ logger = logging.getLogger(__name__)
 
 
 class S3OffloadConfigDict(TypedDict, total=False):
-    """Type alias for S3 offload configuration
+    """Type alias for S3 offload configuration.
+
     Required when user needs to configure Saver, to offload large
     Checkpoints above threshold (350KB) to S3.
     """
 
     bucket_name: str
     endpoint_url: str | None
+    key_prefix: str | None
 
 
 class DynamoDBSaver(BaseCheckpointSaver):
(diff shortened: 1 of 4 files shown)


michaelnchin (approved):
Thanks @Keyur-S-Patel !

Non-blocking note - `_ensure_s3_lifecycle_policy` still creates an LC rule for the whole bucket, this should scoped down to be created per-prefix (specifically because the TTL that is used to generate the ID may match between different apps). Will update in a separate PR

PR #994, opened 2026-04-12
fix(agentcore-codeinterpreter): add async support with dedicated executor and session expiry handling

### LINKED ISSUE #985: AgentCoreSandbox async methods hang when used with active async event loop
### Package

langchain-aws

### Checked other resources

### Example Code

```python
import asyncio

interpreter = CodeInterpreter(region="us-west-2")
interpreter.start()

async def main():
    # This hangs when other async HTTP activity (e.g. LLM calls) is
    # in the same event loop
    result = await backend.awrite("/test.txt", "hello world")
    print(result)

asyncio.run(main())
```

```shell
No stack trace available — the call hangs indefinitely rather than raising an exception. Reported by a user through the LangChain maintainers. The hang occurs on awrite and potentially other async methods (aexecute, aupload_files, adownload_files) when async HTTP is active in the same event loop (e.g., concurrent LLM calls via httpx).
```

### Description

AgentCoreSandbox only implements sync methods (execute, upload_files, download_files). All async variants (awrite, aexecute, etc.) fall through to BackendProtocol’s default implementation which uses asyncio.to_thread with the default executor.

Two issues that probably cause the hang:
	1.	invoke_code_interpreter returns a lazy EventStream. Iterating it in _extract_text_from_stream holds the HTTP connection open and blocks the thread for the entire duration of server-side execution (up to 5 minutes, per read_timeout=300). This can exhaust the default executor’s thread pool when other asyncio.to_thread calls (LLM requests, tool dispatch) are competing for the same pool.
	2.	Session expiry (ResourceNotFoundException) is silently swallowed by a broad except Exception handler in execute(), producing a generic error message instead of a clear “session expired” indication.

sundargthb:
Fixes #985

Addresses two issues reported by a user via the LangChain maintainers:

**Async hang**: `awrite` and other async methods would hang when async
   HTTP was active in the same event loop. Root cause: `invoke_code_interpreter`
   returns a lazy EventStream that holds the HTTP connection open while
   the thread blocks on iteration. Combined with `asyncio.to_thread` using
   the default executor, this could starve other async work.

   Fix: eagerly consume the stream immediately after `invoke()` returns,
   and override all async methods to use a dedicated `ThreadPoolExecutor`
   instead of the default executor.

**Session expiry swallowed**: `execute()` catches `Exception` broadly,
   which silently converts `ResourceNotFoundException` (session expired)
   into a generic error string.

   Fix: catch `ClientError` with `ResourceNotFoundException` specifically
   and surface a clear "session expired" error message.

**Changes:**
Added `_invoke()` helper that eagerly materializes EventStream and
  catches session expiry
Added `SessionExpiredError` exception class
Added async overrides for `aexecute`, `aread`, `awrite`, `aedit`,
  `aupload_files`, `adownload_files` using a dedicated 4-thread executor
Added specific `SessionExpiredError` handling in `execute`,
  `upload_files`, `download_files`

**Testing:**
- All 46 unit tests pass (`make test`), lint clean (`make lint`)
- Mock tests verify eager stream consumption converts lazy generators to lists
- Mock tests verify `SessionExpiredError` is raised for `ResourceNotFoundException` and surfaced clearly in `execute`, `upload_files`, `download_files`
- Async executor isolation verified: `invoke()` runs on `agentcore-sandbox` threads, not the default asyncio executor. This is confirmed by tracking `threading.current_thread().name` inside the mocked `invoke()` call — the thread name must contain `agentcore-sandbox`, not `MainThread`. Note: checking the thread after `await` returns will always show `MainThread` (that's the event loop thread); the correct approach is to instrument the `invoke()` call itself.
- Live smoke test confirmed: sync `execute` and async `aexecute` both return correct results against a real AgentCore session

AI disclaimer: AI agents assisted with code generation and review.

### DIFF
--- a/libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
+++ b/libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
@@ -2,14 +2,20 @@
 
 from __future__ import annotations
 
+import asyncio
 import base64
 import logging
+from concurrent.futures import ThreadPoolExecutor
 from typing import TYPE_CHECKING, Any
 
+from botocore.exceptions import ClientError
 from deepagents.backends.protocol import (
+    EditResult,
     ExecuteResponse,
     FileDownloadResponse,
     FileUploadResponse,
+    ReadResult,
+    WriteResult,
 )
 from deepagents.backends.sandbox import BaseSandbox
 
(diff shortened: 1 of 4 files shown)


michaelnchin (approved):
LGTM - thanks Sundar!

PR #995, opened 2026-04-12
docs: add competitive research agent with Deep Agents and AgentCore

sundargthb:
Adds a runnable Jupyter notebook demonstrating LangChain Deep Agents
orchestrating parallel subagents with Amazon Bedrock AgentCore tools.

The agent launches 3 browser-based researchers in parallel (each in
its own AgentCore Browser MicroVM), passes findings to an analyst
(AgentCore Interpreter), and optionally saves insights to AgentCore
Memory. LangSmith tracing is included for observability.

Files:
samples/agents/competitive-research-agent.ipynb


samples/agents/competitive_research_agent.ipynb
michaelnchin, line 1: Nit: use underscores in name for consistency: `competitive_research_agent.ipynb`

samples/agents/competitive_research_agent.ipynb
```diff
@@ -0,0 +1,620 @@
...
+   "cell_type": "markdown",
+   "id": "2011b783",
+   "metadata": {},
+   "source": [
+    "## Create browser toolkits\n",
+    "\n",
+    "Each competitor gets its own `BrowserToolkit`, which means its own AgentCore Browser MicroVM.\n",
+    "This provides complete isolation between parallel researchers.\n",
+    "\n",
+    "We wrap each toolkit's tools with an `asyncio.Lock` to handle the case where the LLM emits\n",
```
michaelnchin: I don't see the `asyncio.Lock` wrappers in the code? Should implement it or remove this part of the description

samples/agents/competitive_research_agent.ipynb
```diff
@@ -0,0 +1,620 @@
...
+    "# Required\n",
+    "os.environ[\"AWS_REGION\"] = \"us-west-2\"\n",
+    "\n",
+    "# Optional: AgentCore Memory (leave empty to skip)\n",
+    "os.environ[\"AGENTCORE_MEMORY_ID\"] = \"\"  # e.g., \"mem-abc123\"\n",
+    "\n",
+    "# Optional: LangSmith tracing\n",
+    "os.environ[\"LANGCHAIN_TRACING_V2\"] = \"true\"\n",
+    "os.environ[\"LANGCHAIN_API_KEY\"] = \"\"  # your LangSmith API key\n",
+    "os.environ[\"LANGCHAIN_PROJECT\"] = \"competitive-research-agent\"\n",
```
michaelnchin: I think we should comment all of this optional code out - I tried as is with `LANGCHAIN_TRACING_V2=true` + no `LANGCHAIN_API_KEY`, and got spammed with `LangSmithAuthError`s while the agent ran

samples/agents/competitive-research-agent.ipynb
```diff
@@ -0,0 +1,620 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a283c173",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Install dependencies\n",
+    "%pip install -q \"langchain-aws[tools]\" deepagents bedrock-agentcore langgraph-checkpoint-aws"
```
michaelnchin, line 51: Nit: `langgraph-checkpoint-aws` not used, remove
```suggestion
    "%pip install -q \"langchain-aws[tools]\" deepagents bedrock-agentcore"
```

samples/agents/competitive_research_agent.ipynb
```diff
@@ -0,0 +1,620 @@
...
+   "source": [
+    "## Prerequisites\n",
+    "\n",
+    "- AWS account with Amazon Bedrock AgentCore access enabled\n",
+    "- AWS credentials exported as environment variables:\n",
+    "  ```bash\n",
+    "  export AWS_ACCESS_KEY_ID=\"...\"\n",
+    "  export AWS_SECRET_ACCESS_KEY=\"...\"\n",
+    "  export AWS_SESSION_TOKEN=\"...\"  # if using temporary credentials\n",
+    "  export AWS_REGION=\"us-west-2\"\n",
```
michaelnchin, line 37: Nit: closing ``` missing for markdown code block
```suggestion
    "  export AWS_REGION=\"us-west-2\"\n",
    "  ```\n",
```

samples/agents/competitive_research_agent.ipynb
```diff
@@ -0,0 +1,620 @@
...
+    "\n",
+    "    @tool\n",
+    "    async def list_stored_memories(namespace_prefix: str = \"/\", max_results: int = 10) -> str:\n",
+    "        \"\"\"List stored long-term memory records.\"\"\"\n",
+    "        try:\n",
+    "            import boto3\n",
+    "            client = boto3.client(\"bedrock-agentcore\", region_name=AWS_REGION)\n",
+    "            response = client.list_memory_records(\n",
+    "                memoryId=MEMORY_ID, namespace=namespace_prefix, maxResults=max_results,\n",
+    "            )\n",
```
michaelnchin: Curious - you can also call `list_memory_records` [via MemoryClient](https://github.com/aws/bedrock-agentcore-sdk-python/blob/b8dc75135e63b765232f2aea576cc1372df91140/src/bedrock_agentcore/memory/client.py#L52), as done with `create_event` and `retrieve_memories` above - any specific reason for switching to boto3 client directly here?

michaelnchin (commented):
Looks great overall, thanks @sundargthb!

samples/agents/competitive_research_agent.ipynb
sundargthb: Good catch — we removed the lock wrapper when we switched to using `session_wait_timeout` instead. Updated the markdown to match the code.

samples/agents/competitive_research_agent.ipynb
sundargthb: commented out the optional block and added a note

samples/agents/competitive_research_agent.ipynb
sundargthb: No good reason — MemoryClient doesn’t seem to have a list_memory_records wrapper — it has retrieve_memories (semantic search) and list_events (raw events), but neither is equivalent to listing extracted memory records.

michaelnchin: `list_memory_records` isn't explicitly implemented like the other two, but MemoryClient does [forward it](https://github.com/aws/bedrock-agentcore-sdk-python/blob/b8dc75135e63b765232f2aea576cc1372df91140/src/bedrock_agentcore/memory/client.py#L108-L132) and most other native AgentCore API calls directly to the boto3 client(s) that it already creates. 

The `__getattr__` docstring lists it as an example also:
```python
# These calls are forwarded to the appropriate boto3 client
response = client.list_memory_records(memoryId="mem-123", namespace="test/")
```

So I think we can make the above cleaner by just using:
```python
response = memory_client.list_memory_records(
    memoryId=MEMORY_ID, namespace=namespace_prefix, maxResults=max_results,
)
```

sundargthb: You’re right — missed the __getattr__ forwarding. Updated to use memory_client.list_memory_records() directly.

michaelnchin (approved):
LGTM, thanks @sundargthb !

PR #1003, opened 2026-04-15
fix(aws): Ensure that models don't ignore tool results when `guard_last_turn_only=True` on ChatBedrockConverse

### LINKED ISSUE #996: guard_last_turn_only: EMPTY_CONTENT "." guardContent causes model to ignore tool results
## Bug

The fix in #954 (released in v1.4.2) for #900 appends a `guardContent` block with `EMPTY_CONTENT = "."` to tool-result messages when `guard_last_turn_only=True`. This prevents the full-conversation guardrail fallback scan (the original bug), but introduces a new regression: **the model ignores tool results on tool-continuation turns**.

## Root Cause

When a `ToolMessage` is the last user turn, `_apply_guard_last_turn_only` now appends:

```python
{"guardContent": {"text": {"text": "."}}}
```

The Bedrock message sent to the model becomes:

```json
{"role": "user", "content": [
    {"toolResult": {"content": [{"text": "...cancellation steps..."}], "toolUseId": "..."}},
    {"guardContent": {"text": {"text": "."}}}
]}
```

The model sees the `"."` as part of the user turn — effectively a user message containing just a period. It interprets this as a "nothing more to say" signal and responds with a generic sign-off instead of processing the tool result.

## Reproduction

1. Create a `ChatBedrockConverse` with `guardrail_config` and `guard_last_turn_only=True`
2. Use it in a ReAct agent with tools (e.g., via LangGraph or deepagents)
3. Send a message that triggers a tool call (e.g., "how do I cancel Amazon")
4. Tool returns a substantive result (e.g., numbered cancellation steps)
5. **Expected:** Model presents the tool result to the user
6. **Actual:** Model ignores the tool result and responds with a generic message like "I'm here if you need help with anything else"

This is **100% reproducible** — every tool call has its result ignored on the first attempt.

## Trace Evidence

From our LangSmith traces:

- **First LLM call** (user message → tool call): Works correctly, model calls the tool
- **Second LLM call** (tool result → response): Model receives the tool result but ignores it
  - `guardrailCoverage.textCharacters.guarded: 1` (the single `"."` character)
  - `stopReason: "end_turn"` (not blocked — the guardrail passes, but the model just doesn't use the tool result)

When we **pin langchain-aws to 1.4.1** (before the #954 fix), tool results are processed correctly again.

When we replay the exact same messages in the **LangSmith playground** (which bypasses `ChatBedrockConverse` and the guardrail wrapping), the model processes tool results correctly.

## Suggested Fix

Use an empty string `""` instead of `"."` for the no-op guardContent block, or use a whitespace-only string that won't be interpreted as user content:

```python
# Current (causes regression):
EMPTY_CONTENT = "."

# Suggested:
new_content.append({"guardContent": {"text": {"text": ""}}})
```

Alternatively, the guardContent block could use a format that Bedrock recognizes as a no-op but that the model won't interpret as user input.

## Environment

- `langchain-aws`: 1.4.3 (regression present), 1.4.1 (works correctly)
- Model: Claude Haiku 4.5 via Bedrock Converse API (Application Inference Profile)
- Agent framework: deepagents (LangGraph-based)
- Guardrail: Bedrock content filter with `guard_last_turn_only=True`

## Workaround

Pin `langchain-aws` to `1.4.1` to avoid the regression. This re-introduces the full-conversation guardrail scan on tool-result turns (#900), which is harmless if your conversation history doesn't contain adversarial content.

michaelnchin:
Fixes: #996 

### Problem Summary

As part of a [fix](https://github.com/langchain-ai/langchain-aws/pull/954) to #990 in ChatBedrockConverse, no-op `guardContent` block will sometimes be inserted to prevent Bedrock from unnecessarily scanning the whole conversation.

Unfortunately, this had the side effect of some models interpreting the `EMPTY_CONTENT` (`"."`) constant used by `guardContent` as actual user input, causing them to ignore tool results entirely.

### Solution

Replacing `"."` with a metadata tag style string `"[tool_result]"` in the no-op `guardContent` instead was able to ensure all of the following over repeated tests with `guard_last_turn_only=True`:

1. Models will (at least from what I saw) never ignore tool results
2. Full-conversation guardrail scans are prevented
3. Bedrock's non-blank `guardContent` validation passes

**Other attempted fixes:**

We did first try to apply some of the cleaner whitespace solutions suggested, but each had their issues -
> Use an empty string "" instead of "."  

Bedrock's server-side validation rejects this specific case with "The guard content field... is blank".

> or use a whitespace-only string that won't be interpreted as user content

We tested several invisible Unicode characters (zero-width space `\u200b`, word joiner `\u2060`, etc.) to varying degress of success. Initially, they appeared to pass Bedrock validation, but stress tests revealed that models will still sometimes interpret them as "empty message". The effectiveness also varied across models, ex. with `\u2060`, Haiku 4.5 had a fail rate of only 5-10%, but switching to Sonnet 4.6 showed closer to 85%.


michaelnchin (commented):
Integ tests passed: https://github.com/langchain-ai/langchain-aws/actions/runs/24447061689/job/71426060016

PR #1007, opened 2026-04-16
fix(aws): strip streaming-only fields from invalid tool_use blocks

### LINKED ISSUE #1010: ChatBedrock: invalid tool_use blocks leak streaming-only fields (index, partial_json)
## Description

When `ChatBedrock` (via `InvokeModelWithResponseStream`) emits tool arguments that fail to parse, the accumulated `AIMessageChunk` content block retains raw Bedrock-streaming fields (`index` and `partial_json`). These are rejected by Bedrock's Anthropic validator on the next API call in a multi-turn conversation:

```
when calling the InvokeModelWithResponseStream operation:
messages.N.content.M.tool_use.index: Extra inputs are not permitted
```

## Root cause

In `_format_anthropic_messages`, when a `tool_use` block's `id` is **not** in `message.tool_calls` (i.e., it's an invalid tool call), the block is appended verbatim — including the streaming-only `index` and `partial_json` fields that were accumulated during chunk merging.

The adjacent `thinking` / `redacted_thinking` branch already strips its own streaming `index` field before appending, but this wasn't done for the `tool_use` path.

## Reproduction

1. Use `ChatBedrock` with streaming and tool calling
2. Trigger a response where the model emits malformed tool arguments (JSON parse failure)
3. langchain-core records this in `AIMessage.invalid_tool_calls`
4. On the next turn, `_format_anthropic_messages` includes the raw block with `index` and `partial_json`
5. Bedrock rejects the request with the validation error above

## Proposed fix

PR #1007 strips `index` and `partial_json` from invalid `tool_use` blocks before appending, consistent with how `thinking` blocks are already handled.

Related: langchain-ai/langchain#31208 (closed)

jdg9vr:
## Summary

Fixes #1010.

When the model emits tool args that fail to parse:
- langchain-core records the failure in `AIMessage.invalid_tool_calls` (instead of `tool_calls`).
- The accumulated `AIMessageChunk` content block keeps the raw Bedrock-streaming state: `input={}`, plus the streaming-protocol-only fields `index` and `partial_json`.

In `_format_anthropic_messages`, the `tool_use` branch has two paths:
- **id in `message.tool_calls`**: rebuilds a clean block via `_lc_tool_calls_to_anthropic_tool_use_blocks` — correct.
- **id not in `message.tool_calls`** (invalid tool call): appended the raw block verbatim. The leaked `index`/`partial_json` fields are rejected by Bedrock's Anthropic validator on the next replay:

  ```
  when calling the InvokeModelWithResponseStream operation:
  messages.N.content.M.tool_use.index: Extra inputs are not permitted
  ```

The fix mirrors what the adjacent `thinking` / `redacted_thinking` branch already does for its own streaming `index` field — strip the accumulation-only fields before append.

## Test plan

- [x] Added `test__format_anthropic_messages_strips_streaming_fields_on_invalid_tool` in `tests/unit_tests/chat_models/test_bedrock.py`. Confirmed it **fails** on `main` and **passes** with this fix (`AssertionError: assert 'index' not in {...}`).
- [x] `uv run pytest tests/unit_tests/chat_models/test_bedrock.py` — 108 passed.
- [x] `uv run --all-groups ruff check` / `ruff format --diff` — clean.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -628,8 +628,13 @@ def _format_anthropic_messages(
                                 )
                             )
                         else:
-                            item.pop("text", None)
-                            tool_blocks.append(item)
+                            tool_blocks.append(
+                                {
+                                    k: v
+                                    for k, v in item.items()
+                                    if k not in ("text", "index", "partial_json")
+                                }
+                            )
                     elif item["type"] in ["thinking", "redacted_thinking"]:
                         # Store thinking blocks separately
                         thinking_blocks.append(
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock.py
@@ -174,6 +174,44 @@ def test__format_anthropic_messages_with_tool_calls() -> None:
     assert expected == actual
 
 
+def test__format_anthropic_messages_strips_streaming_fields_on_invalid_tool() -> None:
+    """Regression for langchain-ai/langchain#31208."""
+    ai = AIMessage(
+        content=[
+            {
+                "type": "tool_use",
+                "id": "toolu_bad",
+                "name": "web_search",
+                "input": {},
+                "index": 2,
+                "partial_json": '{"bad": json}',
+            },
+        ],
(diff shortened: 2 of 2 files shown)


michaelnchin (approved):
Thanks @jdg9vr !

PR #1013, opened 2026-04-20
fix: Avoid polluting environemnt with bedrock bearer token when possible

jacoblee93:
Populate a session and a static token provider instead. Workaround suggested by boto team.

### DIFF
--- a/libs/aws/langchain_aws/utils.py
+++ b/libs/aws/langchain_aws/utils.py
@@ -5,11 +5,24 @@ from abc import abstractmethod
 from typing import Any, Dict, Generic, Iterator, List, Literal, Optional, TypeVar, Union
 
 from botocore.exceptions import BotoCoreError, UnknownServiceError
+from botocore.tokens import FrozenAuthToken
 from langchain_core.messages import AIMessage
 from packaging import version
 from pydantic import SecretStr
 
 logger = logging.getLogger(__name__)
+
+
+class _StaticTokenProvider:
+    """Token provider that returns a static bearer token."""
+
+    METHOD = "static"
+
+    def __init__(self, token: str) -> None:
+        self._token = token
+
+    def load_token(self, **kwargs: Any) -> FrozenAuthToken:
+        return FrozenAuthToken(self._token)
 
 MESSAGE_ROLES = Literal["system", "user", "assistant"]
 MESSAGE_FORMAT = Dict[Literal["role", "content"], Union[MESSAGE_ROLES, str]]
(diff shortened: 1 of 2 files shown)


libs/aws/langchain_aws/utils.py
```diff
@@ -185,7 +187,54 @@ def create_aws_client(
             )
 
         if has_api_key:
-            os.environ["AWS_BEARER_TOKEN_BEDROCK"] = api_key.get_secret_value()  # type: ignore[union-attr]
+
+            class _StaticTokenProvider:
```
3coins: **Nit**: This class is re-defined on every call. Could be hoisted to module level since it has no closure over the function scope - it just takes `token` as a constructor arg.

libs/aws/langchain_aws/utils.py
```diff
@@ -185,7 +187,54 @@ def create_aws_client(
...
+            # Create a botocore session and inject the static token provider
+            bc_session = get_session()
+            token_provider_chain = bc_session.get_component("token_provider")
+            static_provider = _StaticTokenProvider(api_key.get_secret_value())  # type: ignore[union-attr]
+            # NOTE: botocore does not expose a public API for inserting a
+            # provider into the token provider chain, so we mutate the private
+            # ``_providers`` list. This is intentional and is the same approach
+            # recommended for static-token workarounds; it may need to be
+            # revisited if botocore changes this internal structure.
+            token_provider_chain._providers.insert(0, static_provider)
```
3coins: **Nit**: If botocore ever removes or renames `_providers`, this will raise a bare `AttributeError`. Might be worth wrapping in a try/except that raises a clearer error, e.g. "botocore token provider chain internals have changed — bearer token injection is no longer supported."

3coins (approved):
@jacoblee93 
Looks good, with some minor comments.

`create_aws_bedrock_runtime_client` still writes to `os.environ["AWS_BEARER_TOKEN_BEDROCK"]` and has the same multitenant limitation. The botocore token provider approach won't work here since it uses the smithy SDK - but worth a follow-up issue to track finding an equivalent solution for that path.

libs/aws/langchain_aws/utils.py
jacoblee93: Yup good point!

libs/aws/langchain_aws/utils.py
jacoblee93: Yeah fair, will do

jacoblee93:
Thank you!

PR #1018, opened 2026-04-27
feat(aws): Add ChatBedrockMantle for Bedrock Mantle OpenAI-compatible…

rishabhagrawal1:
## Summary

Adds `ChatBedrockMantle`, a `BaseChatModel` wrapping the OpenAI Python SDK to talk to Bedrock Mantle's OpenAI-compatible endpoints (`bedrock-mantle.<region>.api.aws`). Supports Responses API (default), Chat Completions API (fallback), streaming, async, tool calling, structured output, and conversation state via `previous_response_id`.

Auth supports both static API keys and auto-refreshing short-term tokens from AWS credentials via `_BedrockApiKeyProvider`, which follows botocore's `RefreshableCredentials` pattern with advisory (15 min) and mandatory (10 min) refresh windows.

## Usage

```python
from langchain_aws import ChatBedrockMantle

llm = ChatBedrockMantle(model="deepseek.v3.2", region_name="us-east-1")
response = llm.invoke("What is 2+2?")

# Streaming
for chunk in llm.stream("Say hello"):
    print(chunk.content)

# Async
response = await llm.ainvoke("What is 2+2?")
```

Install: `pip install "langchain-aws[mantle]"`

## Files Changed

| File | Change |
|---|---|
| `langchain_aws/chat_models/bedrock_mantle.py` | NEW — `ChatBedrockMantle` implementation |
| `langchain_aws/utils.py` | MODIFIED — Added `_BedrockApiKeyProvider` |
| `langchain_aws/chat_models/__init__.py` | MODIFIED — Added to `__all__` and lazy imports |
| `langchain_aws/__init__.py` | MODIFIED — Added to exports |
| `pyproject.toml` | MODIFIED — Added `[mantle]` optional dep (`openai>=1.106.0`) |
| `tests/unit_tests/chat_models/test_bedrock_mantle.py` | NEW — 68 unit tests |
| `tests/integration_tests/chat_models/test_bedrock_mantle.py` | NEW — 22 integration tests |

## Testing

- 68 unit tests — all passing
- 22 integration tests — all passing (live Mantle, us-east-1)
- Verified against `openai==1.106.0` (declared floor) — all passing
- Full unit suite (884+) — 0 failures
- Lint clean (ruff)

- `_BedrockApiKeyProvider` token refresh logic in `utils.py` — follows botocore's `RefreshableCredentials` pattern with non-blocking advisory and blocking mandatory windows
- API routing in `_use_responses_api` — auto-detects Responses vs Chat Completions based on parameter keys
- `openai>=1.106.0` version pin — minimum for callable `api_key` support (short-term token refresh)

## Note

AI agents were used to assist with code review and refactoring during development.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_mantle.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_mantle.py
@@ -1,0 +1,1066 @@
+"""ChatBedrockMantle — LangChain chat model for Amazon Bedrock Mantle.
+
+Wraps the OpenAI Python SDK to talk to Mantle's Chat Completions and Responses
+APIs at ``bedrock-mantle.<region>.api.aws``.  Supports both long-term API keys
+and auto-refreshing short-term API keys generated from AWS credentials.
+
+Install with::
+
+    pip install "langchain-aws[mantle]"
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+from typing import (
+    Any,
+    AsyncIterator,
+    Dict,
+    Iterator,
+    List,
+    Optional,
+    Sequence,
+)
+
+from langchain_core.callbacks import (
+    AsyncCallbackManagerForLLMRun,
+    CallbackManagerForLLMRun,
+)
+from langchain_core.language_models import BaseChatModel
+from langchain_core.language_models.base import LangSmithParams
+from langchain_core.messages import (
+    AIMessage,
+    AIMessageChunk,
+    BaseMessage,
+    convert_to_openai_messages,
+)
+from langchain_core.messages.ai import UsageMetadata
+from langchain_core.messages.tool import tool_call_chunk
+from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
+from langchain_core.utils.function_calling import convert_to_openai_tool
+from pydantic import ConfigDict, Field, SecretStr, model_validator
+from typing_extensions import Self
+
+logger = logging.getLogger(__name__)
+
+# Keys that only exist in the Responses API — their presence forces that path.
+_RESPONSES_ONLY_KEYS = frozenset(
+    {
+        "previous_response_id",
+        "text",
+    }
+)
+
+# Keys that only exist in Chat Completions — their presence forces that path.
+_CHAT_COMPLETIONS_ONLY_KEYS = frozenset(
+    {
+        "response_format",
+    }
+)
+
+_MANTLE_BASE_URL_TEMPLATE = "https://bedrock-mantle.{region}.api.aws/v1"
+
+
+def _openai_response_to_ai_message(choice: Any) -> AIMessage:
+    """Convert an OpenAI ChatCompletion choice to an AIMessage."""
+    message = choice.message
+    tool_calls = []
+    if message.tool_calls:
+        for tc in message.tool_calls:
+            args = tc.function.arguments
+            try:
+                parsed = json.loads(args)
+            except (ValueError, TypeError):
+                parsed = args if isinstance(args, dict) else {"raw": args}
+            if not isinstance(parsed, dict):
+                parsed = {"raw": parsed}
+            tool_calls.append(
+                {
+                    "name": tc.function.name,
+                    "args": parsed,
+                    "id": tc.id,
+                    "type": "tool_call",
+                }
+            )
+
+    return AIMessage(
+        content=message.content or "",
+        tool_calls=tool_calls,
+        additional_kwargs={
+            k: v
+            for k, v in {
+                "refusal": getattr(message, "refusal", None),
+            }.items()
+            if v is not None
+        },
+        response_metadata={
+            "finish_reason": choice.finish_reason,
+        },
+    )
+
+
+def _handle_openai_error(error: Exception) -> None:
+    """Handle OpenAI SDK errors with enhanced messages for Mantle.
+
+    Translates OpenAI SDK exceptions into actionable error messages
+    specific to the Bedrock Mantle context.
+    """
+    try:
+        from openai import (
+            AuthenticationError,
+            BadRequestError,
+            NotFoundError,
+            RateLimitError,
+        )
+    except ImportError:
+        raise error
+
+    if isinstance(error, AuthenticationError):
+        raise ValueError(
+            "Bedrock Mantle authentication failed. Your API key or "
+            "short-term token may be invalid or expired. If using "
+            "short-term tokens, ensure your AWS credentials are valid. "
+            "If using a static API key, verify it in the Bedrock Console. "
+            f"Original error: {error}"
+        ) from error
+
+    if isinstance(error, RateLimitError):
+        raise ValueError(
+            "Bedrock Mantle rate limit exceeded. Consider reducing request "
+            "frequency or requesting a quota increase via AWS Support. "
+            f"Original error: {error}"
+        ) from error
+
+    if isinstance(error, NotFoundError):
+        raise ValueError(
+            "Bedrock Mantle model not found. Verify the model ID is correct "
+            "and available in your region. Use list_models() to see available "
+            f"models. Original error: {error}"
+        ) from error
+
+    if isinstance(error, BadRequestError):
+        msg = str(error)
+        if "tool_choice" in msg:
+            raise ValueError(
+                "Bedrock Mantle rejected the tool_choice value. "
+                f"Original error: {error}"
+            ) from error
+        if "does not support" in msg:
+            raise ValueError(
+                "Bedrock Mantle: the requested feature is not supported "
+                f"for this model. Original error: {error}"
+            ) from error
+        raise ValueError(
+            f"Bedrock Mantle bad request. Original error: {error}"
+        ) from error
+
+    raise error
+
+
+def _parse_responses_tool_args(args: Any) -> Dict[str, Any]:
+    """Parse tool-call arguments from the Responses API into a dict.
+
+    LangChain expects tool-call ``args`` to be a ``dict``.  The Responses
+    API may return a JSON string, a dict, a list, or ``None``.
+    """
+    try:
+        parsed = json.loads(args) if isinstance(args, str) else args
+    except (ValueError, TypeError):
+        parsed = args if isinstance(args, dict) else {"raw": args}
+    if not isinstance(parsed, dict):
+        parsed = {"raw": parsed}
+    return parsed
+
+
+def _extract_responses_output(response: Any) -> tuple[str, list]:
+    """Extract content text and tool calls from a Responses API response."""
+    content = ""
+    tool_calls: list = []
+    if response.output:
+        for item in response.output:
+            if hasattr(item, "content"):
+                for block in item.content:
+                    if hasattr(block, "text"):
+                        content += block.text
+            elif hasattr(item, "name"):
+                args = item.arguments if hasattr(item, "arguments") else ""
+                tool_calls.append(
+                    {
+                        "name": item.name,
+                        "args": _parse_responses_tool_args(args),
+                        "id": (item.call_id if hasattr(item, "call_id") else item.id),
+                        "type": "tool_call",
+                    }
+                )
+    return content, tool_calls
+
+
+def _get_last_messages_and_response_id(
+    messages: Sequence[BaseMessage],
+) -> tuple[list[BaseMessage], Optional[str]]:
+    """Extract the most recent ``previous_response_id`` from message history.
+
+    Scans backward for the last ``AIMessage`` whose ``id`` starts with
+    ``resp_``.  Returns the messages after that point and the response ID.
+    If no such message is found, returns all messages and ``None``.
+    """
+    for i in range(len(messages) - 1, -1, -1):
+        msg = messages[i]
+        if isinstance(msg, AIMessage):
+            resp_id = getattr(msg, "id", None) or ""
+            if resp_id.startswith("resp_"):
+                return list(messages[i + 1 :]), resp_id
+    return list(messages), None
+
+
+def _build_chat_completions_result(response: Any) -> ChatResult:
+    """Build a ``ChatResult`` from a Chat Completions API response."""
+    choice = response.choices[0]
+    ai_msg = _openai_response_to_ai_message(choice)
+
+    if response.usage:
+        ai_msg.usage_metadata = UsageMetadata(
+            input_tokens=response.usage.prompt_tokens,
+            output_tokens=response.usage.completion_tokens,
+            total_tokens=response.usage.total_tokens,
+        )
+
+    ai_msg.response_metadata["model"] = response.model
+    ai_msg.response_metadata["id"] = response.id
+    return ChatResult(generations=[ChatGeneration(message=ai_msg)])
+
+
+def _build_responses_result(response: Any) -> ChatResult:
+    """Build a ``ChatResult`` from a Responses API response object."""
+    content, tool_calls = _extract_responses_output(response)
+
+    ai_msg = AIMessage(
+        content=content,
+        tool_calls=tool_calls,
+        id=response.id,
+        response_metadata={
+            "id": response.id,
+            "model": response.model,
+            "status": response.status if hasattr(response, "status") else None,
+        },
+    )
+
+    if hasattr(response, "usage") and response.usage:
+        ai_msg.usage_metadata = UsageMetadata(
+            input_tokens=response.usage.input_tokens,
+            output_tokens=response.usage.output_tokens,
+            total_tokens=response.usage.total_tokens,
+        )
+
+    return ChatResult(generations=[ChatGeneration(message=ai_msg)])
+
+
+class ChatBedrockMantle(BaseChatModel):
+    """LangChain chat model for Amazon Bedrock via the Mantle inference engine.
+
+    Supports models available on the ``bedrock-mantle`` endpoint using the
+    OpenAI-compatible Responses API and Chat Completions API.
+
+    Setup:
+        Install the optional Mantle dependencies::
+
+            pip install "langchain-aws[mantle]"
+
+    Instantiate:
+        .. code-block:: python
+
+            from langchain_aws import ChatBedrockMantle
+
+            llm = ChatBedrockMantle(
+                model="deepseek.v3.2",
+                region_name="us-east-1",
+            )
+
+    Invoke:
+        .. code-block:: python
+
+            response = llm.invoke("What is 2+2?")
+
+    Stream:
+        .. code-block:: python
+
+            for chunk in llm.stream("Say hello in 3 languages"):
+                print(chunk.content)
+
+    Async:
+        .. code-block:: python
+
+            response = await llm.ainvoke("What is 2+2?")
+
+    Conversation state with previous_response_id:
+        .. code-block:: python
+
+            llm = ChatBedrockMantle(
+                model="deepseek.v3.2",
+                region_name="us-east-1",
+            )
+            response = llm.invoke("Hello")
+            response_id = response.response_metadata["id"]
+            followup = llm.invoke(
+                "Tell me more",
+                previous_response_id=response_id,
+            )
+    """
+
+    model: str = Field(..., alias="model")
+    """Mantle model ID, e.g. ``deepseek.v3.2``, ``qwen.qwen3-32b``."""
+
+    region_name: Optional[str] = None
+    """AWS region. Used to construct the Mantle endpoint URL."""
+
+    base_url: Optional[str] = None
+    """Override the Mantle endpoint URL."""
+
+    temperature: Optional[float] = None
+    max_tokens: Optional[int] = None
+    model_kwargs: Dict[str, Any] = Field(default_factory=dict)
+    """Additional parameters passed to the underlying SDK."""
+
+    use_responses_api: Optional[bool] = None
+    """Whether to use the Responses API instead of Chat Completions.
+
+    Applies to OpenAI-compatible models only.
+
+    - ``True``: always use the Responses API.
+    - ``False``: always use the Chat Completions API.
+    - ``None`` (default): auto-detect per request. Defaults to the
+      Responses API unless Chat-Completions-only parameters (e.g.
+      ``response_format``) are present.
+    """
+
+    use_previous_response_id: bool = False
+    """If ``True``, automatically pass ``previous_response_id``.
+
+    Uses the ID of the most recent ``AIMessage`` in the input message
+    sequence. Messages up to that response are dropped from the request
+    payload. Applies to OpenAI-compatible models only (Responses API).
+    """
+
+    stream_usage: Optional[bool] = None
+    """Whether to include usage metadata in streaming output.
+
+    Applies to OpenAI-compatible models only.
+
+    If ``True``, ``stream_options={"include_usage": True}`` is sent.
+    If ``False``, usage is not requested.  If ``None`` (default),
+    usage is included.
+    """
+
+    max_retries: int = 2
+    """Maximum number of retries for transient errors (5xx, connection, timeout)."""
+
+    timeout: Optional[float] = None
+    """Request timeout in seconds. Applies to both connect and read."""
+
+    bedrock_api_key: Optional[SecretStr] = Field(default=None, alias="api_key")
+    """Static Bedrock API key. If provided, short-term key generation is skipped."""
+
+    _sync_client: Any = None
+    _async_client: Any = None
+    _resolved_region: str = ""
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    @model_validator(mode="after")
+    def _validate_environment(self) -> Self:
+        try:
+            from openai import AsyncOpenAI, OpenAI
+        except ImportError as exc:
+            raise ModuleNotFoundError(
+                "Could not import openai. "
+                'Please install it via: pip install "langchain-aws[mantle]"'
+            ) from exc
+
+        region = (
+            self.region_name
+            or os.environ.get("AWS_REGION")
+            or os.environ.get("AWS_DEFAULT_REGION")
+        )
+        if not region and not self.base_url:
+            try:
+                import botocore.session
+
+                region = botocore.session.Session().get_config_variable("region")
+            except (ImportError, OSError, ValueError) as exc:
+                # ImportError: botocore not installed
+                # OSError: config file not readable
+                # ValueError: malformed config
+                logger.debug("Could not resolve region from botocore: %s", exc)
+        if not region and not self.base_url:
+            raise ValueError(
+                "region_name must be provided, or set AWS_REGION / "
+                "AWS_DEFAULT_REGION, or provide base_url."
+            )
+        self._resolved_region = region or ""
+
+        url = self.base_url or _MANTLE_BASE_URL_TEMPLATE.format(
+            region=self._resolved_region
+        )
+
+        if self.bedrock_api_key:
+            api_key: Any = self.bedrock_api_key.get_secret_value()
+        else:
+            try:
+                from importlib.util import find_spec
+
+                if find_spec("aws_bedrock_token_generator") is None:
+                    raise ImportError("aws_bedrock_token_generator not found")
+            except ImportError as exc:
+                raise ModuleNotFoundError(
+                    "Could not import aws-bedrock-token-generator. "
+                    'Please install it via: pip install "langchain-aws[mantle]"'
+                ) from exc
+
+            from langchain_aws.utils import _BedrockApiKeyProvider
+
+            api_key = _BedrockApiKeyProvider(self._resolved_region)
+
+        client_kwargs: Dict[str, Any] = {
+            "api_key": api_key,
+            "base_url": url,
+            "max_retries": self.max_retries,
+            "default_headers": {
+                "x-client-framework": "langchain-aws",
+            },
+        }
+        if self.timeout is not None:
+            client_kwargs["timeout"] = self.timeout
+
+        self._sync_client = OpenAI(**client_kwargs)
+        async_kwargs = {**client_kwargs}
+        if callable(api_key) and hasattr(api_key, "async_call"):
+            async_kwargs["api_key"] = api_key.async_call
+        self._async_client = AsyncOpenAI(**async_kwargs)  # type: ignore[arg-type]
+        return self
+
+    @property
+    def _llm_type(self) -> str:
+        return "bedrock-mantle"
+
+    def _get_ls_params(
+        self, stop: Optional[List[str]] = None, **kwargs: Any
+    ) -> LangSmithParams:
+        """Get standard params for LangSmith tracing."""
+        ls_params = LangSmithParams(
+            ls_provider="amazon_bedrock_mantle",
+            ls_model_name=self.model,
+            ls_model_type="chat",
+            ls_temperature=self.temperature,
+        )
+        if self.max_tokens:
+            ls_params["ls_max_tokens"] = self.max_tokens
+        if stop:
+            ls_params["ls_stop"] = stop
+        return ls_params
+
+    def _use_responses_api(self, params: Dict[str, Any]) -> bool:
+        """Decide whether to route this request to the Responses API.
+
+        If ``use_responses_api`` is set explicitly (``True`` / ``False``),
+        that value is honored.  Otherwise defaults to the Responses API
+        unless Chat-Completions-only parameters are present.
+        """
+        if isinstance(self.use_responses_api, bool):
+            return self.use_responses_api
+
+        # If Responses-API-only keys are present, always use Responses API
+        if _RESPONSES_ONLY_KEYS.intersection(params):
+            return True
+
+        # Fall back to Chat Completions for Chat-Completions-only keys
+        if _CHAT_COMPLETIONS_ONLY_KEYS.intersection(params):
+            return False
+
+        # Default: Responses API
+        return True
+
+    def _generate(
+        self,
+        messages: List[BaseMessage],
+        stop: Optional[List[str]] = None,
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+        **kwargs: Any,
+    ) -> ChatResult:
+        if self.use_previous_response_id:
+            messages, prev_id = _get_last_messages_and_response_id(messages)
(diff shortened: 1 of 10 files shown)


libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+    """Whether to use the Responses API instead of Chat Completions.
+
+    Applies to OpenAI-compatible models only.
+
+    - ``True``: always use the Responses API.
+    - ``False``: always use the Chat Completions API.
+    - ``None`` (default): auto-detect per request. Defaults to the
+      Responses API unless Chat-Completions-only parameters (e.g.
+      ``response_format``) are present.
+    """
```
3coins, line 508: Do we want to support chat completions as responses API is the standard since Mar 2025?

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+
+        return None
+
+    def bind_tools(
+        self,
+        tools: Sequence[Any],
+        *,
+        tool_choice: Optional[str | Dict] = None,
+        **kwargs: Any,
+    ) -> "ChatBedrockMantle":
```
3coins:
```suggestion
    ) -> Runnable[LanguageModelInput, BaseMessage]:
```

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+                "name": schema_name,
+                "schema": json_schema,
+                "strict": True,
+            }
+            llm = self.bind(text={"format": text_format}, **kwargs)
+            output_parser = (
+                PydanticOutputParser(pydantic_object=schema)
+                if is_pydantic
+                else JsonOutputParser()
+            )
```
3coins: The **with_structured_output(method="json_schema")** should use `response_format` (the standard OpenAI parameter) instead of the responses API specific text parameter. The impact of this is that users can't switch between ChatOpenAI and ChatBedrockMantle without behavior changes. 

What **ChatOpenAI** does:                                                                                                     
                                                                                                                              
For **method="json_schema"**, it converts the schema into an OpenAI response_format dict and binds it:                          

```python                                                                                                                              
response_format = _convert_to_openai_response_format(schema, strict=strict)                                                 
llm = self.bind(response_format=response_format, **kwargs)                                                                  
```
                                                                                                                              
The response_format parameter works with both Chat Completions (natively) and Responses API (via                    _construct_responses_api_payload which remaps it). The OpenAI SDK's chat.completions.parse() method handles response_format with json_schema type directly.
                                                                                                                              
What the PR should do:

```python                                                                                                                              
  elif method == "json_schema":                                                                                               
      if is_pydantic:                                                                                                         
          json_schema = schema.model_json_schema()                                                                            
      elif isinstance(schema, dict):                                                                                          
          json_schema = schema                                                                                                
      else:                                                                                                                   
          msg = (                                                                                                             
              f"Unsupported schema type for json_schema method: "                                                             
              f"{type(schema)}. Use a Pydantic model or dict."                                                                
          )                                                                                                                   
          raise ValueError(msg)                                                                                               
                                                                                                                              
      schema_name = json_schema.get(                                                                                          
          "title", getattr(schema, "__name__", "Output")                                                                      
      )                                                                                                                       
      response_format = {                                                                                                     
          "type": "json_schema",                                                                                              
          "json_schema": {                                                                                                    
              "name": schema_name,                                                                                            
              "schema": json_schema,                                                                                          
              "strict": True,                                                                                                 
          },                                                                                                                  
      }                                                                                                                       
      llm = self.bind(response_format=response_format, **kwargs)                                                              
      output_parser = (                                                                                                       
          PydanticOutputParser(pydantic_object=schema)                                                                        
          if is_pydantic                                                                                                      
          else JsonOutputParser()                                                                                             
      )                                                                                                                       
```
                                                                                                                              
  This way:                                                                                                                   
                                                                                                                              
- response_format is the standard OpenAI parameter that both APIs understand                                                
- The routing logic already sends response_format to Chat Completions (since _CHAT_COMPLETIONS_ONLY_KEYS contains "response_format")      
- It matches ChatOpenAI's behavior exactly                                                                                  
- No need for the warning about use_responses_api=False — it just works on both paths                                       
                                                                                                                            
The one caveat is that _prepare_responses_params would need to remap response_format to the Responses API text format if the request ends up going through the Responses API path. But the simpler approach is: response_format naturally routes to Chat Completions via the existing _CHAT_COMPLETIONS_ONLY_KEYS check, which is actually the correct behavior, json_schema   structured output works best through chat.completions.parse() anyway.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+            }.items()
+            if v is not None
+        },
+        response_metadata={
+            "finish_reason": choice.finish_reason,
+        },
+    )
+
+
+def _handle_openai_error(error: Exception) -> None:
```
3coins, line 106: This wraps typed OpenAI errors into generic `ValueError`. This loses the error type hierarchy. Code that catches **openai.RateLimitError** or **openai.AuthenticationError** won't work, they'll get ValueError instead. ChatOpenAI re-raises the original OpenAI errors (with special handling only for BadRequestError context overflow). This is a behavioral regression for anyone writing error-handling code.                                                                          

Missing fields compared to `ChatOpenAI`. The following Responses API features supported by ChatOpenAI are absent:         
  - reasoning / reasoning_effort — for reasoning models                                                                     
  - truncation — context management                                                                                         
  - context_management — compaction                                                                                         
  - include — additional response fields                                                                                    
  - output_version — AIMessage format versioning                                                                            
  - disabled_params — parameter filtering                                                                                   
  - parallel_tool_calls — tool calling control                                                                              
  - strict parameter on bind_tools and with_structured_output                                                               
                                                                                                                            
Not all of these may be relevant for Mantle, but reasoning, strict, and parallel_tool_calls are commonly used.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+      ``response_format``) are present.
+    """
+
+    use_previous_response_id: bool = False
+    """If ``True``, automatically pass ``previous_response_id``.
+
+    Uses the ID of the most recent ``AIMessage`` in the input message
+    sequence. Messages up to that response are dropped from the request
+    payload. Applies to OpenAI-compatible models only (Responses API).
+    """
```
3coins, line 537: `previous_response_id` + Checkpointer/Memory would cause a dual state problem. The core issue is that use_previous_response_id=True creates two independent sources of truth for conversation state:                                               
                                                                                                                                                      
As per my understanding, this is how it works in the PR:
                                                                                                                                                      
1. LangGraph checkpointer saves the full message history (all HumanMessage, AIMessage, ToolMessage)                                                                 
2. On the next invocation, the checkpointer restores all messages into state["messages"]                                                                            
3. _get_last_messages_and_response_id() scans backward, finds the last AIMessage with id starting with resp_, and drops everything before it                        
4. Only the messages after that AIMessage + previous_response_id are sent to Mantle                                                                                 
5. Mantle's server-side state reconstructs the full context from the response ID                                                                                    
                                                                                                                                                      
The risks:                                                                                                                                                          
                                                                                                                                                      
1. Server-side state expiry. Mantle's Responses API stores conversation state server-side, but this has a TTL (likely hours/days, not documented in the PR). If a LangGraph graph is resumed after the TTL expires, the previous_response_id becomes invalid. The checkpointer has the full history, but the code has no fallback - it doesn't retry with the full message list when the response ID is rejected. ChatOpenAI has the same limitation, but OpenAI's state retention is well-documented. Mantle's isn't.
2. Tool messages silently dropped. In an agent loop: Human → AI(tool_call, id=resp_1) → Tool(result) → AI(final, id=resp_2) → Human(followup). With use_previous_response_id=True, only [Human(followup)] is sent with previous_response_id=resp_2. The intermediate tool call and result are assumed to be in       server-side state. If they're not (e.g., state expired, different endpoint), the model loses that context entirely.                                                 
3. Cross-region/endpoint resume breaks. If a LangGraph graph checkpoints in one region and resumes in another (or the base_url changes), the resp_ IDs are meaningless to the new endpoint. The checkpointer has the messages but they'll never be sent.                                                       
4. `msg.id` vs `msg.response_metadata["id"]` mismatch. The PR checks msg.id (the LangChain message ID field), while ChatOpenAI checks msg.response_metadata.get("id"). This matters because:                                                                                                              
    - For Responses API: the PR sets AIMessage(id=response.id), so msg.id == "resp_xxx" ✓                                                                             
    - For Chat Completions API: _build_chat_completions_result sets ai_msg.response_metadata["id"] = response.id but does NOT set ai_msg.id — so msg.id gets a random LangChain UUID, not chatcmpl_xxx                                         
    - If a conversation mixes both API paths, the msg.id check could pick up a stale Responses API message even when the most recent exchange used Chat Completions   
                                                                                                                                                        
5. Redundant token usage. The checkpointer stores all messages, but they're never sent to the API. Users pay for checkpointer storage (DynamoDB, etc.) for messages that serve no purpose when use_previous_response_id=True. This isn't a bug, but it's a cost/confusion concern worth documenting.                                    
                                                                                                                                                      
What `ChatOpenAI` does identically (same risks): ChatOpenAI has the exact same _get_last_messages pattern and the same lack of fallback. So this isn't unique to this PR, but it's worth noting that ChatOpenAI at least documents the behavior and has the truncation="auto" option as a safety net for context overflow. The PR has neither.
                                                                                                                                                      
**Recommendation**: At minimum, the PR should:
                                                                                                                                                      
- Document that use_previous_response_id=True relies on Mantle server-side state and what happens when it expires                                                   
- Consider adding a fallback: if the API returns an error for an invalid previous_response_id, retry with the full message history from the checkpointer 
- Align the ID extraction to use response_metadata["id"] like ChatOpenAI does, not msg.id

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+        ai_msg.usage_metadata = UsageMetadata(
+            input_tokens=response.usage.input_tokens,
+            output_tokens=response.usage.output_tokens,
+            total_tokens=response.usage.total_tokens,
+        )
+
+    return ChatResult(generations=[ChatGeneration(message=ai_msg)])
+
+
+class ChatBedrockMantle(BaseChatModel):
```
3coins, line 432: `_identifying_params` or `_get_invocation_params` are used by LangSmith and callbacks for tracing. The _get_ls_params is implemented, which is good, but the other two are missing.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Keys that only exist in the Responses API — their presence forces that path.
+_RESPONSES_ONLY_KEYS = frozenset(
+    {
+        "previous_response_id",
+        "text",
+    }
+)
```
3coins, line 56: `ChatOpenAI` also includes `"reasoning"`, `"truncation"`, `"context_management"`, `"include"`,                    `"previous_response_id"`. The routing logic is simpler but may not be forward-compatible as Mantle adds features. Are there any Bedrock specific restrictions here, that this list is shorter?

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+                    "type": "function",
+                    **tc["function"],
+                }
+                tc = resp_params["tool_choice"]
+            if isinstance(tc, str) and tc not in ("auto", "none"):
+                logger.warning(
+                    "Bedrock Mantle Responses API may only support "
+                    "tool_choice='auto' or 'none'. Got %r — the API "
+                    "may reject this.",
+                    tc,
```
3coins: Is this an actual restriction of the API? It seems overly restrictive, `_prepare_responses_params` for example `"required"` is a valid Responses API value. `ChatOpenAI` supports `"required"` and also converts dict-format toolchoice (e.g., {"type": "function", "function": {"name": "..."}})  to the Responses API format.

libs/aws/tests/unit_tests/chat_models/test_bedrock_mantle.py
3coins, line 1: There are standard tests from LangChain, that we should implement as well, and make sure they pass.
https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/tests/unit_tests/test_standard.py

libs/aws/tests/integration_tests/chat_models/test_bedrock_mantle.py
3coins, line 1: Not a blocker but similar to unit tests, there is standard test suite, that we should implement. The only caveat here is that I am not aware if the AWS account in CI needs separate permissions, and if that is setup already. I think we should add the tests, verify they pass, but add a skip marker with a message that unskip once the CI has permissions for the API.
https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/tests/integration_tests/chat_models/test_standard.py

3coins (commented):
@rishabhagrawal1 
Thanks for submitting this change. I have several comments, I think we should start with making sure standard tests pass, that would catch any BaseChatModel incompatibilities. Some of the recommendations are simple enough, but others might require more thought. 
Unless we expect API changes due to any of the comments, we can defer some of these as follow up work.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
+"""ChatBedrockMantle — LangChain chat model for Amazon Bedrock Mantle.
+
+Wraps the OpenAI Python SDK to talk to Mantle's Chat Completions and Responses
+APIs at ``bedrock-mantle.<region>.api.aws``.  Supports both long-term API keys
+and auto-refreshing short-term API keys generated from AWS credentials.
+
+Install with::
+
+    pip install "langchain-aws[mantle]"
```
Keyur-S-Patel: lets remove this installation instruction class level comments, to follow similar pattern from other chat model classes

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+    )
+
+
+def _handle_openai_error(error: Exception) -> None:
+    """Handle OpenAI SDK errors with enhanced messages for Mantle.
+
+    Translates OpenAI SDK exceptions into actionable error messages
+    specific to the Bedrock Mantle context.
+    """
+    try:
```
Keyur-S-Patel, line 113: I would just the bubble error message without adding additional text to message, it makes it long and does not add much value.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+        resp_params = self._prepare_responses_params(params)
+        openai_msgs = convert_to_openai_messages(messages)
+        logger.debug(
+            "ChatBedrockMantle streaming (responses) with params: %s", resp_params
+        )
+
+        try:
+            stream = self._sync_client.responses.create(
+                input=openai_msgs, stream=True, **resp_params
+            )
```
Keyur-S-Patel, line 985: Might have to close the stream here 
i.e. 
https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/llms/bedrock.py#L1345

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+                run_manager.on_llm_new_token(token, chunk=gen_chunk)
+            yield gen_chunk
+
+    def _stream_responses(
+        self,
+        messages: List[BaseMessage],
+        params: Dict[str, Any],
+        run_manager: Optional[CallbackManagerForLLMRun] = None,
+    ) -> Iterator[ChatGenerationChunk]:
+        """Stream via the Responses API (``/v1/responses``)."""
```
Keyur-S-Patel, line 975: We should double check if we need to implement provider specific logic here. Mostly we dont need to but good to  double check 

https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/llms/bedrock.py#L1245

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+            content=content,
+            tool_call_chunks=tc_chunks,
+        )
+
+        if chunk.choices[0].finish_reason:
+            msg_chunk.response_metadata["finish_reason"] = chunk.choices[
+                0
+            ].finish_reason
+
+        if hasattr(chunk, "usage") and chunk.usage:
```
Keyur-S-Patel: worth adding unit test on this

libs/aws/langchain_aws/utils.py
```diff
@@ -495,3 +498,109 @@ def trim_message_whitespace(messages: List[Any]) -> List[Any]:
...
+    _MANDATORY_REFRESH_SECONDS = 600  # must refresh 10 min before expiry
+
+    def __init__(self, region: str) -> None:
+        self._region = region
+        self._token: Optional[str] = None
+        self._expires_at: float = 0.0
+        self._lock = threading.Lock()
+
+    def __call__(self) -> str:
+        """Return a cached API key, refreshing if near expiry.
```
Keyur-S-Patel, line 603: instead of `# type: ignore[return-value]` ignoring return type we should match to avoid miss matching fields later in the execution

libs/aws/langchain_aws/utils.py
```diff
@@ -495,3 +498,109 @@ def trim_message_whitespace(messages: List[Any]) -> List[Any]:
                     last_message.content[j] = block_dict
 
     return messages
+
+
+class _BedrockApiKeyProvider:
```
Keyur-S-Patel, line 565: This class misses the unit test, we should add unit test to make sure, multi-thread related raise condition are handled, auto refresh is working as expected.

libs/aws/tests/integration_tests/chat_models/test_bedrock_mantle.py
```diff
@@ -0,0 +1,341 @@
...
+
+These tests require:
+  - pip install "langchain-aws[mantle]"
+  - Valid AWS credentials with Bedrock access
+  - Access to Mantle models in the configured region
+
+Run with:
+  AWS_REGION=us-east-1 \
+    uv run --group test --group test_integration \
+    pytest tests/integration_tests/chat_models/test_bedrock_mantle.py -v
```
Keyur-S-Patel: should remove the installation comments from this class to match package standards.

libs/aws/tests/unit_tests/chat_models/test_bedrock_mantle.py
Keyur-S-Patel: `ChatModelIntegrationTests` is for integration test

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
```diff
@@ -0,0 +1,1066 @@
...
+            AuthenticationError,
+            BadRequestError,
+            NotFoundError,
+            RateLimitError,
+        )
+    except ImportError:
+        raise error
+
+    if isinstance(error, AuthenticationError):
+        raise ValueError(
```
hardik-k-shah: What is the specific motivation for wrapping OpenAI errors into a generic type? 

While it simplifies the caller's code by removing the need for OpenAI-specific imports, it makes programmatic error handling more difficult. Callers would be forced to parse error messages for actionable insights rather than catching specific error types.

libs/aws/langchain_aws/utils.py
```diff
@@ -495,3 +498,109 @@ def trim_message_whitespace(messages: List[Any]) -> List[Any]:
...
+    The instance is callable (sync) and also exposes an ``async_call`` method
+    for use with async clients.
+
+    See: https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-generate.html
+    """
+
+    # Bedrock short-term API keys have a 12 h max lifetime as of 2026-04.
+    # If the service changes this, update accordingly or parse the actual
+    # expiry from the token response when available.
+    _TTL_SECONDS = 43200  # 12 h max
```
hardik-k-shah: is it guaranteed to get 12h life time token even if underlying IAM creds might be expiring within 1 hr?   

Should we let caller configure it?

libs/aws/langchain_aws/utils.py
```diff
@@ -495,3 +498,109 @@ def trim_message_whitespace(messages: List[Any]) -> List[Any]:
...
+    # Matches botocore's _DEFAULT_ADVISORY_REFRESH_TIMEOUT (900s) and
+    # _DEFAULT_MANDATORY_REFRESH_TIMEOUT (600s) for credential refresh.
+    _ADVISORY_REFRESH_SECONDS = 900  # try to refresh 15 min before expiry
+    _MANDATORY_REFRESH_SECONDS = 600  # must refresh 10 min before expiry
```
hardik-k-shah, line 580: isn't 10 min too aggressive? should we do mandatory refresh just 2 min in advance instead 10 min?

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Removed.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Updated set is - {"context_management", "previous_response_id", "reasoning", "text", "truncation"}
The only difference is include - which Mantle rejects with a 400 validation error.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Error handling: Fixed. _handle_openai_error no longer wraps typed OpenAI exceptions into ValueError. AuthenticationError, RateLimitError, NotFoundError, and generic BadRequestError now propagate unchanged - callers can catch them by type as expected. Only two specific BadRequestError cases (tool_choice rejection and "does not support" messages) are wrapped with Mantle-specific guidance, using from error to preserve the chain.

Missing fields: Added reasoning, truncation, and context_management to _RESPONSES_ONLY_KEYS. Added strict and parallel_tool_calls as explicit parameters on bind_tools. with_structured_output(method="function_calling") now sets parallel_tool_calls=False internally. Added disabled_params field and _filter_disabled_params() method for per-model parameter filtering. include excluded - Mantle rejects it. output_version not applicable - ChatOpenAI-internal for AIMessage format migration.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Fixed. Original errors now propagate unchanged. Only two specific BadRequestError patterns are wrapped with brief context.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Updated. Original OpenAI errors now re-raise unchanged. Only two specific BadRequestError patterns ("tool_choice" and "does not support") are wrapped to add Mantle-specific guidance, with the original error chained via from error.

ketiko:
Bedrock is starting to only launch new models on Mantle. It would be great to get this in even if it needs refinement over time.

rishabhagrawal1:
Thanks Ryan. Will be prioritizing to close it this week.

akamick86:
> Thanks Ryan. Will be prioritizing to close it this week.

Love it, would be much appreciated.

akamick86:
@ketiko were you able to make progress on this?

libs/aws/tests/unit_tests/chat_models/test_bedrock_mantle.py
rishabhagrawal1: Added both in next commit. Thanks for the suggestion.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Chat Completions is still required for models Mantle serves today. For example - qwen and deepseek models only work through /v1/chat/completions

ketiko:
> Ryan Hansen (@ketiko) were you able to make progress on this?

I'm not with langchain. Just someone hoping to get Mantle support in langchain.

rishabhagrawal1:
On it. Almost done. There were some changes in supported endpoint URL which is causing additional work.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Updated to `Runnable[LanguageModelInput, AIMessage]`.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Done - matches with preferred simpler approach: `response_format` is now in `_CHAT_COMPLETIONS_ONLY_KEYS`, so `with_structured_output(method="json_schema")` naturally routes through `chat.completions.parse()`. 
The `_prepare_responses_params` remap is still there as a safety net for the case where callers explicitly set `use_responses_api=True` (e.g., GPT-5 which is Responses-only), so both paths produce valid requests. 
Covered by `test_json_schema_routes_to_chat_completions` and `test_json_schema_remapped_when_forced_to_responses_api`.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Thanks for such detailed feedback. Addressed the two mechanical items:

1. **Docstring** - expanded `use_previous_response_id` field with the four operational trade-offs: server-side state dependency + expiry (with explicit note that we don't retry on rejection), non-portability across regions/endpoints, silent dropping of intermediate tool messages, and redundant checkpointer storage.
2. **ID extraction** - switched `_get_last_messages_and_response_id` to read `msg.response_metadata.get("id")`, matching ChatOpenAI. `_build_responses_result` already sets both `msg.id` and `msg.response_metadata["id"]` from `response.id` on real API responses, so real traffic is unaffected.

Retry-on-invalid-id fallback deferred to a follow-up - it's a runtime behavior change that deserves its own tests, and ChatOpenAI doesn't do it either, so the risk profile stays identical between the two implementations for now.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Fixed - added `_identifying_params` returning `model`, `region_name`, `base_url`, `temperature`, `max_tokens`, `use_responses_api`. Combined with the existing `_get_ls_params`, LangSmith and callback consumers now get proper tracing metadata. `_get_invocation_params` is inherited from `BaseChatModel` - matches the ChatBedrockConverse pattern (also inherits it and only overrides `_get_ls_params`). Not overriding it explicitly since our tracing needs are covered by `_identifying_params` + `_get_ls_params`, and inheriting keeps behavior aligned with the parent contract.

libs/aws/tests/integration_tests/chat_models/test_bedrock_mantle.py
rishabhagrawal1: Thanks for the pointer. Added.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Fixed - no client-side restriction on `tool_choice` values any more. `bind_tools` accepts `"required"`, `"auto"`, `"none"`, tool-name strings, dict format, `True`, and `"any"`. 
`_prepare_responses_params` converts the Chat Completions dict format (`{"type": "function", "function": {"name": "..."}}`) to the Responses API format (`{"type": "function", "name": "..."}`), matching ChatOpenAI. Unit test `test_required_is_valid_responses_value` asserts `"required"` passes through.

One Mantle server-side quirk observed and documented: `openai.gpt-oss-*` models return 400 for `tool_choice="required"` on `/v1/responses` - the integration test class sets `has_tool_choice = False` for gpt-oss to reflect that model limitation, but it's not a client-side constraint.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Good catch - this is handled now. All four streaming methods wrap the SDK stream in a `with` / `async with` block, which auto-closes the underlying HTTP connection on normal exit, exception, or early consumer break. `bedrock.py:1345` uses an older explicit-close pattern; the `with`-block approach here is equivalent (and matches ChatOpenAI).

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Double-checked - not needed. Mantle exposes OpenAI-compatible endpoints, so the SDK yields a uniform event schema across all underlying models. `bedrock.py:1245` branches per-provider because `invoke_model` returns different JSON shapes per Anthropic/AI21/Cohere/Meta; that's not the situation here.

libs/aws/langchain_aws/utils.py
rishabhagrawal1: Was missed in git add. Thanks. Added now.

libs/aws/tests/integration_tests/chat_models/test_bedrock_mantle.py
rishabhagrawal1: Done.

libs/aws/langchain_aws/utils.py
rishabhagrawal1: Fixed - changed `self._token: Optional[str] = None` → `self._token: str = ""`. Removed all three `# type: ignore[return-value]` comments.

libs/aws/langchain_aws/chat_models/bedrock_mantle.py
rishabhagrawal1: Added following tests - 
- `test_default_includes_usage`: default sends `stream_options={"include_usage": True}`
- `test_explicit_true`: `stream_usage=True` → same
- `test_explicit_false_omits_stream_options`: `stream_usage=False` → not sent
- `test_chat_completions_final_usage_only_chunk_surfaced`: fakes a two-chunk stream ending in a usage-only chunk (`choices=[]`, populated `usage`) and asserts the final `AIMessageChunk` carries `usage_metadata` — regression guard against silently dropping the terminal usage chunk.

libs/aws/langchain_aws/utils.py
rishabhagrawal1: Both handled:
- Configurable via `bedrock_api_key_ttl_seconds` on `ChatBedrockMantle` (default 12h max; validated 1..43200).
- Not guaranteed: `_do_refresh` computes `effective_ttl = min(requested_ttl, credentials_seconds_remaining)` - if underlying STS creds have 1h left, the token is created with 1h expiry and the refresh clock reflects that. Documented the following: "The effective TTL is further capped by the underlying IAM credentials' remaining lifetime."

libs/aws/langchain_aws/utils.py
rishabhagrawal1: Kept at 10 min to match botocore's `_DEFAULT_MANDATORY_REFRESH_TIMEOUT` (default across all AWS creds).

rishabhagrawal1:
Known follow-ups (planned for a subsequent PR):

1. **Default API routing** - `use_responses_api=None` currently defaults to Responses API, but Many Mantle models are Chat-Completions-only (For ex. - qwen, deepseek, gpt-oss-safeguard, and others). Workaround: pass `use_responses_api=False` explicitly.

2. **Non-default base paths** - hardcoded `/v1`; frontier OpenAI (`gpt-5.x`), `google.gemma-4-*`, and `xai.grok-4.3` are served on `/openai/v1` and return 401 "Berm is not enabled" on `/v1`. Workaround: pass `base_url="https://bedrock-mantle.{region}.api.aws/openai/v1"` explicitly.

3. **`use_previous_response_id` failure recovery** - server-side state expiry, cross-region resume, or endpoint change causes a 404 that propagates. No auto-retry with full history. Workaround: leave `use_previous_response_id=False` (default) and send full history.

4. **Multimodal input** - `image_url`, `file`, `input_audio` content blocks are dropped by `_reduce_content_to_text` even though Mantle Chat Completions accepts them. Workaround: text-only until the multimodal support PR.

michaelnchin (commented):
Thanks for all of your work on this @rishabhagrawal1! Looks good from manual testing.

In regards to some of the follow ups mentioned - to smooth over the user experience, I think it would be good to automatically set the `openai/v1` and/or chat-completions API usage as default for the models/providers that specifically require them (especially for the latter, since it's mostly older models that aren't likely to pick up Responses API support).

Also, we should add some extra handling to exit out early for Claude models, which Mantle supports only through the [Messages API endpoint](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-messages-api.html). Currently, Mantle throws a confusing error (`the requested feature is not supported for this model`) when users try to use Claude with either of the OpenAI APIs. We should override this with a redirect to use ChatBedrockConverse for now instead.

Non-blocking though, these can both be taken up in a separate PR.

rishabhagrawal1:
Thanks @michaelnchin for checking. I will add your suggestions in the follow up PR as they are not blocker for now.

ccurme:
Thanks for all the work here. We went with https://github.com/langchain-ai/langchain-aws/pull/1189, which just sub-classes BaseChatOpenAI. @rishabhagrawal1 I've given you co-authorship on the commit https://github.com/langchain-ai/langchain-aws/commit/f7b06414fef81e472d279f9979eed486e128b988.

PR #1023, opened 2026-05-04
fix(aws): include cache reads and writes in input token counts

### LINKED ISSUE #1858: Issue: Wrong token calculation for Bedrock (ChatBedrockConverse) results in negative costs
### Issue you'd like to raise.

Langsmith assumes every provider will log `input_tokens` as the total prompt tokens whether they're cached or not. So it takes the provided `input_tokens`, then subtracts the `cache_read` tokens and then multiplies by the input price.

Bedrock logs `cache_read` and `cache_creation` tokens outside of the provided `input_tokens`, but Langsmith subtracts the cache reads from the provided input tokens, resulting in negative token counts, and therefore negative costs for very cache heavy runs. Additionally, it completely ignores `cache_creation` tokens.

So far I managed to find a workaround:

1. Create a `cache_creation` param in custom model pricing in langsmith
2. Set `cache_read` price to the real cache read price + the input_price, to cancel out the subtraction

Image

### Suggestion:

_No response_

ccurme:
For consistency with other providers.

Resolves https://github.com/langchain-ai/langsmith-sdk/issues/1858.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -1997,11 +1997,14 @@ def _extract_response_metadata(response: Dict[str, Any]) -> Dict[str, Any]:
 def _extract_usage_metadata(response: Dict[str, Any]) -> UsageMetadata:
     usage_dict = response.pop("usage")
 
-    input_tokens = usage_dict.get("inputTokens", 0)
+    bedrock_input_tokens = usage_dict.get("inputTokens", 0)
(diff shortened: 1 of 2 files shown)


sailtonight:
This shipped a silent semantic change to input_tokens — a field you know feeds billing — in a patch release, with a two-sentence description. Our system double-counted cache tokens and overcharged customers ~4.7x until we bisected to this
  commit.
  
  The fix being spec-correct doesn't excuse this. Redefining what a token-count field means is a breaking change, full stop. It should never reach downstream users via a routine patch bump. This is exactly the kind of mistake a library at this
  scale must not make.

Issue #766, opened 2025-11-19
DynamoDB `aput_writes` runs many writes sequentially


ltoniazzi:
### Package

langgraph-checkpoint-aws

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation with the integrated search
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code

```python
import random
from langgraph_checkpoint_aws import DynamoDBSaver
import asyncio
from time import time

TABLE_NAME = "test-langgraph" # Replace with your DynamoDB table name
REGION_NAME = "eu-central-1" # Replace with your AWS region

thread_id = str(random.randint(1, 1_000_000))

inputs_aput_writes = {
    "config": {
        "tags": [],
        "metadata": {
            "thread_id": thread_id,
            "user_id": "mock-id"
        },
        "callbacks": None,
        "recursion_limit": 25,
        "configurable": {
            "checkpoint_ns": "",
            "thread_id": thread_id,
            "user_id": "mock-id",
            "checkpoint_id": "1f0c530b-1ed7-6d1e-bfff-a40ab47d5463"
        }
    },
    "writes": [
        ('user_id', 'mock-id'),
        ('language', 'en'),
        ('thread_id', thread_id),
        ('timestamp', '251119_101538082238'),
        ('data', 'good data'),
        ('branch:to:parse_listing_qualifications', None)
    ],
    "task_id": "576d569d-52df-e0bf-c0ad-5606841615f8",
    "task_path": '~__pregel_pull, __start__'
}


dynamo_db_saver = DynamoDBSaver(
    table_name=TABLE_NAME,
    region_name=REGION_NAME,
)

async def main():
    t0 = time()
    await dynamo_db_saver.aput_writes(**inputs_aput_writes)

    print("Time to run aput_writes: {:.4f} seconds".format(time() - t0))
    
if __name__ == "__main__":

    asyncio.run(main())
```

### Error Message and Stack Trace (if applicable)

```shell
Time to run aput_writes: 1.0175 seconds
```

### Explanation
The async method `aput_writes` runs `put_writes` in `run_in_executor` ([here](https://github.com/langchain-ai/langchain-aws/blob/4062b71d3d213867046bb3c84958fb2ff279a75b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/saver.py#L357)) to be async.

[...]

PR #1029, opened 2026-05-09
perf: Improve performance of put_writes/aput_writes with DynamoDB backend

Kogia-sima:
## Summary

Replaces the per-item `put_item` loop in `UnifiedRepository.put_writes` with a single chunked `batch_write_item` call for metadata writes, collapsing N sequential round-trips into `N / 25` batched requests.

Related issue: #766

## Why

`put_writes` is on the hot path for `aput_writes`, which `langgraph` invokes once per task. The old implementation issued N synchronous `put_item` calls inside an `asyncio` thread-pool, so `aput_writes` have to wait for the event loop for `O(N)` round-trips even though the writes are independent.

For the reproduction in #766 (6 channel writes), this is ~1 s of latency that ought to be a single network call.

## Out of scope

Payload writes via `storage.store_data` (S3 + DynamoDB chunk table) are still sequential. Batching them requires a separate refactor inside `StorageStrategy` and is left for a follow-up.

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/unified_repository.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/unified_repository.py
@@ -22,6 +22,11 @@ else:
 
 logger = logging.getLogger(__name__)
 
+# DynamoDB ``batch_write_item`` per-request item cap.
+_BATCH_WRITE_MAX_ITEMS = 25
+# Maximum retries for ``UnprocessedItems`` returned by ``batch_write_item``.
+_BATCH_WRITE_MAX_RETRIES = 5
+
 
 # ========== Key Generation Functions ==========
 
@@ -587,7 +592,14 @@ class UnifiedRepository:
         task_id: str,
         task_path: str = "",
     ) -> None:
-        """Store writes with intelligent offloading and parallel execution.
+        """Store writes with intelligent offloading and batched metadata writes.
+
+        Metadata items for all writes are committed in a single
+        ``batch_write_item`` request (chunked at 25 items, the DynamoDB
+        per-request limit). Payloads are then stored via the storage
+        strategy. This preserves the metadata-before-payload invariant
+        while collapsing N sequential ``put_item`` calls into
+        ``ceil(N / 25)`` batch requests.
 
         Args:
             thread_id: Thread identifier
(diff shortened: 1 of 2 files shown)


kimnamu (commented):
Thanks for the #766 fix, @Kogia-sima — I'm not a maintainer, just another langchain-aws/Bedrock user who reviewed this and ran your branch (`27 passed`). The `UnprocessedItems` exponential-backoff retry is exactly the right DynamoDB pattern, and it's the piece the ThreadPool approach (#1093) lacks.

One behavior change I'd want the maintainers to weigh in on, with a repro:

**Dropping the conditional put changes write-once → overwrite-always for regular writes.** The old `put_writes` routed non-special writes through `put_single_write_item`, which sets `ConditionExpression="attribute_not_exists(PK)"` and silently swallows `ConditionalCheckFailedException` (write-once dedup on replay). `batch_write_item` can't carry that condition — I confirmed against the DynamoDB service model that `BatchWriteItem`'s `PutRequest` accepts only `Item` (no `ConditionExpression`; `PutItem` is the only one that takes it), and in moto the second write overwrites:
```
conditional put_item: 2nd write -> ConditionalCheckFailedException -> existing value preserved
batch_write_item:     2nd write -> overwrites (value = SECOND_REPLAY)
```
The docstring acknowledges this ("replays write identical data"). That holds for deterministic LangGraph writes, but it also introduces a **metadata/payload asymmetry**: metadata becomes overwrite-always while the payload `store_data` *keeps* `attribute_not_exists` (write-once-skip). On a replay where the serialized payload differs for the same `(task_id, idx)`, the metadata would update but the payload skip would leave stale bytes — the conditional was the safety net against exactly that skew.

Two reuse/scope notes (non-blocking):
- There's already a sibling `_batch_delete_metadata` (same file, ~L1014) doing chunked `batch_write_item` + `UnprocessedItems` retry. Your `_batch_put_write_items` is more robust (retry cap + backoff), but it's a second, slightly-divergent batch-retry helper — maintainers may prefer extracting one shared helper.
- `put_single_write_item` is now unused in the `put_writes` path (still defined). Worth noting in case it should be kept for the conditional semantics above.
- **vs #1093 (ThreadPool):** that PR parallelizes *both* legs (incl. the larger payload/S3 writes you leave sequential — `ceil(N/25)+N` round-trips here) and preserves the conditional put, but lacks your `UnprocessedItems` retry. The ideal would be batched metadata *with* write-once semantics, but `batch_write_item` precludes the condition.

Solid work on the retry handling. The conditional-put trade-off is the one thing I'd want a maintainer call on before merge. (Prepared with the help of an AI agent (Claude Code); I verified the repro and the service-model check myself before posting.)

PR #1032, opened 2026-05-10
fix(aws): support langchain-core standard format for `document` content blocks in ChatBedrockConverse

### LINKED ISSUE #1031: Standardize multimodal content block handling in `ChatBedrockConverse` to support langchain-core cross-provider format
### Package

langchain-aws

### Checked other resources

### Feature Description

 `ChatBedrockConverse` currently only accepts a `langchain-aws`-specific format using a `source` sub-dict, causing a `KeyError: 'source'` when the standard format is used.

I want this to  allow provider-agnostic code to work across `ChatOpenAI`, `ChatAnthropic`, `ChatBedrockConverse`, etc. without modification.

### Use Case

Adding multimodal message on  `ChatBedrockConverse` messages

```python
`_lc_content_to_bedrock` should accept both:

- The langchain-core standard format (`base64` + `mime_type` keys)
- The existing `langchain-aws`-specific `source` dict format (for backwards compatibility)

and convert both to the Bedrock Converse API wire format.

The same audit should be applied to other content block types (document, video, tool_result) to prevent recurrence.
```

### Additional Context

I would love to contribute

konbraphat51:
Solving #1031 

## Why

`_lc_content_to_bedrock` in `ChatBedrockConverse` accepted a proprietary `source`-dict format for multimodal content but did not handle the standard cross-provider format that langchain-core defines for `document` blocks:

```python
# Standard langchain-core format — previously raised KeyError: 'document'
{
    "type": "document",
    "base64": "<base64-encoded bytes>",
    "mime_type": "application/pdf",
    "name": "report.pdf",
}
```

## Changes
`_format_data_content_block` — added a "document" branch that mirrors the existing "file" branch: extracts base64 bytes, resolves the MIME type, and handles name inference from name, filename, metadata, or extras keys, emitting a warning and auto-generating a name when none is found.

`_lc_content_to_bedrock` — the document branch now checks for the "document" key first (existing Bedrock-format path, unchanged) and falls back to _format_data_content_block for standard-format blocks.


michaelnchin (commented):
Hi @konbraphat51, thanks for opening a pull request.


`_format_data_content_block` is specifically for handling conversion of LangChain v1 standard content blocks to Converse API format. `document` isn't a [valid multimodal standard format](https://docs.langchain.com/oss/javascript/langchain/messages#content-block-reference) - it's a type specific to Bedrock, so we shouldn't check for it in here. (I don't think it's even possible for `document` type to reach this function as they would fail the [explicit check](https://github.com/langchain-ai/langchain/blob/9545d058823b38421f943edd4ff1407e199fecb3/libs/core/langchain_core/messages/content.py#L919-L920) in `is_data_content_block`, but correct me if there's an actual bug here.)

Please use the equivalent `file` LangChain standard type, which we do have handling for (as you've duplicated the logic from). Alternatively, you can still pass `document` blocks but they have to be in [Converse API format](https://docs.aws.amazon.com/boto3/latest/reference/services/bedrock-runtime/client/converse.html#converse) instead.

PR #1045, opened 2026-05-15
fix(aws): use resolved base model for ChatBedrockConverse tracing

ccurme:
When ChatBedrockConverse is invoked through an Inference Profile, `model_id` is an opaque ARN that doesn't match LangSmith's price table.

Here we update the `ls_model_name` in `LangSmithParams` to prefer the resolved name.

We also make the same change in the `"model_name"` field in `AIMessage.response_metadata`, and in a check for Nova models in caching logic.

This would be released in a minor version bump (1.5).


MElkady:
This is an important improvement to fix the Amazon Bedrock application inference profiles issue. Looking forward to get it merged

michaelnchin (approved):
Thanks Chester! LGTM whenever you're ready for merge

PR #1046, opened 2026-05-16
fix(agentcore-codeinterpreter): preserve binary blob bytes and normalize ./ paths in readFiles

### LINKED ISSUE #1047: AgentCoreSandbox.download_files() silently corrupts binary files — double base64 decode + ./ path lookup failure in _extract_files_from_stream
### Submission checklist

- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox
- [x] Other / not sure / general

### Related Issues / PRs

None found.

```python

interpreter = CodeInterpreter(region="ap-northeast-2")
interpreter.start()

try:
    # Generate a binary file via shell (matplotlib PNG)
    backend.execute(
        "python3 -c \""
        "import matplotlib; matplotlib.use('Agg');"
        "import matplotlib.pyplot as plt;"
        "plt.plot([1,2,3]);"
        "plt.savefig('test.png')"
        "\""
    )

    # Bug 1: download_files returns corrupted bytes for binary files
    results = backend.download_files(["test.png"])
    content = results[0].content
    print(f"size   : {len(content)}")             # expected ~10000+, actual ~110
    print(f"valid  : {content[:4] == b'\\x89PNG'}")  # expected True, actual False

    # Bug 2: ./-prefixed path always returns file_not_found
    results2 = backend.download_files(["./test.png"])
    print(f"error  : {results2[0].error}")        # expected None, actual 'file_not_found'

finally:
    interpreter.stop()
```

```shell
No exception is raised — the corruption is silent.
download_files returns a FileDownloadResponse with content that appears valid (non-None)
but contains garbage bytes, making the bug easy to miss.

For Bug 2, the response is:
FileDownloadResponse(path='./test.png', content=None, error='file_not_found')
```

### Description

I'm trying to use `AgentCoreSandbox` as a deepagents backend to download binary files (e.g. matplotlib PNG plots) generated by the agent inside the sandbox, and upload them to Slack.

I expect `download_files(["plot.png"])` to return the original PNG bytes.

Instead, it returns either:
- **corrupted/truncated bytes** (~110 bytes) that are not valid PNG — even though the file is confirmed to be ~26 KB on the sandbox filesystem, or
- **`content=None` with `error='file_not_found'`** — even though the file exists — when the path has a `./` prefix (e.g. `"./data/plot.png"`).

There are two bugs in `_extract_files_from_stream` inside `langchain_agentcore_codeinterpreter/sandbox.py`:

### Bug 1 — `base64.b64decode` applied to blob that is already `bytes`

The AgentCore stream deserializes `resource["blob"]` as raw `bytes` (confirmed via debug inspection), but `_extract_files_from_stream` unconditionally calls `base64.b64decode()` on it — treating raw binary data as base64-encoded text.

```python
# Current code (buggy)
elif "blob" in resource:
    content = base64.b64decode(resource["blob"])
```

`base64.b64decode(validate=False)` silently ignores non-base64 bytes, so it produces ~110 bytes of garbage instead of the full 26 KB PNG.

**Debug evidence:**
```python
# resource["blob"] inspected at runtime:
type(resource["blob"])       # <class 'bytes'>
len(resource["blob"])        # 26340  ← correct file size
resource["blob"][:8].hex()  # '89504e470d0a1a0a'  ← valid PNG magic bytes

# After base64.b64decode():
len(content)                 # 110  ← corrupted
content[:4] == b'\x89PNG'   # False
```

### Bug 2 — `./`-prefixed paths never match the URI lookup

`_to_relative_path` only strips leading `/`, not `./`. As a result, paths like `"./data/plot.png"` are passed to `readFiles` and stored in `path_lookup` with the `./` prefix intact. The API returns URI `file:///data/plot.png`, which after processing becomes `"data/plot.png"` — this never matches `"./data/plot.png"` in the lookup, so `file_contents.get(path)` always returns `None`.

```
requested path  →  path_lookup key : "./data/plot.png"
API returns URI →  file_path key   : "data/plot.png"   ← mismatch → file_not_found
```

* I'm trying to use `AgentCoreSandbox.download_files()` to retrieve binary files (PNG plots) generated inside the sandbox.
* I expect `download_files(["plot.png"])` to return the original PNG bytes (~26 KB).
* Instead, it returns either corrupted truncated bytes (~110 bytes) due to double base64 decode, or `content=None` with `error='file_not_found'` when the path has a `./` prefix.

### Environment / System Info

```
OS: Amazon Linux 2 (x86_64)
Python: 3.13
deepagents: 0.6.1
langchain-agentcore-codeinterpreter: 0.0.2
bedrock-agentcore: 1.9.1
AWS region: ap-northeast-2
```

choyongho-bart0401:
## Summary

Fixes two bugs in `_extract_files_from_stream` / `_to_relative_path` that make `AgentCoreSandbox.download_files()` silently corrupt binary files and fail to find any path passed with a `./` prefix.

Related: langchain-ai/langchain-aws#1047

### Bug 1 — Double base64 decode of binary blobs

The AgentCore stream deserializes `resource["blob"]` as raw `bytes` before `_extract_files_from_stream` sees it. The code then unconditionally calls `base64.b64decode()`, treating valid binary content as base64-encoded text. With `validate=False` (the default), this silently drops non-base64 bytes and returns a small, corrupt result.

```python
# before
elif "blob" in resource:
    content = base64.b64decode(resource["blob"])  # blob is already bytes
```

Runtime evidence on a 26 KB PNG:

```
type(resource["blob"])       -> <class 'bytes'>
len(resource["blob"])        -> 26340
resource["blob"][:8].hex()  -> '89504e470d0a1a0a'   # valid PNG magic
len(base64.b64decode(resource["blob"])) -> 110      # corrupted output
```

### Bug 2 — `./` paths never match the URI-derived lookup key

`_to_relative_path` stripped leading `/` but not `./`. Worse, `_extract_files_from_stream` had its own inline `path.lstrip(\"/\")` when building `path_lookup`. The result: paths like `\"./data/foo.png\"` were stored in `path_lookup` with the `./` intact, while the URI returned by the API (`file:///data/foo.png`) normalized to `\"data/foo.png\"` — permanent key mismatch, permanent `file_not_found`.

## Fix

- Add a module-level `_normalize_relative_path` that strips both leading `/` and `./` prefixes.
- Use it in both `_to_relative_path` and `_extract_files_from_stream` (for `path_lookup` keys and the URI-derived `file_path`) so the two sides agree.
- In `_extract_files_from_stream`, only call `base64.b64decode` when the blob is not already `bytes`.

No public API changes.

## Tests

New unit tests (all 50 unit tests pass, `ruff check` and `ruff format --check` clean):

- `test_extract_files_blob_already_bytes` — raw-bytes blob is returned verbatim, not double-decoded.
- `test_extract_files_dot_slash_path` — `./`-prefixed requested path matches the URI-derived key.
- `test_download_files_dot_slash_path` — end-to-end: `download_files([\"./data/foo.png\"])` returns the original bytes with `error=None` and calls `readFiles` with the normalized `\"data/foo.png\"`.
- `test_relative_path_strips_dot_slash` — covers `./`, repeated `././`, and `/./` prefixes.

The existing `test_extract_files_blob_resource` (base64-encoded `str` path) still passes, confirming both blob shapes are handled.

## Checklist

- [x] Tests added for both bugs
- [x] `_extract_files_from_stream` handles both `bytes` and `str` blob
- [x] `_to_relative_path` handles `./`, `././`, and `/` prefixes
- [x] `path_lookup` and URI-derived path use the same normalization
- [x] No change to public API surface
- [x] `ruff check` / `ruff format --check` pass

### DIFF
--- a/libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
+++ b/libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
@@ -32,6 +32,21 @@ _AGENTCORE_EXECUTOR = ThreadPoolExecutor(
 )
 
 
+def _normalize_relative_path(path: str) -> str:
+    """Strip leading slashes and ``./`` prefixes to a canonical relative path.
+
+    Args:
+        path: File path (absolute or relative).
+
+    Returns:
+        Canonical relative path string with no leading ``/`` or ``./``.
+    """
+    path = path.lstrip("/")
+    while path.startswith("./"):
+        path = path[2:]
+    return path
+
+
 class SessionExpiredError(Exception):
     """Raised when the AgentCore session has expired or been terminated."""
 
@@ -104,8 +119,7 @@ def _extract_files_from_stream(
     """
     path_lookup: dict[str, str] = {}
     for path in requested_paths:
-        stripped = path.lstrip("/")
-        path_lookup[stripped] = path
+        path_lookup[_normalize_relative_path(path)] = path
 
     files: dict[str, bytes] = {}
 
@@ -117,13 +131,16 @@ def _extract_files_from_stream(
                 continue
             resource = item.get("resource", {})
             uri = resource.get("uri", "")
-            file_path = uri.replace("file://", "").lstrip("/")
+            file_path = _normalize_relative_path(uri.replace("file://", ""))
 
             content: bytes | None = None
             if "text" in resource:
                 content = resource["text"].encode("utf-8")
             elif "blob" in resource:
-                content = base64.b64decode(resource["blob"])
+                blob = resource["blob"]
+                # The AgentCore stream may deliver blob as already-decoded bytes.
+                # Only base64-decode when it arrives as encoded text.
+                content = blob if isinstance(blob, bytes) else base64.b64decode(blob)
 
             if content is not None:
                 original_path = path_lookup.get(file_path, file_path)
@@ -178,7 +195,7 @@ class AgentCoreSandbox(BaseSandbox):
 
     @staticmethod
     def _to_relative_path(path: str) -> str:
-        """Strip leading slashes so paths are relative for AgentCore APIs.
+        """Strip leading slashes and ``./`` prefixes for AgentCore APIs.
 
         Args:
             path: File path (absolute or relative).
@@ -186,7 +203,7 @@ class AgentCoreSandbox(BaseSandbox):
         Returns:
             Relative path string.
         """
-        return path.lstrip("/")
+        return _normalize_relative_path(path)
 
     def _invoke(self, method: str, params: dict[str, Any]) -> dict[str, Any]:
         """Invoke the interpreter and eagerly consume the response stream.
--- a/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
+++ b/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
@@ -228,6 +228,36 @@ def test_download_files_handles_session_expiry() -> None:
     assert results[0].error == "permission_denied"
 
 
+def test_download_files_dot_slash_path() -> None:
+    """./-prefixed paths must round-trip through readFiles and lookup."""
+    fake_png = b"\x89PNG\r\n\x1a\n" + b"\x00" * 100
+    sandbox, mock = _make_sandbox(
+        {
+            "stream": [
+                {
+                    "result": {
+                        "content": [
+                            {
+                                "type": "resource",
+                                "resource": {
+                                    "uri": "file:///data/foo.png",
+                                    "blob": fake_png,
+                                },
+                            }
+                        ]
+                    }
+                }
+            ]
+        }
+    )
+    results = sandbox.download_files(["./data/foo.png"])
+    mock.invoke.assert_called_once_with(
+        method="readFiles", params={"paths": ["data/foo.png"]}
+    )
+    assert results[0].error is None
+    assert results[0].content == fake_png
+
+
 # ------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


choyongho-bart0401:
Hi @michaelnchin, @sundargthb, @ccurme — sorry to ping you directly, and thank you in advance for your time.

Whenever it's convenient, I would be very grateful if one of you could take a look at this PR. It addresses two silent-corruption / lookup bugs in `AgentCoreSandbox.download_files()` that I ran into while building a deepagents flow:

1. `_extract_files_from_stream` double-base64-decodes binary blobs that arrive as raw `bytes`, so binary downloads (e.g. matplotlib PNGs) are silently truncated to ~110 bytes of garbage.
2. `./`-prefixed paths never match the URI-derived `path_lookup` key, so `download_files(["./foo.png"])` always returns `file_not_found` even when the file exists.

The full repro and root-cause analysis are in the linked issue (langchain-ai/langchain-aws#1047). All 50 existing unit tests still pass, new unit tests cover both bugs, and `ruff check` / `ruff format --check` are clean. No public API changes.

Happy to revise the patch in any direction you prefer — thank you again!

sundargthb (commented):
LGTM, thanks for the PR. Approved from my side but i am not a maintainer here. cc: @michaelnchin

michaelnchin (approved):
Thanks @choyongho-bart0401 !

PR #1042, opened 2026-05-15
feat(aws): Enable streaming for `prompt_prefill` structured output mode

michaelnchin:
Following up on #1026 - 

The `_extract_prompt_prefill_text` handler is redundant and can be removed:
1. For the list content conditional, ChatGeneration's [`set_text`](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/outputs/chat_generation.py#L44-L84) validator already joins text blocks into `.text` before being read by the parse
2. JsonOutputParser (and PydanticOutputParser) strips both ` ```json` and the stop backticks internally [inside of `parse_json_markdown`](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/utils/json.py#L154-L162).

With this part of the chain removed, streaming is unblocked, as we can now pass individual `AIMessageChunk`s directly to the output parser.

PR #1058, opened 2026-05-21
fix(aws): Fixed prompt prefill instructions to match Pydantic format

gabrielrfg:
Changed the prompt prefill instructions so smaller LLMs don't return "properties" in the final JSON, breaking output parsing.


gabrielrfg:
@michaelnchin Tagging you here as this was derived from the investigation of the streaming enablement on #1042

michaelnchin (approved):
Thanks @gabrielrfg !

PR #1073, opened 2026-05-30
fix(agentcore-codeinterpreter): fix write() path doubling and PermissionError when sandbox cwd != /

### LINKED ISSUE #1055: [langchain-agentcore-codeinterpreter] write_file creates files at doubled path (or PermissionError) when sandbox cwd != "/"
## Package

`langchain-agentcore-codeinterpreter`

## Checked other resources

## Description

`AgentCoreSandbox.write()` (inherited from `BaseSandbox`) has two internal consumers with **conflicting path requirements** when the sandbox working directory is not `/` (e.g., `/opt/amazon/genesis1p-tools/var/`):

| Consumer | Mechanism | Requires |
|---|---|---|
| `_write_preflight` | `execute()` → shell command | Real absolute filesystem path |
| `upload_files` | AgentCore `writeFiles` API | Path relative to sandbox cwd |

`_to_relative_path` strips only the leading `/`, which coincidentally satisfies both when `cwd == /`. When `cwd != /`, neither path form works:

**Case 1 — full absolute path** (e.g., `/opt/amazon/genesis1p-tools/var/workspace/hello.py`):
- Preflight: `makedirs('/opt/amazon/genesis1p-tools/var/workspace/')` ✓ succeeds
- Upload: `_to_relative_path` strips `/` → `opt/amazon/genesis1p-tools/var/workspace/hello.py`, `writeFiles` API resolves against cwd → **doubled path**: `/opt/amazon/genesis1p-tools/var/opt/amazon/genesis1p-tools/var/workspace/hello.py` ✗

**Case 2 — virtual/short path** (e.g., `/workspace/hello.py`):
- Upload: `_to_relative_path` strips `/` → `workspace/hello.py`, `writeFiles` API resolves against cwd → correct location ✓
- Preflight: `makedirs('/workspace')` → **`PermissionError: [Errno 13] Permission denied: '/workspace'`** ✗

The only case that accidentally works is a bare filename (no subdirectory), e.g. `/hello.py`, because the preflight's `makedirs('/')` is a no-op and the upload receives `hello.py` which resolves correctly under cwd.

## Example Code

```python

interpreter = CodeInterpreter(region="us-west-2")
interpreter.start()

# Sandbox cwd is /opt/amazon/genesis1p-tools/var/ (not /)
cwd = backend.execute("pwd").output.strip()
print(f"cwd: {cwd}")  # /opt/amazon/genesis1p-tools/var/

# Case 1: full absolute path → file created at DOUBLED path
result = backend.write(f"{cwd}/workspace/hello.py", "hello")
actual = backend.execute(f"find {cwd} -name 'hello.py'").output
print(actual)
# /opt/amazon/genesis1p-tools/var/opt/amazon/genesis1p-tools/var/workspace/hello.py  ← wrong

# Case 2: virtual path → PermissionError in preflight
result2 = backend.write("/workspace/hello.py", "hello")
print(result2.error)
# Error: PermissionError: [Errno 13] Permission denied: '/workspace'

interpreter.stop()
```

## Error Message and Stack Trace (if applicable)

**Case 1 — wrong path (no error raised, file silently created at wrong location):**
```
File created at: /opt/amazon/genesis1p-tools/var/opt/amazon/genesis1p-tools/var/workspace/hello.py
Expected:        /opt/amazon/genesis1p-tools/var/workspace/hello.py
```

**Case 2 — PermissionError:**
```
Traceback (most recent call last):
  File "<string>", line 8, in <module>
  File "<frozen os>", line 225, in makedirs
PermissionError: [Errno 13] Permission denied: '/workspace'
```

## Proposed Fix

`_to_relative_path` should strip the cwd prefix rather than just the leading `/`. The sandbox needs to know its cwd at construction time, and `write()` should normalize virtual paths to real absolute paths before the preflight shell command.

```python
class AgentCoreSandbox(BaseSandbox):
    def __init__(self, *, interpreter: CodeInterpreter, cwd: str) -> None:
        super().__init__(interpreter=interpreter)
        self._cwd = cwd.rstrip("/")

    def _to_relative_path(self, path: str) -> str:
        # Strip cwd prefix so writeFiles API receives a truly cwd-relative path
        cwd_prefix = self._cwd + "/"
        if path.startswith(cwd_prefix):
            return path[len(cwd_prefix):]
        return path.lstrip("/")

    def _to_cwd_absolute(self, path: str) -> str:
        """Resolve virtual paths (/workspace/f.py) to real absolute paths (/opt/.../workspace/f.py)."""
        if path.startswith(self._cwd + "/") or path == self._cwd:
            return path
        return self._cwd + "/" + path.lstrip("/")

    def write(self, file_path: str, content: str) -> WriteResult:
        # Normalize to real absolute path for the preflight shell command
        abs_path = self._to_cwd_absolute(file_path)
        preflight_error = self._write_preflight(abs_path)
        if preflight_error is not None:
            return preflight_error
        # upload_files → _to_relative_path strips cwd prefix correctly
        responses = self.upload_files([(abs_path, content.encode("utf-8"))])
        ...
```

Verified against `langchain-agentcore-codeinterpreter==0.0.3` (latest). The `_normalize_relative_path` refactor in 0.0.3 does not address this issue.

awg66:
## Summary

Fixes #1055 — `AgentCoreSandbox.write()` produces a doubled file path (or a `PermissionError`) when the sandbox working directory is not `/`.

- **Root cause**: `write()` had two consumers with conflicting path requirements. The preflight shell command (`makedirs`) needs a real absolute path, while the AgentCore `writeFiles` API needs a path relative to the sandbox cwd. The previous `_to_relative_path` only stripped the leading `/`, which happens to work when `cwd == /` but breaks otherwise.
- **Case 1 (doubled path)**: passing the full real absolute path (e.g. `/opt/amazon/.../workspace/hello.py`) caused `writeFiles` to prepend cwd again → doubled path.
- **Case 2 (PermissionError)**: passing a virtual path (e.g. `/workspace/hello.py`) caused the preflight `makedirs('/workspace')` to fail with a permission error.

## Changes

- Add optional `cwd: str | None = None` keyword argument to `AgentCoreSandbox.__init__` (fully backward-compatible — defaults to lazy detection via `pwd` on the first `write()` call).
- `_get_cwd()`: lazily executes `pwd` and caches the result.
- `_to_absolute_path(path)`: resolves virtual/relative paths to their real absolute location under the sandbox cwd, used by `write()` before the preflight shell command.
- `_to_relative_path(path)`: promoted from `@staticmethod` to instance method; strips the cwd prefix when known (preventing path doubling), otherwise falls back to the original leading-slash stripping (no behavior change for callers that don't go through `write()`).
- `write()` override: normalizes the path via `_to_absolute_path` before both `_write_preflight` and `upload_files`, and returns the resolved absolute path.
- 11 new unit tests covering both bug cases, lazy cwd detection, constructor injection, and `cwd="/"` backward compatibility.

- `_to_absolute_path` edge case: when `cwd` resolves to `""` (sandbox at `/`), the guard `if not cwd: return path` preserves the original behavior for all callers.
- `_to_relative_path` only strips the cwd prefix when `self._cwd is not None`; direct calls to `upload_files`/`download_files` that bypass `write()` do not trigger lazy cwd detection, preserving the existing call semantics for those methods.

## Test plan

- [x] All 62 unit tests pass (`uv run --group test pytest tests/unit_tests/`)
- [x] `ruff check` and `ruff format --check` pass
- [x] Verify Case 1 (doubled path) is fixed: `test_write_strips_cwd_prefix_from_upload_path`
- [x] Verify Case 2 (PermissionError) is fixed: `test_write_resolves_virtual_path_for_preflight`
- [x] Verify lazy cwd detection: `test_write_lazy_cwd_detection`
- [x] Verify backward compatibility: `test_write_root_cwd_preserves_existing_behavior`

> This PR was implemented with the assistance of an AI agent (Claude Sonnet 4.6 via Claude Code).

### DIFF
--- a/libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
+++ b/libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
@@ -289,12 +289,7 @@ class AgentCoreSandbox(BaseSandbox):
         if preflight_error is not None:
             return preflight_error
         responses = self.upload_files([(abs_path, content.encode("utf-8"))])
-        if not responses:
-            msg = (
-                f"Responses was expected to return 1 result, but it returned "
-                f"{len(responses)} with type {type(responses)}"
-            )
-            raise AssertionError(msg)
+        assert responses, "upload_files returned no responses"
         response = responses[0]
         if response.error:
             return WriteResult(
--- a/libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
+++ b/libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
@@ -167,6 +167,15 @@ class AgentCoreSandbox(BaseSandbox):
     The caller is responsible for managing the interpreter lifecycle
     (``start()`` / ``stop()``).
 
+    !!! note
+
+        When the sandbox working directory is not ``/`` (e.g.
+        ``/opt/amazon/genesis1p-tools/var/``), paths must be resolved
+        against the real cwd before shell preflight commands and stripped of
+        the cwd prefix before the AgentCore ``writeFiles``/``readFiles`` APIs.
+        Pass the known cwd via the ``cwd`` constructor argument, or let the
+        sandbox detect it automatically on the first ``write()`` call.
+
     Example:
         .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
```diff
@@ -185,26 +194,114 @@ class AgentCoreSandbox(BaseSandbox):
...
+        preflight_error = self._write_preflight(abs_path)
+        if preflight_error is not None:
+            return preflight_error
+        responses = self.upload_files([(abs_path, content.encode("utf-8"))])
+        if not responses:
+            msg = (
+                f"Responses was expected to return 1 result, but it returned "
+                f"{len(responses)} with type {type(responses)}"
+            )
+            raise AssertionError(msg)
```
michaelnchin: Nit: I don't really see the need for the extra responses details here if we know it's empty - this should work just as well:
```suggestion
        assert responses, "upload_files returned no responses"
```

michaelnchin: Actually, I think we can remove this check entirely, as it doesn't seem possible for `responses` to be empty:

We're always calling `upload_files` with one input tuple (regardless of whether `content` is empty as the second element gets encoded empty bytestring which is still valid), which will result in `responses` always being created with [exactly one `FileUploadResponse` element](https://github.com/awg66/langchain-aws/blob/fix/agentcore-write-path-normalization/libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py#L464).

I'll leave in that simplified assertion above though, just in case there's something else I'm missing.

libs/agentcore-codeinterpreter/langchain_agentcore_codeinterpreter/sandbox.py
```diff
@@ -185,26 +194,114 @@ class AgentCoreSandbox(BaseSandbox):
...
-    def _to_relative_path(path: str) -> str:
-        """Strip leading slashes and ``./`` prefixes for AgentCore APIs.
+    def _get_cwd(self) -> str:
+        """Return the sandbox working directory, detecting it lazily if needed.
+
+        Returns:
+            The working directory with any trailing slash stripped.
+        """
+        if self._cwd is None:
+            result = self.execute("pwd")
```
michaelnchin, line 224: If the executeCommand fails it'll return empty output, which will return cwd as an empty string -> _to_absolute_path thinks cwd is at root `/` and returns a possible incorrect path.

Should detect said executeCommand failures and throw an error instead:
```suggestion
            result = self.execute("pwd")
            if result.exit_code != 0 or not result.output.strip():
                raise RuntimeError(
                    f"Failed to detect sandbox working directory: "
                    f"exit_code={result.exit_code}, output={result.output!r}"
                )
```

michaelnchin (commented):
Thank you for the submission @awg66, looks good overall with a few minor comments.

Also, thanks for the additional callout here:
> direct calls to upload_files/download_files that bypass write() do not trigger lazy cwd detection

Will address the same for `download_files` in a separate PR

michaelnchin (approved):

PR #1076, opened 2026-06-03
fix(agentcore-codeinterpreter): resolve to correct download_files() path when sandbox cwd is not root

michaelnchin:
Follow up for https://github.com/langchain-ai/langchain-aws/pull/1073#pullrequestreview-4415377963

Applies the same `cwd`-aware path resolution from #1073 to `download_files()`, which had the same path-doubling bug on the read path. `download_files()` will now make one-time `pwd` call on first invocation if `cwd` hasn't been detected yet.


michaelnchin (commented):
Integ test passing: https://github.com/langchain-ai/langchain-aws/actions/runs/26913879830/job/79398734021

PR #1085, opened 2026-06-08
fix: preserve non-ASCII in tool, schema, and Bedrock-agent JSON serialization

HumphreySun98:
## Description

Several \`json.dumps\` call sites in \`langchain-aws\` build payloads that flow directly into model prompts, tool schemas returned to callers, or persisted trace logs. They all use Python's default \`ensure_ascii=True\`, so CJK / emoji / accented characters from end users land in those payloads as \`\\uXXXX\` escape sequences — unreadable in logs, and counted against token budgets at ~6× the raw cost.

Sites changed:

- \`chat_models/bedrock_converse.py\`
  - Tool-call → text conversion (\`[Called tool with parameters: {…}]\`) — sent verbatim to models without native tool support.
  - Prompt-prefill JSON schema dump (one occurrence had \`ensure_ascii=True\` set **explicitly** — flipped, see review note below).
  - AWS exception detail formatting.
  - \`toolResult\` JSON content blocks.
  - Two \`with_structured_output\` schema dumps.
- \`function_calling.py:_get_type\` — serializes tool parameter schemas, including any CJK descriptions / titles, into the Anthropic tool spec embedded in the system message.
- \`agents/utils.py\`
  - Bedrock agent stream \`returnControl\` event → \`response_text\` returned to the caller.
  - Accumulated \`trace_log\` for downstream logging.

This is the same convention already established in \`langchain-openai\`'s chat model and \`langchain-core\`'s [\`messages/utils.py:1810\`](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/messages/utils.py#L1810).

## Relevant issues

No tracking issue — surfaced via a repo-wide grep for \`json.dumps\` without \`ensure_ascii\` after the same audit ran across \`langchain-google\`. Same impact for end users: foreign-language input becomes \`\\uXXXX\` in storage / prompts.

## Changes

- \`libs/aws/langchain_aws/chat_models/bedrock_converse.py\`: add \`ensure_ascii=False\` to all six \`json.dumps\` sites (and flip one pre-existing explicit \`ensure_ascii=True\`).
- \`libs/aws/langchain_aws/function_calling.py\`: add \`ensure_ascii=False\` to the three \`_get_type\` dumps.
- \`libs/aws/langchain_aws/agents/utils.py\`: add \`ensure_ascii=False\` to the two \`parse_agent_response\` dumps.
- \`libs/aws/tests/unit_tests/test_ensure_ascii_in_serialization.py\`: a single static-source regression test per module — grep-scans the source for any \`json.dumps(\` call missing \`ensure_ascii=False\`. The Bedrock client mocking these modules require for end-to-end tests would make a behavioural assertion brittle; a static check survives that and any future refactor that keeps the call shape but flips the kwarg back.

## Testing

\`\`\`
$ python -m pytest libs/aws/tests/unit_tests/test_ensure_ascii_in_serialization.py -v
3 passed in 3.91s
\`\`\`

\`ruff check\` and \`ruff format --check\` pass on the modified files.

## Highlights for review

- **\`bedrock_converse.py:1991\` previously set \`ensure_ascii=True\` explicitly** on a prompt-prefill schema dump. I read this as intentional ASCII-only output (perhaps for a model variant that rejects high-bit chars in prompts). Flipping it to \`False\` matches the rest of the audit and what users hit on the other call sites, but if that explicit flag was load-bearing for some model behaviour, happy to revert just that one line — please let me know.
- The static test is intentionally module-scoped: extending it to the full package would catch \`json.dumps\` in unrelated paths (CLI tooling, helper scripts) that aren't user-facing.

_Disclaimer: this PR was prepared with the assistance of an AI agent (Claude Code). All code and test changes were reviewed by the author before submission._


libs/aws/tests/unit_tests/test_ensure_ascii_in_serialization.py
michaelnchin, line 1: These source-scanning tests don't really do much other than verifying that what we implemented is present. I think we would be better served by class-specific behavioral ones that actually test the execution through the through the modified paths with ASCII vs. non-ASCII characters

michaelnchin (commented):
Thanks @HumphreySun98! Checked and this looks consistent with most of the other provider integrations' (HuggingFace, Perplexity, Groq, Mistral, etc.), so I think this is a fair change.

libs/aws/tests/unit_tests/test_ensure_ascii_in_serialization.py
michaelnchin: Will remove this file and add new tests in a separate PR

michaelnchin (approved):
Integ CI passing on branch: https://github.com/langchain-ai/langchain-aws/actions/runs/27174707251

libs/aws/tests/unit_tests/test_ensure_ascii_in_serialization.py
HumphreySun98: Thanks for the merge, Michael — and totally fair point on the tests. The static scan was a hedge against mocking the Bedrock client surface for a behavioural assertion, but you're right that it doesn't actually exercise the modified paths.

If you haven't started the follow-up yet, happy to take it — feed CJK / non-ASCII into each path (`_messages_to_bedrock`, `_get_type`, `parse_agent_response`) and assert the raw bytes survive. Otherwise looking forward to seeing your version.

PR #1089, opened 2026-06-09
fix(agentcore-codeinterpreter): send binary uploads as raw blobs

### LINKED ISSUE #1083: langchain-agentcore-codeinterpreter: AgentCore writeFiles stores binary blobs as base64 text instead of decoded bytes
## Package

`langchain-agentcore-codeinterpreter`

## Checked other resources

## Description

When uploading binary (non-UTF-8) files via `AgentCoreSandbox.upload_files()`, the AgentCore `writeFiles` API receives the file content as a base64-encoded blob but writes the raw base64 string as text to disk instead of decoding it first.

As a result, binary files (e.g. images, PDFs, compiled artifacts) end up on the sandbox filesystem as base64-encoded text rather than their actual binary content. Any subsequent operation that reads the file (execution, download, etc.) will receive corrupted data.

The `upload_files` implementation in `sandbox.py` currently does:

```python
except UnicodeDecodeError:
    encoded = base64.b64encode(content).decode("ascii")
    file_list.append({"path": rel_path, "blob": encoded})
```

The `writeFiles` call reaches the AgentCore service with a base64 string in the `blob` field, but the service stores that string literally on disk instead of decoding it back to bytes.

## Example Code

```python

interpreter = CodeInterpreter(region="us-east-1")
interpreter.start()

# Upload a minimal PNG (binary, non-UTF-8)
png_bytes = (
    b"\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\x00\x00\x00\x01"
    b"\x00\x00\x00\x01\x08\x02\x00\x00\x00\x90wS\xde\x00\x00"
    b"\x00\x0cIDATx\x9cc\xf8\x0f\x00\x00\x01\x01\x00\x05\x18"
    b"\xd8N\x00\x00\x00\x00IEND\xaeB`\x82"
)
backend.upload_files([("/workspace/test.png", png_bytes)])

# Verify — this returns the base64 string instead of the decoded PNG bytes
result = backend.execute("xxd /workspace/test.png | head -2")
print(result.output)
# Expected: binary PNG header (89 50 4e 47 ...)
# Actual:   ASCII base64 characters (69 56 42 4f ...)
```

## Error Message and Stack Trace (if applicable)

No exception is raised. The failure is silent: `upload_files` returns success but the file on disk contains the raw base64 string instead of the decoded binary content.

## Workaround

We are currently working around this by subclassing `AgentCoreSandbox` and decoding files in-place via a follow-up `execute()` call after the upload:

```python
class BinaryAwareAgentCoreSandbox(AgentCoreSandbox):
    def upload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadResponse]:
        text_files, binary_files = [], []
        for path, content in files:
            try:
                content.decode("utf-8")
                text_files.append((path, content))
            except UnicodeDecodeError:
                binary_files.append((path, content))

        responses = list(super().upload_files(text_files)) if text_files else []

        if binary_files:
            bin_responses = super().upload_files(binary_files)
            if any(r.error for r in bin_responses):
                responses.extend(
                    FileUploadResponse(path=path, error="permission_denied")
                    for path, _ in binary_files
                )
            else:
                # Decode the base64 text that writeFiles wrote to disk back to bytes
                paths = "\n".join(f"  {self._to_absolute_path(path)!r}," for path, _ in binary_files)
                result = self.execute(
                    f"python3 << 'PYEOF'\n"
                    f"import base64\n"
                    f"for p in [\n{paths}\n]:\n"
                    f"    data = open(p, 'rb').read()\n"
                    f"    open(p, 'wb').write(base64.b64decode(data))\n"
                    f"PYEOF"
                )
                error = None if result.exit_code == 0 else "permission_denied"
                responses.extend(
                    FileUploadResponse(path=path, error=error)
                    for path, _ in binary_files
                )

        return responses
```

## Environment

- `langchain-agentcore-codeinterpreter` version: 0.0.4 (also present in 0.0.3)
- Python: 3.12

WOLIKIMCHENG:
Fixes #1083

## Why

`AgentCoreSandbox.upload_files()` sent non-UTF-8 content as a base64-encoded string in the `blob` field. AgentCore `writeFiles` stores that string literally, so binary files are corrupted on disk.

## What changed

- Send non-UTF-8 upload content as raw `blob` bytes.
- Keep UTF-8 content on the existing `text` path.
- Add unit coverage for raw binary blobs, mixed text/binary uploads, and async upload delegation.
- Tighten the binary integration test expectation to require byte-for-byte roundtrip.

## Testing

- `UV_CACHE_DIR=/tmp/openclaw-uv-cache-langchain-aws make test TEST_FILE=tests/unit_tests/test_sandbox.py`
- `UV_CACHE_DIR=/tmp/openclaw-uv-cache-langchain-aws uv lock --check`
- `UV_CACHE_DIR=/tmp/openclaw-uv-cache-langchain-aws uv run --no-sync pytest -m compile tests/integration_tests`
- `git diff --check`

### DIFF
--- a/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
+++ b/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
@@ -7,7 +7,6 @@ AWS credential requirements.
 from __future__ import annotations
 
 import asyncio
-import base64
 from collections.abc import Iterator
 from typing import Any
 from unittest.mock import MagicMock
--- a/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
+++ b/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
@@ -145,7 +145,6 @@ def test_upload_files_binary_uses_blob() -> None:
     """Non-UTF-8 content should be uploaded as raw blob bytes."""
     sandbox, mock = _make_sandbox()
     binary_content = b"\x80\x81\x82"
-    base64_content = base64.b64encode(binary_content)
 
     sandbox.upload_files([("/data.bin", binary_content)])
 
@@ -153,9 +152,6 @@ def test_upload_files_binary_uses_blob() -> None:
     assert "blob" in content
     assert "text" not in content
     assert content["blob"] == binary_content
-    assert isinstance(content["blob"], bytes)
-    assert content["blob"] != base64_content
-    assert content["blob"] != base64_content.decode("ascii")
 
 
 def test_upload_files_mixed_text_and_binary() -> None:
@@ -762,21 +758,6 @@ def test_aupload_files_uses_dedicated_executor() -> None:
     result = asyncio.run(sandbox.aupload_files([("/test.txt", b"data")]))
     assert result[0].error is None
     assert any("agentcore-sandbox" in name for name in thread_names)
-
-
-def test_aupload_files_binary_uses_raw_blob() -> None:
-    """aupload_files() should preserve raw blob bytes for binary uploads."""
-    sandbox, mock = _make_sandbox()
-    binary_content = b"\x80\x81\x82"
-    base64_content = base64.b64encode(binary_content)
-
-    result = asyncio.run(sandbox.aupload_files([("/data.bin", binary_content)]))
-
-    content = mock.invoke.call_args.kwargs["params"]["content"][0]
-    assert result[0].error is None
-    assert content == {"path": "data.bin", "blob": binary_content}
-    assert content["blob"] != base64_content
-    assert content["blob"] != base64_content.decode("ascii")
 
 
 def test_adownload_files_uses_dedicated_executor() -> None:
--- a/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
+++ b/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
@@ -7,6 +7,7 @@ AWS credential requirements.
 from __future__ import annotations
 
 import asyncio
+import base64
 from collections.abc import Iterator
 from typing import Any
 from unittest.mock import MagicMock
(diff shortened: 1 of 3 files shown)


libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
```diff
@@ -141,12 +142,41 @@ def test_upload_files_text() -> None:
...
+
+    sandbox.upload_files([("/data.bin", binary_content)])
+
     content = mock.invoke.call_args.kwargs["params"]["content"][0]
     assert "blob" in content
     assert "text" not in content
+    assert content["blob"] == binary_content
+    assert isinstance(content["blob"], bytes)
+    assert content["blob"] != base64_content
+    assert content["blob"] != base64_content.decode("ascii")
```
michaelnchin: Last 3 assertions seem redundant given that `assert content["blob"] == binary_content` would already confirm that's it's bytes/binary type

libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
```diff
@@ -734,6 +764,21 @@ def tracking_invoke(**kwargs: Any) -> dict[str, Any]:
     assert any("agentcore-sandbox" in name for name in thread_names)
 
 
+def test_aupload_files_binary_uses_raw_blob() -> None:
```
michaelnchin: This test seems redundant too, `test_aupload_files_uses_dedicated_executor` checks that `aupload` files delegates the `upload_files` calls to threads, and the other `test_upload_files`  the rest of the path after the delegation. So I think we can remove this

michaelnchin (commented):
Thanks @WOLIKIMCHENG ! Fix looks good, couple comments about the tests -

michaelnchin (approved):

PR #1093, opened 2026-06-11
perf(langgraph-checkpoint-aws): parallelize DynamoDB `put_writes`

### LINKED ISSUE #766: DynamoDB `aput_writes` runs many writes sequentially

HumphreySun98:
## Description

\`UnifiedRepository.put_writes\` issues N×2 **sequential** calls for N graph-state updates: a conditional \`put_item\` (metadata) plus a chunk-store write (payload) per update, one after another. \`aput_writes\` runs the whole method in an executor, so an async caller with N pending writes pays N×2 round trips of serialized latency — defeating DynamoDB's write concurrency, as reported in #766 (which compares the Postgres saver's concurrent \`executemany\`).

This PR runs the per-write *(metadata, payload)* pairs on a \`ThreadPoolExecutor\`:

- Each pair still executes **in order on its own worker thread**, so the crash-consistency invariant — *no payload chunk without its metadata row* — is preserved per write.
- boto3 clients are thread-safe; concurrency is capped at **10** to match boto3's default \`max_pool_connections\`.
- A single write runs inline (no pool overhead).
- The first failure propagates to the caller, matching the sequential behavior of raising on error (earlier writes could already be persisted in both versions).
- \`BatchWriteItem\` was **not** an option: the metadata put uses \`attribute_not_exists\` conditions, which batch writes don't support. \`TransactWriteItems\` supports conditions but changes semantics (all-or-nothing) and costs 2× WCU — out of scope here.

## Relevant issues

Fixes #766

## Highlights for review — intentional test-semantics change

\`TestMetadataBeforePayloadOrdering\` pinned the *sequential* implementation, including a test literally named \`test_put_writes_sequential_not_parallel\`. I read the load-bearing invariant as **per-write** metadata-before-payload (which concurrency preserves), not global serialization — but if global ordering was intentional (e.g., for a crash-recovery scan that assumes prefix-completeness across writes), please say so and I'll rework this to batched fan-out with a barrier between metadata and payload phases instead.

Two of the old trackers were also racy under any concurrent implementation (a non-thread-safe shared call counter, and a global alternating-order assertion), so the rewrites double as de-flaking:

- \`test_put_writes_sequential_not_parallel\` → **\`test_put_writes_executes_concurrently\`**: a \`threading.Barrier\` sized to the write count proves concurrency *behaviorally* — every \`put_item\` blocks until all writes have entered the call; a sequential implementation deadlocks on the first write and fails via barrier timeout. (No timing sleeps, deterministic both ways.)
- \`test_put_writes_metadata_before_payload_per_write\`: now asserts the per-write invariant via per-thread call ordering — each worker thread must show strict \`put_item\` → \`store_data\` alternation.
- \`test_put_writes_failure_stops_at_failed_write\` → **\`test_put_writes_failed_write_never_stores_its_payload\`**: failure injection keyed on the write's channel (deterministic) instead of a shared counter (racy); asserts the failed write's payload chunk (idx-suffixed key) is never stored and the error propagates.
- New: single-write inline path; concurrent error propagation.

## Testing

\`\`\`
$ python -m pytest tests/unit_tests/checkpoint/dynamodb/test_unified_repository.py -q
28 passed in 0.12s   (3 consecutive runs, no flakes)
\`\`\`

Reverting only the \`unified_repository.py\` change while keeping the new tests makes \`test_put_writes_executes_concurrently\` fail deterministically via barrier timeout (~10s), confirming the test pins the sequential bottleneck. \`ruff check\` and \`ruff format --check\` pass.

_Disclaimer: this PR was prepared with the assistance of an AI agent (Claude Code). All code and test changes were reviewed by the author before submission._

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/unified_repository.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/unified_repository.py
@@ -3,6 +3,7 @@
 import logging
 import time
 from collections.abc import Callable, Iterator, Sequence
+from concurrent.futures import ThreadPoolExecutor
 from typing import TYPE_CHECKING, Any, cast
 
 from botocore.exceptions import ClientError
@@ -607,15 +608,20 @@ class UnifiedRepository:
         all_special = all(w[0] in WRITES_IDX_MAP for w in writes)
 
         # Build all write items with payload storage
-        for base_item, ref_item in self._build_write_items(
-            thread_id,
-            checkpoint_ns,
-            checkpoint_id,
-            writes,
-            task_id,
-            task_path,
-            allow_overwrites=all_special,
-        ):
+        write_items = list(
+            self._build_write_items(
+                thread_id,
+                checkpoint_ns,
+                checkpoint_id,
+                writes,
+                task_id,
+                task_path,
+                allow_overwrites=all_special,
+            )
+        )
+
+        def _store_write_item(item_pair: tuple[dict[str, Any], dict[str, Any]]) -> None:
+            base_item, ref_item = item_pair
             # Store metadata to base table
             self.put_single_write_item(
                 checkpoint_id, base_item, allow_overwrites=all_special
@@ -628,6 +634,26 @@ class UnifiedRepository:
                 serialized_data=ref_item["value_data"],
                 allow_overwrite=all_special,
             )
+
+        if len(write_items) == 1:
+            _store_write_item(write_items[0])
+            return
+
+        # Each write targets an independent item, so the per-item conditional
+        # put (`attribute_not_exists`) semantics are unaffected by ordering.
+        # Concurrency is capped at boto3's default connection-pool size to
+        # avoid exhausting the shared client's urllib3 pool.
+        max_workers = min(len(write_items), 10)
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
+            futures = [
+                executor.submit(_store_write_item, item_pair)
+                for item_pair in write_items
+            ]
+            for future in futures:
+                # Propagate the first failure (matches the sequential
+                # behavior of raising on error; earlier writes may have
+                # already been persisted in both versions).
+                future.result()
 
     def _build_write_items(
         self,
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/unified_repository.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/unified_repository.py
@@ -641,8 +641,11 @@ class UnifiedRepository:
 
         # Each write targets an independent item, so the per-item conditional
         # put (`attribute_not_exists`) semantics are unaffected by ordering.
-        # Concurrency is capped at boto3's default connection-pool size to
-        # avoid exhausting the shared client's urllib3 pool.
+        # Concurrency is capped at boto3's *default* connection-pool size
+        # (`max_pool_connections=10`) to avoid exhausting the shared client's
+        # urllib3 pool. This is intentionally the default rather than the
+        # client's configured value, since the configured pool size isn't
+        # introspectable from the client here; 10 is a safe conservative cap.
         max_workers = min(len(write_items), 10)
         with ThreadPoolExecutor(max_workers=max_workers) as executor:
             futures = [
--- a/libs/langgraph-checkpoint-aws/tests/unit_tests/checkpoint/dynamodb/test_unified_repository.py
+++ b/libs/langgraph-checkpoint-aws/tests/unit_tests/checkpoint/dynamodb/test_unified_repository.py
@@ -528,6 +528,114 @@ class TestPendingWrites:
                 task_id="task1",
             )
 
+    def test_put_writes_executes_concurrently(self):
+        """Multiple writes must run concurrently, not sequentially.
+
+        `aput_writes` delegates to `put_writes` via `run_in_executor`; with N
+        graph-state updates the old implementation issued N*2 sequential
+        `put_item` calls, serializing what DynamoDB happily accepts in
+        parallel (#766).
+
+        A `threading.Barrier` sized to the number of writes proves
+        concurrency behaviorally: every `put_item` blocks until all writes
+        have entered the call. A sequential implementation deadlocks on the
+        first write and trips the barrier timeout, failing the test.
+        """
+        import threading
+
+        num_writes = 3
+        barrier = threading.Barrier(num_writes, timeout=10)
+
+        mock_client = Mock()
+        mock_serializer = Mock()
+        mock_storage = Mock()
+        mock_serializer.serialize.return_value = ("msgpack", b"data", False)
+        mock_storage.store_data.return_value = "DYNAMODB"
+
+        def blocking_put_item(**kwargs):
+            # Passes only if all writes arrive here concurrently.
+            barrier.wait()
+            return {}
+
+        mock_client.put_item.side_effect = blocking_put_item
+
+        repo = UnifiedRepository(
+            dynamodb_client=mock_client,
+            table_name=TEST_TABLE_NAME,
+            serializer=mock_serializer,
+            storage_strategy=mock_storage,
+        )
+
+        writes = [(f"channel{i}", {"value": i}) for i in range(num_writes)]
+        repo.put_writes(
+            TEST_THREAD_ID,
+            TEST_CHECKPOINT_NS,
+            TEST_CHECKPOINT_ID,
+            writes,
+            task_id="task1",
+        )
+
+        assert mock_client.put_item.call_count == num_writes
+        assert mock_storage.store_data.call_count == num_writes
+
+    def test_put_writes_single_write_runs_inline(self):
+        """A single write must not pay thread-pool overhead and must still
+        store both the base item and the chunk payload."""
+        mock_client = Mock()
+        mock_serializer = Mock()
+        mock_storage = Mock()
+        mock_serializer.serialize.return_value = ("msgpack", b"data", False)
+        mock_storage.store_data.return_value = "DYNAMODB"
+
+        repo = UnifiedRepository(
+            dynamodb_client=mock_client,
+            table_name=TEST_TABLE_NAME,
+            serializer=mock_serializer,
+            storage_strategy=mock_storage,
+        )
+
+        repo.put_writes(
+            TEST_THREAD_ID,
+            TEST_CHECKPOINT_NS,
+            TEST_CHECKPOINT_ID,
+            [("channel1", {"value": 1})],
+            task_id="task1",
+        )
+
+        assert mock_client.put_item.call_count == 1
+        assert mock_storage.store_data.call_count == 1
+
+    def test_put_writes_concurrent_error_propagates(self):
+        """A failure in any concurrent write must propagate to the caller,
+        matching the sequential behavior of raising on error."""
+        mock_client = Mock()
+        mock_serializer = Mock()
+        mock_storage = Mock()
+        mock_serializer.serialize.return_value = ("msgpack", b"data", False)
+        mock_storage.store_data.return_value = "DYNAMODB"
+
+        mock_client.put_item.side_effect = ClientError(
+            {"Error": {"Code": "ProvisionedThroughputExceededException"}},
+            "PutItem",
+        )
+
+        repo = UnifiedRepository(
+            dynamodb_client=mock_client,
+            table_name=TEST_TABLE_NAME,
+            serializer=mock_serializer,
+            storage_strategy=mock_storage,
+        )
+
+        writes = [(f"channel{i}", {"value": i}) for i in range(3)]
+        with pytest.raises(ClientError):
+            repo.put_writes(
+                TEST_THREAD_ID,
+                TEST_CHECKPOINT_NS,
+                TEST_CHECKPOINT_ID,
+                writes,
+                task_id="task1",
+            )
+
 
 class TestDeleteOperations:
     """Tests for delete operations."""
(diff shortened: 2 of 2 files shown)


kimnamu (commented):
Thanks for tackling #766, @HumphreySun98 — I'm not a maintainer, just another langchain-aws/Bedrock user who reviewed this carefully and ran your branch locally (`28 passed`, no flakes).

The design is the right call on two points I verified:

1. **Conditional-put semantics preserved.** You keep `put_single_write_item`, so the `attribute_not_exists(PK)` write-once guard on non-special writes stays intact. I confirmed in moto that this matters: a replay of a regular write raises `ConditionalCheckFailedException` (which the existing code silently swallows → existing value preserved), i.e. write-once dedup. The competing batch approach (#1029) can't carry a `ConditionExpression` — I checked the DynamoDB service model and `BatchWriteItem`'s `PutRequest` accepts only `Item` (no `ConditionExpression`), whereas `PutItem` does — so it silently switches to last-writer-wins. Your approach has no such behavior change.
2. **Both legs parallelized.** You fan out the full `(metadata, payload)` pair, so the payload `store_data` (the larger write, possibly S3) is parallelized too — not just metadata.

The `threading.Barrier` test is a clean behavioral proof of concurrency (it deadlocks→times-out on a sequential impl), deterministic in both directions. Nice.

Two non-blocking notes for the maintainers' decision (just flagging, not requesting changes):
- **vs #1029 (batch):** #1029 collapses the *metadata* leg into `ceil(N/25)` calls and adds `UnprocessedItems` retry, but leaves payloads sequential and drops the conditional put. Your PR keeps semantics and parallelizes both legs. They're partly complementary, but batched metadata *with* conditional semantics isn't possible via `batch_write_item` (the condition can't ride along), so it's "fewer RPCs but overwrite-always" vs "more RPCs but write-once + payload parallelism."
- **cap=10 / `max_pool_connections`:** matching the default boto3 pool (10) is sound. If a user passes a client with a larger `max_pool_connections`, the cap stays at 10 — fine as a conservative default; might be worth a one-line comment noting the cap is intentionally the *default* pool size, not the client's configured value.

Clean, well-scoped, semantics-preserving — I'd love to see it land. (This review was prepared with the help of an AI agent (Claude Code); I ran the branch and verified the claims, including the moto repro and the service-model check, myself before posting.)

HumphreySun98:
Thanks for the unusually thorough review @kimnamu — especially running the moto repro for the conditional-put semantics and checking the `BatchWriteItem` service model. That's exactly the right framing of the tradeoff vs #1029.

I've addressed your one actionable note in dea3b0a: added a comment clarifying that `max_workers=10` is intentionally boto3's *default* `max_pool_connections`, not the client's configured value (which isn't introspectable from the client object here), so it stays a conservative cap.

To restate the comparison for whoever picks this up, since #1029 is the obvious alternative:

- **#1029 (batch):** collapses the metadata leg into `ceil(N/25)` `BatchWriteItem` calls + `UnprocessedItems` retry — but `BatchWriteItem`'s `PutRequest` can't carry a `ConditionExpression`, so it drops the `attribute_not_exists(PK)` write-once dedup (silently becomes last-writer-wins), and it leaves the payload `store_data` leg sequential.
- **This PR:** keeps the conditional-put write-once semantics unchanged and parallelizes *both* legs (including the larger payload/S3 write). More RPCs than batch, but no behavior change.

The two aren't mutually exclusive — batched metadata *with* write-once semantics just isn't possible through `batch_write_item` — so I'd lean toward landing the semantics-preserving version, but happy to defer to maintainer preference. CI is green and the branch is up to date.

HumphreySun98:
@michaelnchin — friendly nudge on this one when you have a moment. It fixes #766 (the `aput_writes` sequential-writes issue you triaged), @kimnamu did a thorough independent review (ran the moto repro, verified the conditional-put semantics and the `BatchWriteItem`-can't-carry-a-condition tradeoff vs #1029), and CI is green. It's a one-function change that keeps the write-once semantics while parallelizing both the metadata and payload legs. Happy to adjust if you'd prefer a different shape.

michaelnchin (commented):
Hi @HumphreySun98, can you post a few screenshots/videos of you running this change against DDB with the implemented change vs. the current build, to demonstrate that the fix is working?

HumphreySun98:
Thanks @michaelnchin — here's the change vs current `main` run against **DynamoDB Local** (real DDB engine), 10 writes, 3 trials each:

| build | trial 1 | trial 2 | trial 3 | median |
|---|---|---|---|---|
| current `main` (sequential) | 450.2 ms | 333.4 ms | 317.9 ms | **~333 ms** |
| this PR (concurrent) | 62.4 ms | 56.4 ms | 52.2 ms | **~56 ms** |

**~5.9× faster**, and correctness is identical: **20 rows persisted** both ways (10 metadata + 10 payload chunks), and the write-once conditional put still dedups — re-running the identical writes doesn't duplicate (`rows after identical re-run: 20`). Screenshot of the full run below.

One note for transparency: DynamoDB Local runs on localhost with ~0 ms RTT, which hides the network latency the fix actually targets, so I inject a fixed **12 ms/call** to represent a typical AWS same-region round-trip — it changes only timing, not behavior. The repro is a ~90-line `demo_put_writes.py` + `run_demo.sh` (starts the container, runs current vs this branch); happy to attach both, or to run it against a real throwaway AWS table (no injected latency) if you'd prefer numbers straight from AWS.

<details>
<summary>Repro scripts (<code>demo_put_writes.py</code> + <code>run_demo.sh</code>) — needs Docker</summary>

Run against a fresh clone with both refs:
```bash
git clone https://github.com/HumphreySun98/langchain-aws.git && cd langchain-aws
git remote add upstream https://github.com/langchain-ai/langchain-aws.git && git fetch upstream main
git checkout perf/dynamodb-put-writes-concurrent
pip install -e libs/langgraph-checkpoint-aws langchain-core
# save the two files below into the repo root, then:
./run_demo.sh
```

`demo_put_writes.py`
```python
"""Demo: DynamoDBSaver.put_writes against real DynamoDB Local.

Shows the parallelize-put_writes change (#1093) working end-to-end:
  * correctness: all writes land; write-once conditional put still holds,
  * latency: N writes go from ~2N sequential round-trips to concurrent.

Localhost DynamoDB Local has ~0ms RTT, which hides the network cost the
fix targets, so we inject a fixed per-call latency to represent a typical
AWS same-region round-trip (clearly labelled). This changes only timing,
not behaviour.

Run against a DynamoDB Local container on :8000. Reports which code path
(sequential vs concurrent) is active by inspecting the source.
"""

import time
import uuid

import boto3

from langgraph_checkpoint_aws.checkpoint.dynamodb.saver import DynamoDBSaver

ENDPOINT = "http://localhost:8000"
REGION = "us-east-1"
TABLE = "lc_demo_checkpoints"
RTT_MS = 12  # representative AWS same-region round-trip; localhost has ~0
N_WRITES = 10

# dummy creds for DynamoDB Local
import os

os.environ.setdefault("AWS_ACCESS_KEY_ID", "dummy")
os.environ.setdefault("AWS_SECRET_ACCESS_KEY", "dummy")


def ensure_table():
    ddb = boto3.client("dynamodb", region_name=REGION, endpoint_url=ENDPOINT)
    existing = ddb.list_tables()["TableNames"]
    if TABLE in existing:
        ddb.delete_table(TableName=TABLE)
        ddb.get_waiter("table_not_exists").wait(TableName=TABLE)
    ddb.create_table(
        TableName=TABLE,
        KeySchema=[
            {"AttributeName": "PK", "KeyType": "HASH"},
            {"AttributeName": "SK", "KeyType": "RANGE"},
        ],
        AttributeDefinitions=[
            {"AttributeName": "PK", "AttributeType": "S"},
            {"AttributeName": "SK", "AttributeType": "S"},
        ],
        BillingMode="PAY_PER_REQUEST",
    )
    ddb.get_waiter("table_exists").wait(TableName=TABLE)
    return ddb


def which_impl():
    import inspect

    from langgraph_checkpoint_aws.checkpoint.dynamodb import unified_repository

    src = inspect.getsource(unified_repository.UnifiedRepository.put_writes)
    return "CONCURRENT (fix)" if "ThreadPoolExecutor" in src else "SEQUENTIAL (baseline)"


def inject_latency(saver):
    real_put = saver.client.put_item

    def slow_put(**kw):
        time.sleep(RTT_MS / 1000)
        return real_put(**kw)

    saver.client.put_item = slow_put


def main():
    ddb = ensure_table()
    saver = DynamoDBSaver(table_name=TABLE, region_name=REGION, endpoint_url=ENDPOINT)
    inject_latency(saver)

    thread_id = f"thread-{uuid.uuid4()}"
    config = {
        "configurable": {
            "thread_id": thread_id,
            "checkpoint_ns": "",
            "checkpoint_id": "1ef00000-0000-6000-8000-000000000000",
        }
    }
    # N distinct channel writes (non-special -> conditional write-once put)
    writes = [(f"channel_{i}", {"value": i, "text": f"payload-{i}"}) for i in range(N_WRITES)]

    impl = which_impl()
    print(f"impl: {impl}   writes: {N_WRITES}   injected RTT: {RTT_MS}ms/call")

    t = time.perf_counter()
    saver.put_writes(config, writes, task_id="task-1")
    elapsed = (time.perf_counter() - t) * 1000
    print(f"put_writes wall-clock: {elapsed:7.1f} ms")

    # correctness: count persisted rows for this thread
    scan = ddb.scan(TableName=TABLE)["Items"]
    pk_rows = [it for it in scan if thread_id in it.get("PK", {}).get("S", "")]
    print(f"rows persisted for thread: {len(pk_rows)}")

    # write-once: re-run identical writes; conditional put must NOT duplicate
    saver.put_writes(config, writes, task_id="task-1")
    scan2 = ddb.scan(TableName=TABLE)["Items"]
    pk_rows2 = [it for it in scan2 if thread_id in it.get("PK", {}).get("S", "")]
    dedup_ok = len(pk_rows2) == len(pk_rows)
    print(f"rows after identical re-run: {len(pk_rows2)}  (write-once dedup {'OK' if dedup_ok else 'BROKEN'})")


if __name__ == "__main__":
    main()

```

`run_demo.sh`
```bash
#!/usr/bin/env bash
# Reproducible before/after demo of #1093 against real DynamoDB Local.
#   1. starts DynamoDB Local (docker) on :8000
#   2. runs demo_put_writes.py on the CURRENT build (sequential) and the FIX
#      (concurrent), 3 trials each, against the same real DDB engine.
set -euo pipefail
cd "$(dirname "$0")"

REPO_FILE=libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/dynamodb/unified_repository.py

echo "### DynamoDB Local"
docker rm -f ddblocal >/dev/null 2>&1 || true
docker run -d --name ddblocal -p 8000:8000 amazon/dynamodb-local >/dev/null
sleep 3

run_side () {
  local label=$1 ref=$2
  git checkout -q "$ref" -- "$REPO_FILE"
  pip install -e libs/langgraph-checkpoint-aws -q --no-deps >/dev/null 2>&1
  echo "=================  $label  ================="
  for i in 1 2 3; do python demo_put_writes.py | grep -E "impl|wall-clock|rows|dedup"; echo "--"; done
}

run_side "CURRENT BUILD (sequential)" upstream/main
run_side "THIS PR (concurrent)"       perf/dynamodb-put-writes-concurrent

echo "### done — DynamoDB Local left running (docker rm -f ddblocal to stop)"

```
</details>

HumphreySun98:
> Hi Humphrey (@HumphreySun98), can you post a few screenshots/videos of you running this change against DDB with the implemented change vs. the current build, to demonstrate that the fix is working?

demo_screenshot
here is the demo screenshot

PR #1098, opened 2026-06-12
feat(aws): surface per-TTL cache breakdown from cacheDetails

onraz:
## Summary

 Adds per-TTL cache token breakdown to `input_token_details` when Bedrock returns `cacheDetails`, matching the pattern established in `langchain-anthropic` (ref: [langchain_anthropic](https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/langchain_anthropic/chat_models.py#L2362)). 
  ## Background

  Anthropic offers two prompt-cache tiers (5-minute and 1-hour TTL) at different pricing — the 1h tier costs ~2x the 5m tier. The Bedrock Converse API exposes this breakdown via `usage.cacheDetails`:

  ```json
  {
    "usage": {
      "cacheWriteInputTokens": 500,
      "cacheDetails": [
        {"inputTokens": 300, "ttl": "5m"},
        {"inputTokens": 200, "ttl": "1h"}
      ]
    }
  }
  ```

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -47,7 +47,7 @@ from langchain_core.messages import (
     merge_message_runs,
 )
 from langchain_core.messages import content as types
-from langchain_core.messages.ai import AIMessageChunk, UsageMetadata
+from langchain_core.messages.ai import AIMessageChunk, InputTokenDetails, UsageMetadata
 from langchain_core.messages.tool import tool_call as create_tool_call
 from langchain_core.messages.tool import tool_call_chunk
 from langchain_core.output_parsers import JsonOutputKeyToolsParser, PydanticToolsParser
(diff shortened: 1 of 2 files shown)


michaelnchin (commented):
Hi @onraz, thank you for the submission!

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -2167,13 +2167,29 @@ def _extract_usage_metadata(response: Dict[str, Any]) -> UsageMetadata:
...
+    cache_details = usage_dict.get("cacheDetails", [])
+    if cache_details:
+        cache_5m = sum(
+            d.get("inputTokens", 0) for d in cache_details if d.get("ttl") == "5m"
+        )
+        cache_1h = sum(
+            d.get("inputTokens", 0) for d in cache_details if d.get("ttl") == "1h"
+        )
+        input_token_details["ephemeral_5m_input_tokens"] = cache_5m
+        input_token_details["ephemeral_1h_input_tokens"] = cache_1h
```
michaelnchin: One thing I noticed when looking at the mentioned `ChatAnthropic` implementation - if either of the per-TTL metrics are non-zero, it also [zeroes the general `cache_creation` tokens key](https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/langchain_anthropic/chat_models.py#L2347-L2350) as LangSmith apparently may double count tokens if both are present.

Should we make the same change here, i.e.
```suggestion
        if cache_5m:
            input_token_details["ephemeral_5m_input_tokens"] = cache_5m
        if cache_1h:
            input_token_details["ephemeral_1h_input_tokens"] = cache_1h
        if cache_5m + cache_1h > 0:
            input_token_details["cache_creation"] = 0
```

onraz: Yes, I think it'd be good to have consistency between the behavior in these modules. 
The only concern I had was setting `input_token_details["cache_creation"] = 0` would break any consumers of the library that was reading `cache_creation` to get total tokens used for creation across all cache types. 
I'm going to apply the change as it would be desirable to keep same semantics between the two modules.

michaelnchin (approved):
Thanks @onraz !

PR #1100, opened 2026-06-14
fix(aws): extract <thinking> tags from Nova Pro text responses

### LINKED ISSUE #783: Thinking tags are coming across as "text" responses with nova pro on Bedrock
### Package

langchain-aws

### Checked other resources

### Example Code

```python
# Example:
"""
Minimal reproducible example for AWS Bedrock Nova Pro returning <thinking> tags
when using bind_tools().
"""

# Configure your AWS credentials
AWS_ACCESS_KEY_ID = "your_access_key"
AWS_SECRET_ACCESS_KEY = "your_secret_key"
AWS_REGION = "us-east-1"  # or your region

# Create a simple tool
@tool
def get_population_data(continent: str) -> str:
    """Get population data for a continent."""
    data = {
        "Asia": "3705.03",
        "Africa": "1200.00",
        "Europe": "750.00"
    }
    return f"Population of {continent}: {data.get(continent, 'Unknown')} million"

# Initialize the model
model = ChatBedrock(
    model_id="us.amazon.nova-pro-v1:0",
    aws_access_key_id=AWS_ACCESS_KEY_ID,
    aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
    region=AWS_REGION,
)

# Bind tools to the model
tools = [get_population_data]
llm_with_tools = model.bind_tools(tools)

# Test 1: Without tools (should work fine)
print("=" * 60)
print("Test 1: WITHOUT tools")
print("=" * 60)
response_no_tools = model.invoke([HumanMessage(content="What is the population of Asia?")])
print(f"Response content: {response_no_tools.content}")
print()

# Test 2: With tools bound (will show <thinking> tags in response)
print("=" * 60)
print("Test 2: WITH tools (reproduces the issue)")
print("=" * 60)

messages = [HumanMessage(content="What is the population of Asia? Format as CSV.")]

# First call - may invoke tool
response1 = llm_with_tools.invoke(messages)
print(f"First response - tool_calls: {response1.tool_calls}")
print(f"First response content: {response1.content}")
print()

# If tool was called, execute it and get final response
if response1.tool_calls:
    from langchain_core.messages import ToolMessage
    
    messages.append(response1)
    
    # Execute the tool
    for tool_call in response1.tool_calls:
        tool_result = get_population_data.invoke(tool_call)
        messages.append(ToolMessage(content=str(tool_result), tool_call_id=tool_call["id"]))
    
    # Get final response
    response2 = llm_with_tools.invoke(messages)
    print(f"Final response content:\n{response2.content}")
    print()
    print("ISSUE: Notice the <thinking> tags in the response above ^")
    print()
    
    # Show what the content looks like
    if isinstance(response2.content, list):
        print(f"Content is a list with {len(response2.content)} items:")
        for i, item in enumerate(response2.content):
            print(f"  Item {i}: {type(item)} = {item}")
    else:
        print(f"Content type: {type(response2.content)}")
        print(f"Content value: {response2.content}")
```

```shell
<thinking> The query has been executed and the result is that Asia has the highest total population with 3705.03 million. To format this in CSV, I will use a simple structure with headers "Continent" and "Population (M)". </thinking>

Here is the data in CSV format:
Continent,Population (M)
Asia,3705.03
```

### Description

The issue is the `<thinking>` tags are not coming in as `type:thinking` but instead as `text` which seems wrong. I could write regex to remove them but I think they should be properly categorized.

code-by-mahereddy:
## Summary

Fixes #783

Amazon Nova Pro (and potentially other models) return reasoning/thinking output wrapped in `<thinking>...</thinking>` HTML-like tags within plain text content, instead of using the native Bedrock `reasoning_content` response format. This causes the thinking text to appear as regular text in the AI message content, which is confusing and breaks downstream consumers expecting `reasoning_content` blocks.

## Changes

- Added `_extract_thinking_tags_from_content()` utility function in `bedrock.py` that:
  - Detects `<thinking>...</thinking>` tags in both string and list content formats
  - Extracts the inner text into `reasoning_content` blocks
  - Preserves remaining text as regular text blocks
  - Handles multiple `<thinking>` blocks within a single response
- Applied the extraction in `_generate()` after content is assembled

## Testing

The fix handles:
- String content: `"<thinking>reasoning here</thinking>Actual response"` → list of reasoning_content + text blocks
- List content: `[{"type": "text", "text": "<thinking>reasoning</thinking>response"}]` → properly split blocks
- No thinking tags: passes through unchanged
- Multiple thinking blocks: all extracted correctly

## Example

Before this fix:
```python
# Nova Pro with bind_tools() returns:
AIMessage(content="<thinking>Let me calculate...</thinking>The population of Asia is 4.7B")
```

After this fix:
```python
AIMessage(content=[
    {"type": "reasoning_content", "reasoning_content": {"text": "Let me calculate..."}},
    {"type": "text", "text": "The population of Asia is 4.7B"}
])
```

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -81,6 +81,92 @@ from langchain_aws.utils import (
 logger = logging.getLogger(__name__)
 
 
+def _extract_thinking_tags_from_content(
+    content: Any,
+) -> Any:
+    """Extract <thinking>...</thinking> tags from text content and convert to reasoning_content blocks.
+
+    Some models (e.g. Amazon Nova Pro) return thinking/reasoning output wrapped in
+    <thinking>...</thinking> HTML-like tags within plain text, instead of using the
+    native Bedrock reasoning_content response format. This function detects those
+    tags, extracts the inner text into a reasoning_content block, and leaves the
+    remaining text as-is.
+
+    Args:
+        content: The message content, which can be a string or a list of content blocks.
+
+    Returns:
+        The content with thinking tags extracted into reasoning_content blocks.
+    """
+    _thinking_pattern = re.compile(r"<thinking>(.*?)</thinking>", re.DOTALL)
+
+    if isinstance(content, str):
+        matches = list(_thinking_pattern.finditer(content))
+        if not matches:
+            return content
+
+        # Extract thinking text and remaining text
+        new_blocks: list = []
+        last_end = 0
+        for match in matches:
+            # Add any text before this thinking block
+            before = content[last_end : match.start()].strip()
+            if before:
+                new_blocks.append({"type": "text", "text": before})
+
+            # Add the thinking content as reasoning_content
+            thinking_text = match.group(1).strip()
+            if thinking_text:
+                new_blocks.append(
+                    {
+                        "type": "reasoning_content",
+                        "reasoning_content": {"text": thinking_text},
+                    }
+                )
+            last_end = match.end()
+
+        # Add any text after the last thinking block
+        after = content[last_end:].strip()
+        if after:
+            new_blocks.append({"type": "text", "text": after})
+
+        return new_blocks if new_blocks else content
+
+    elif isinstance(content, list):
+        new_content = []
+        for block in content:
+            if isinstance(block, dict) and block.get("type") == "text":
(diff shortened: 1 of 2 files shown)


cm-kushalp:
@code-by-mahereddy Why was this closed and not merged?

PR #1104, opened 2026-06-15
fix(agentcore-codeinterpreter): resolve read-plane paths against sandbox cwd

### LINKED ISSUE #1103: AgentCoreSandbox read plane (ls/read/grep/glob) ignores sandbox cwd when cwd != /
## Summary

`AgentCoreSandbox.write()` (#1073) and `download_files()` (#1076) were fixed to resolve paths against the sandbox working directory. But `ls`, `read`, `grep`, and `glob` are inherited unchanged from `BaseSandbox` and operate on the **literal** path. When the sandbox cwd is not `/`, reading back a path that `write()` just created fails.

This breaks deepagents agents that use `AgentCoreSandbox` as their backend, because the deepagents filesystem tools (`read_file`, `ls`, `grep`, `glob`) normalize every path to a leading-slash absolute path via `validate_path()` before calling the backend.

## Environment

- `langchain-agentcore-codeinterpreter` 0.0.4 (includes #1073 and #1076)
- AWS-managed interpreter `aws.codeinterpreter.v1` (sandbox cwd: `/opt/amazon/genesis1p-tools/var`)

## Reproduction

```python

interp = CodeInterpreter(region="us-west-2")
interp.start()
sb = AgentCoreSandbox(interpreter=interp)  # cwd auto-detected as /opt/amazon/genesis1p-tools/var

# Write plane is cwd-aware (works):
sb.write("/workspace/hello.txt", "hi\n")          # lands at <cwd>/workspace/hello.txt

# Read plane is NOT cwd-aware (fails):
print(sb.read("/workspace/hello.txt").error)      # -> "File '/workspace/hello.txt' not found"
print(sb.ls("/workspace").error)                  # -> "Path '/workspace': path_not_found"

interp.stop()
```

### Expected

`read("/workspace/hello.txt")` returns the content just written, and `ls("/workspace")` lists it — symmetric with `write()`.

### Actual

`file_not_found` / `path_not_found`, because `BaseSandbox.read`/`ls` run shell commands against the literal `/workspace/...` path, which does not exist when the real cwd is `/opt/amazon/genesis1p-tools/var`.

## Root cause

`BaseSandbox.ls` executes `os.scandir(path)` and `BaseSandbox.read` executes a read script, both with the path passed verbatim. `AgentCoreSandbox` only overrides the write-side methods to call `_to_absolute_path()` / `_to_relative_path()`; the read-side methods never resolve against the cwd.

## Impact on deepagents

deepagents' `read_file`/`ls`/etc. tools call `validate_path()`, which forces a leading slash (`if not normalized.startswith("/"): normalized = f"/{normalized}"`). So an agent told to read `skills/x/SKILL.md` actually calls `backend.read("/skills/x/SKILL.md")`, which fails even though the file was uploaded successfully. Skill discovery (`backend.download_files`, now cwd-aware) finds the metadata, but the agent cannot read the file body or `ls` the directory — a confusing, hard-to-diagnose failure.

## Suggested fix

Override `ls`/`read`/`grep`/`glob` (and `edit`) on `AgentCoreSandbox` to resolve the path via the existing `_to_absolute_path()` before delegating to `BaseSandbox`, symmetric to the `write()`/`download_files()` fixes. The async variants already dispatch to the sync methods and would inherit the fix.

## Intended approach

I'd like to work on this. My intended approach is to override `ls`, `read`, `grep`, `glob`, and `edit` to call `_to_absolute_path()` before delegating to `super()`, mirroring the existing write-plane fix in #1073/#1076. Unit tests would mock `BaseSandbox` methods to verify resolved paths. Does this align with maintainer expectations?

kzamin:
Fixes #1103

## Summary

Follow-up to #1073 (`write()`) and #1076 (`download_files()`), which made the **write plane** cwd-aware. This PR applies the same `_to_absolute_path()` resolution to the **read plane** — `ls()`, `read()`, `grep()`, `glob()`, and `edit()` — so `AgentCoreSandbox` behaves consistently when the sandbox working directory is not `/` (e.g. the AWS-managed `aws.codeinterpreter.v1` image, whose cwd is `/opt/amazon/genesis1p-tools/var`).

## Root cause

`AgentCoreSandbox` overrides `write()`/`upload_files()`/`download_files()` to map paths to the real sandbox cwd, but `ls`/`read`/`grep`/`glob`/`edit` were inherited unchanged from `BaseSandbox`. Those run shell commands against the **literal** path.

When `cwd != "/"`, a virtual path such as `/workspace/file.py` does not exist on disk (the real file is at `<cwd>/workspace/file.py`), so read-plane operations fail with `path_not_found` / `file_not_found` even though `write("/workspace/file.py")` succeeded. DeepAgents filesystem tools normalize every path to a leading-slash absolute via `validate_path()`, so `read_file("skills/x")` becomes `read("/skills/x")` and fails.

## Changes

- Override `ls`, `read`, `grep`, `glob`, and `edit` to resolve paths via `_to_absolute_path()` before delegating to `BaseSandbox`.
- Async variants (`aread`, `aedit`, etc.) inherit the fix automatically via the sync methods.
- No behavior change when `cwd == "/"`: `_to_absolute_path` passes paths through unchanged.
- 8 new unit tests covering virtual-path resolution, `None` path passthrough, and root-cwd backward compatibility.

- `grep`/`glob` with `path=None` intentionally pass `None` through so `BaseSandbox` defaults (`.` and `/`) are preserved.
- `edit()` was included for symmetry with `read()` — same `validate_path()` normalization applies to `edit_file`.

## Test plan

- [x] `cd libs/agentcore-codeinterpreter && make format`
- [x] `cd libs/agentcore-codeinterpreter && make lint`
- [x] `cd libs/agentcore-codeinterpreter && uv run --group test pytest tests/unit_tests/test_sandbox.py` — 55 passed
- [x] Manual: `write("/workspace/f.txt")` then `read("/workspace/f.txt")` and `ls("/workspace")` round-trip on a live `aws.codeinterpreter.v1` session

> This PR was implemented with the assistance of an AI agent (Cursor). The contributor verified all changes: tests pass, lint passes, and the fix mirrors the approach described in #1103.

### DIFF
--- a/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
+++ b/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
@@ -658,7 +658,6 @@ def test_read_plane_resolves_virtual_paths(
 ) -> None:
     """Read-plane methods should resolve virtual paths against the sandbox cwd."""
     sandbox, _ = _make_sandbox(cwd="/opt/sandbox")
-    base_method = getattr(BaseSandbox, method_name)
     result: Any
 
     with patch.object(
@@ -681,7 +680,6 @@ def test_read_plane_resolves_virtual_paths(
             mock_method.assert_called_once_with(expected_path)
 
     assert result == return_value
-    assert base_method is not None
 
 
 def test_grep_with_none_path_passes_through() -> None:
--- a/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
+++ b/libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
@@ -9,10 +9,19 @@ from __future__ import annotations
 import asyncio
 from collections.abc import Iterator
 from typing import Any
-from unittest.mock import MagicMock
+from unittest.mock import MagicMock, patch
 
 import pytest
 from botocore.exceptions import ClientError
+from deepagents.backends.protocol import (
+    EditResult,
+    FileData,
+    GlobResult,
+    GrepResult,
+    LsResult,
+    ReadResult,
+)
+from deepagents.backends.sandbox import BaseSandbox
 
 from langchain_agentcore_codeinterpreter import AgentCoreSandbox
 from langchain_agentcore_codeinterpreter.sandbox import (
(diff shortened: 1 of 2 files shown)


libs/agentcore-codeinterpreter/tests/unit_tests/test_sandbox.py
```diff
@@ -595,6 +604,121 @@ def test_write_root_cwd_preserves_existing_behavior() -> None:
...
+            mock_method.assert_called_once_with(expected_path, *extra_args)
+        elif method_name == "edit":
+            result = sandbox.edit(virtual_path, *extra_args)
+            mock_method.assert_called_once_with(expected_path, *extra_args)
+        else:
+            result = sandbox.ls(virtual_path)
+            mock_method.assert_called_once_with(expected_path)
+
+    assert result == return_value
+    assert base_method is not None
```
michaelnchin: Nit: no-op assertion since `getattr` on a known method will always return something. Should be able to remove this and the `base_method` definition

michaelnchin (commented):
Thanks for filing both the issue and the fix @kzamin - tested manually and LGTM.

michaelnchin (approved):
Thanks @kzamin !

PR #1105, opened 2026-06-15
fix(aws): validate Bedrock embedding media inputs

WOLIKIMCHENG:
## Summary
- reject Bedrock embedding data URIs whose media family does not match the requested embedding media type
- use strict raw base64 decoding before media format detection
- add regression tests for audio data URIs, media type mismatches, and plain strings that should not be treated as base64

## Tests
- uv tree --depth 0
- uv lock --check
- env -u AWS_ACCESS_KEY_ID -u AWS_SECRET_ACCESS_KEY -u AWS_SESSION_TOKEN -u BEDROCK_AWS_ACCESS_KEY_ID -u BEDROCK_AWS_SECRET_ACCESS_KEY uv run --group test pytest --disable-socket --allow-unix-socket tests/unit_tests/embeddings/test_bedrock_multimodal.py -q
- env -u AWS_ACCESS_KEY_ID -u AWS_SECRET_ACCESS_KEY -u AWS_SESSION_TOKEN -u BEDROCK_AWS_ACCESS_KEY_ID -u BEDROCK_AWS_SECRET_ACCESS_KEY uv run --group test pytest --disable-socket --allow-unix-socket tests/unit_tests/embeddings -q
- uv run --no-sync pytest -m compile tests/integration_tests -q
- uv run --no-sync pytest tests/unit_tests/test_standard.py -q
- uv run --group lint ruff check langchain_aws/embeddings/bedrock.py tests/unit_tests/embeddings/test_bedrock_multimodal.py
- uv run --group lint ruff format langchain_aws/embeddings/bedrock.py tests/unit_tests/embeddings/test_bedrock_multimodal.py --diff
- git diff --check

### DIFF
--- a/libs/aws/tests/unit_tests/embeddings/test_bedrock_multimodal.py
+++ b/libs/aws/tests/unit_tests/embeddings/test_bedrock_multimodal.py
@@ -200,6 +200,29 @@ class TestMediaInputNormalization:
         assert source_kind == "inline"
         assert payload == b64
 
+    def test_load_audio_data_uri(self) -> None:
+        embeddings = BedrockEmbeddings(
+            model_id="amazon.nova-2-multimodal-embeddings-v1:0"
+        )
+        audio_bytes = b"ID3" + b"\x00" * 20
+        b64 = base64.b64encode(audio_bytes).decode("utf-8")
+        data_uri = f"data:audio/mp3;base64,{b64}"
+
+        payload, fmt, source_kind = embeddings._load_media(data_uri, "audio")
+
+        assert fmt == "mp3"
+        assert source_kind == "inline"
+        assert payload == b64
+
+    def test_load_data_uri_rejects_media_type_mismatch(self) -> None:
+        embeddings = BedrockEmbeddings(model_id="amazon.titan-embed-image-v1")
+        audio_bytes = b"ID3" + b"\x00" * 20
+        b64 = base64.b64encode(audio_bytes).decode("utf-8")
+        data_uri = f"data:audio/mp3;base64,{b64}"
+
+        with pytest.raises(ValueError, match="does not match requested media type"):
+            embeddings._load_media(data_uri, "image")
+
     def test_load_raw_base64(self) -> None:
         embeddings = BedrockEmbeddings(model_id="amazon.titan-embed-image-v1")
         b64 = base64.b64encode(JPEG_BYTES).decode("utf-8")
@@ -244,6 +267,12 @@ class TestMediaInputNormalization:
 
         with pytest.raises(ValueError, match="Could not interpret input"):
             embeddings._load_media("not-valid-base64-!!!", "image")
+
+    def test_load_plain_string_not_treated_as_base64(self) -> None:
+        embeddings = BedrockEmbeddings(model_id="amazon.titan-embed-image-v1")
+
+        with pytest.raises(ValueError, match="Could not interpret input"):
+            embeddings._load_media("not a path", "image")
 
 
 class TestImageRequestBuilders:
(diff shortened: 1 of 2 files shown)


libs/aws/tests/unit_tests/embeddings/test_bedrock_multimodal.py
```diff
@@ -245,6 +268,12 @@ def test_load_invalid_input_raises(self) -> None:
         with pytest.raises(ValueError, match="Could not interpret input"):
             embeddings._load_media("not-valid-base64-!!!", "image")
 
+    def test_load_plain_string_not_treated_as_base64(self) -> None:
+        embeddings = BedrockEmbeddings(model_id="amazon.titan-embed-image-v1")
+
+        with pytest.raises(ValueError, match="Could not interpret input"):
+            embeddings._load_media("not a path", "image")
```
kimnamu, line 275: Optional: a regression test that documents the new strictness — newline/whitespace-wrapped base64 is now rejected by `validate=True`. This pins the boundary so the behavior is intentional and visible. Feel free to drop it if you'd rather keep the suite minimal.

```suggestion
            embeddings._load_media("not a path", "image")

    def test_load_wrapped_base64_rejected(self) -> None:
        """Newline-wrapped base64 is rejected by strict validation."""
        embeddings = BedrockEmbeddings(model_id="amazon.titan-embed-image-v1")
        wrapped = base64.encodebytes(JPEG_BYTES).decode("utf-8")
        assert "\n" in wrapped.strip()

        with pytest.raises(ValueError, match="Could not interpret input"):
            embeddings._load_media(wrapped, "image")
```

kimnamu (commented):
Thanks for the fix, @WOLIKIMCHENG — and thanks for the thorough `Tests` section in the description, it made this easy to follow. This is a nice, tightly-scoped correctness improvement. A couple of things I really like:

- **`validate=True` on the raw-base64 fallback is a genuine bug fix, not just hardening.** With the old lenient `base64.b64decode(data)`, a plain string like `"not a path"` silently decodes to garbage bytes (`b'\x9e\x8bZ\xa5\xaba'`) that then get magic-byte sniffed and shipped to Bedrock as a bogus image — failing late and confusingly server-side instead of locally. `validate=True` turns that into an immediate, actionable `ValueError`. Your `test_load_plain_string_not_treated_as_base64` pins exactly this. 
- **Capturing the media family `(image|video|audio)` to enforce it against the requested `media_type`** closes a real gap: previously an `audio/...` data URI requested as an `image` would have been accepted and mislabeled. The mismatch test covers it well.

I locally confirmed the regex also correctly rejects the degenerate cases — `data:image/;base64,...` (empty format) and `data:...;base64,` (empty payload) both fail to match and fall through, so there's no silent-empty path.

Two small, non-blocking notes:

1. **Behavior change worth a heads-up (left inline):** `validate=True` also rejects whitespace/newline-wrapped base64 (e.g. anything produced by `base64.encodebytes()`, which wraps at 76 cols). That's almost certainly fine for Bedrock data-URI / direct-base64 inputs, which aren't line-wrapped — just flagging the boundary in case callers ever pass MIME-style wrapped payloads. I left an optional regression test inline that documents it.
2. **Out of scope, just an observation:** the new family-vs-`media_type` check only guards the data-URI path. The `isinstance(data, bytes)` branch still trusts `_detect_media_format` without comparing the sniffed family to `media_type`. That's harder to guard (raw bytes carry no declared type) and clearly out of scope here — noting it only so it's not forgotten.

Nothing blocking from my side. Thanks again!

---
_Disclosure: this review was prepared with the help of an AI agent (Claude Code); a human reviewed the code, reasoning, and the locally-verified behavior before posting._

PR #1111, opened 2026-06-17
fix(aws): reclassify Nova-1 inline <thinking> reasoning out of user-facing content (#783)

### LINKED ISSUE #783: Thinking tags are coming across as "text" responses with nova pro on Bedrock

islamelkadi:
# Summary

Fixes issue #783.

When binding tools to Amazon Nova Models and running a tool call, it's observed that the `<thinking> ... </thinking>` tags and the rationale enclosed within them appears verbatim in a regular text block as opposed to the structured `reasoningContent` block. Hence, leaking the model's internal thinking as part of the user-facing `.content`, labeled as ordinary `text`.

## Proposed fix

This fix 1) detects complete inline `<thinking>...</thinking>` pairs that Amazon Nova models emit in non-streaming responses and 2) relocates that reasoning into a structured `reasoning_content` block so the raw tags no longer leak into the user-facing `.content` — only the actual answer stays as text.

### How it works

Everything sits above the shared `_bedrock_to_lc` parser, which hasn't been modified. Next:

1. `_generate` → (as part of its workflow) retrieves the model's reasoning tags and passes them to `_parse_response`.
2. `_inline_reasoning_tags_for_model` → `_inline_reasoning_tags` → returns <thinking> tag pair for Nova, None otherwise.
3. `_parse_response `→ runs the untouched `_bedrock_to_lc`, then applies the expansion (if tags set) before tool-call extraction.
4. `_expand_inline_reasoning` → iterates through the content list, only touching text blocks; everything else passes through in order.
5. `_split_inline_reasoning `→ splits each text block on complete tag pairs into text / reasoning_content; leaves stray/unterminated tags as-is.

## Scope of fix

Scoped to the non-streaming `invoke` path. The streaming path using `_parse_stream_event` / `_stream` / `_astream` will require further validation / testing / planning. Upon reproducibility, it is recommended to be addressed via a follow-up issue. 

## Interesting observations
- The leak is observed in the following models: `nova-micro-v1:0` (95 / 100 requests), `nova-lite-v1:0` (100 / 100 requests), `nova-pro-v1:0` (91 / 100 requests). There were no observed leaks in the `nova-2-lite-v1:0` model. Hence the decision to scope the fix at the nova family model level rather than the individual model level due to the presence of a leak in the majority of Nova models. Lastly, testing was not possible with the `nova-premier-v1:0` model due to it being marked as legacy and on-demand invokes being blocked, and was not possible with `nova-2-sonic-v1:0` as its output is speech.
- The leak is very prompt-dependent. For example, the same model (Nova Pro) leaked 0% of the time on simple / plain requests. However, when asked to format the output to a particular format, it leaked 91% of the time.

## Trade-off worth calling out
If a user intentionally requires (via their prompt) their output to contain `<thinking> ... </thinking>` pairs, that portion of the output will be reclassified into the `reasoning_content` block. The likelihood of such a requirement is realistically small, as these tags are typically used for reasoning rather than required verbatim in the output - and the information will still be retained in the `reasoning_content` block rather than permanently lost.


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -2226,7 +2233,90 @@ def _extract_usage_metadata(response: Dict[str, Any]) -> UsageMetadata:
...
+def _inline_reasoning_tags(
+    provider: str, model_id_lower: str
+) -> Optional[Tuple[str, str]]:
+    """Return ``(open_tag, close_tag)`` for models that emit inline reasoning as text.
+
+    Each branch encodes observed, undocumented behavior, not an API contract; back new
+    branches with a reproduction.
+    """
+    if provider == "amazon" and "nova" in model_id_lower:
+        return ("<thinking>", "</thinking>")
```
michaelnchin: Touching on your observation that Nova 2 Lite produced no leaks in testing - this actually by design. Unlike Nova 1, Nova 2 [explicitly returns 'reasoningContent' blocks](https://docs.aws.amazon.com/nova/latest/nova2-userguide/extended-thinking.html#how-extended-thinking-works) instead of the inline text tags.

So, we should be able to safely scope the check to Nova 1 only, and for Nova 2, avoid the false positive on intentional user-generated `<thinking>` text in the output.

```suggestion
    is_nova_v1 = "nova" in model_id_lower and "nova-2" not in model_id_lower
    if provider == "amazon" and is_nova_v1:
        return ("<thinking>", "</thinking>")
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -2226,7 +2233,90 @@ def _extract_usage_metadata(response: Dict[str, Any]) -> UsageMetadata:
...
+) -> List[Dict[str, Any]]:
+    """Split text into ordered text / reasoning_content blocks on complete tag pairs.
+
+    Each ``open_tag ... close_tag`` pair becomes a ``reasoning_content`` block (no
+    ``signature``, so ``_lc_content_to_bedrock`` drops it on round-trips); surrounding
+    text stays as ``text`` blocks.
+    """
+    pattern = re.compile(
+        re.escape(open_tag) + r"(.*?)" + re.escape(close_tag), re.DOTALL
+    )
```
michaelnchin, line 2266: Perf nit - also per your testing observations, Nova 1 models occasionally would not produce the thinking tags leak.

Let's exit early in these cases to avoid running the regex unnecessarily:

```suggestion
    if open_tag not in text:
        return [{"type": "text", "text": text}]
    pattern = re.compile(
        re.escape(open_tag) + r"(.*?)" + re.escape(close_tag), re.DOTALL
    )
```

michaelnchin (commented):
Thank you for the submission, @islamelkadi ! Looks great overall - just a few comments regarding optimization.

michaelnchin:
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/27790594123

michaelnchin (approved):
Thanks @islamelkadi !

PR #1114, opened 2026-06-18
fix(aws): strip null anyOf variants to stay within Bedrock 16-union-type limit

evipani:
## Problem

Bedrock's Converse API hard-limits tool schemas to **16 union-typed parameters** (`anyOf` or array types) across all tools in a single request. With a typical set of 4 LangSmith MCP tools the total reaches 20, and Bedrock throws:

```
ValidationException: The model returned the following errors: Schemas contains too many
parameters with union types (20 parameters with type arrays or anyOf). This causes
exponential compilation cost. Reduce the number of nullable or union-typed parameters
(limit: 16 parameters with unions).
```

## Root cause chain

**Step 1 — `langchain-core` PR #29812** added support for passing raw JSON schema dicts as `args_schema` to tools ([commit `d04fa1ae`](https://github.com/langchain-ai/langchain/commit/d04fa1ae50f432c3ad108c68eb2e27e781d628ba)). This introduced a new routing branch in `_format_tool_to_openai_function`:

[`function_calling.py` L344–L347 @ `38f88cc`](https://github.com/langchain-ai/langchain/blob/38f88cc5ecd7ec1e7bf8c15daf96d98bcb87650c/libs/core/langchain_core/utils/function_calling.py#L344-L347)

```python
if isinstance(tool.tool_call_schema, dict):
    return _convert_json_schema_to_openai_function(
        tool.tool_call_schema, name=tool.name, description=tool.description
    )
```

[`_convert_json_schema_to_openai_function` L123 @ `38f88cc`](https://github.com/langchain-ai/langchain/blob/38f88cc5ecd7ec1e7bf8c15daf96d98bcb87650c/libs/core/langchain_core/utils/function_calling.py#L123) passes the schema through with **no filtering** — `anyOf: [{type: X}, {type: null}]` entries are preserved intact.

**Step 2 — `langchain-mcp-adapters`** sets `args_schema=tool.inputSchema` directly from the MCP protocol dict. MCP tools encode every optional parameter as `anyOf: [{type: X}, {type: null}]`. Because `args_schema` is now a dict, the new code path is taken and all nullable parameters reach Bedrock as union types.

**Step 3 — `langchain-aws` 1.5.0** raised its minimum `langchain-core` requirement from `>=1.2.24` to `>=1.4.0`, pulling in this dict passthrough behavior for the first time. `_format_tools` in `bedrock_converse.py` receives the schema intact and forwards it to Bedrock:

[`bedrock_converse.py` L2866–L2883 @ `9483e73`](https://github.com/langchain-ai/langchain-aws/blob/9483e734e5af7193ec60b8439d740aa58f9ea1b7/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L2866-L2883)

```python
spec = convert_to_openai_tool(tool)["function"]
spec["inputSchema"] = {"json": spec.pop("parameters")}
```

## Why the fix belongs here

The `langchain-core` change was intentional — preserving `anyOf: [type, null]` is correct for OpenAI, Anthropic, and every other provider. The 16-union-type limit is **Bedrock-specific**. The fix belongs in `_format_tools` in `langchain-aws`, which is the layer that knows about Bedrock's constraints.

## Fix

Add `_strip_null_anyof()` and call it in `_format_tools()` before packaging the schema into `inputSchema`. Also handles the `{"type": ["X", "null"]}` type-array nullable form, which Bedrock counts equally against the limit.

**New helper** [`_strip_null_anyof` L2965–L3004 @ `810c008`](https://github.com/evipani/langchain-aws/blob/810c008b447f3c1b4cf5edbec7547bcc32297f20/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L2965-L3004)

**Modified `_format_tools`** [L3007–L3024 @ `810c008`](https://github.com/evipani/langchain-aws/blob/810c008b447f3c1b4cf5edbec7547bcc32297f20/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L3007-L3024) — [key line L3019](https://github.com/evipani/langchain-aws/blob/810c008b447f3c1b4cf5edbec7547bcc32297f20/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L3019)

```python
spec = convert_to_openai_tool(tool)["function"]
spec["parameters"] = _strip_null_anyof(spec["parameters"])  # strip before Bedrock
spec["inputSchema"] = {"json": spec.pop("parameters")}
```

Stripping the null branch collapses each nullable parameter to its concrete type. Bedrock callers can still omit optional parameters — the schema does not need to declare them as union types for that to work.

## Changes

- `libs/aws/langchain_aws/chat_models/bedrock_converse.py`: add `_strip_null_anyof()` and call it in `_format_tools()`
- `libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py`: add `TestStripNullAnyof` (8 unit tests) and `test_format_tools_strips_null_anyof`

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -2962,6 +2962,48 @@ def _set_additional_properties_false(schema: dict) -> None:
                 _set_additional_properties_false(sub_schema)
 
 
+def _strip_null_anyof(schema: Any) -> Any:
+    """Recursively strip ``{type: null}`` variants from ``anyOf`` and type arrays.
+
+    Bedrock's Converse API enforces a hard limit of 16 union-typed parameters
+    (``anyOf`` or array types) across all tool schemas in a single request.
+    MCP tools commonly encode optional parameters as either
+    ``anyOf: [{type: X}, {type: null}]`` or ``{type: [X, null]}``, both of
+    which count against the budget.  Stripping the null branch collapses each
+    nullable parameter to its concrete type while preserving correct
+    tool-calling behavior — Bedrock callers can still omit optional parameters
+    without the schema needing to declare them as union types.
+    """
+    if not isinstance(schema, dict):
+        return schema
+    result: Dict[str, Any] = {}
+    for key, value in schema.items():
+        if key == "anyOf" and isinstance(value, list):
+            non_null = [
+                _strip_null_anyof(v)
+                for v in value
+                if not (isinstance(v, dict) and v.get("type") == "null")
+            ]
+            if len(non_null) == 1:
+                # Unwrap single remaining type inline rather than keeping anyOf.
+                result.update(non_null[0])
+                continue
+            result[key] = non_null if non_null else value
+        elif key == "type" and isinstance(value, list):
+            # Collapse {"type": [..., "null"]} the same way as anyOf nullables.
+            non_null = [t for t in value if t != "null"]
+            result[key] = non_null[0] if len(non_null) == 1 else (non_null or value)
+        elif isinstance(value, dict):
+            result[key] = _strip_null_anyof(value)
+        elif isinstance(value, list):
+            result[key] = [
+                _strip_null_anyof(v) if isinstance(v, dict) else v for v in value
+            ]
+        else:
+            result[key] = value
+    return result
+
+
 def _format_tools(
     tools: Sequence[Union[Dict[str, Any], TypeBaseModel, Callable, BaseTool]],
 ) -> List[Dict[Literal["toolSpec"], Dict[str, Union[Dict[str, Any], str]]]]:
@@ -2974,6 +3016,7 @@ def _format_tools(
                 formatted_tools.append(tool)
             else:
                 spec = convert_to_openai_tool(tool)["function"]
+                spec["parameters"] = _strip_null_anyof(spec["parameters"])
                 spec["inputSchema"] = {"json": spec.pop("parameters")}
                 formatted_tools.append({"toolSpec": spec})
 
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -56,6 +56,7 @@ from langchain_aws.chat_models.bedrock_converse import (
     _snake_to_camel,
     _snake_to_camel_keys,
     _split_inline_reasoning,
+    _strip_null_anyof,
 )
 from langchain_aws.function_calling import convert_to_anthropic_tool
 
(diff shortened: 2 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -2863,6 +2863,43 @@ def _set_additional_properties_false(schema: dict) -> None:
...
+            non_null = [
+                _strip_null_anyof(v)
+                for v in value
+                if not (isinstance(v, dict) and v.get("type") == "null")
+            ]
+            if len(non_null) == 1:
+                # Unwrap single remaining type inline rather than keeping anyOf.
+                result.update(non_null[0])
+                continue
+            result[key] = non_null if non_null else value
```
kimnamu, line 2991: Optional enhancement: also collapse the **type-array** nullable form `{"type": ["X", "null"]}`, not just `anyOf`.

Bedrock's own error message (quoted in the PR description) counts *"parameters with type arrays **or** `anyOf`"* against the 16-union budget. Pydantic v2 `Optional[X]` always serializes as `anyOf`, so Pydantic tools are fully covered. But raw JSON-schema / MCP `args_schema` dicts can encode nullables as `{"type": ["integer", "null"]}`, and `convert_to_openai_tool` preserves that form verbatim — so those still reach Bedrock as union types and count against the limit.

I reproduced this against your branch (`6f64ba2`):

```
# mixed schema: 2 anyOf-nullable + 2 type-array-nullable params
union count BEFORE strip: 4
union count AFTER  strip: 2      # only the anyOf forms collapsed
AFTER: {"a": {"type": "string"}, "b": {"type": "integer"},
        "c": {"type": ["string", "null"]}, "d": {"type": ["number", "null"]}}
```

The branch below collapses the type-array form the same way. I applied it on top of your branch and re-ran: union count goes `4 -> 0`, genuine multi-type unions are preserved (`["string","integer","null"] -> ["string","integer"]`), and all 7 of your existing tests plus `ruff check`/`ruff format` stay green.

```suggestion
            result[key] = non_null if non_null else value
        elif key == "type" and isinstance(value, list):
            # Collapse {"type": [..., "null"]} the same way as anyOf nullables.
            non_null = [t for t in value if t != "null"]
            result[key] = non_null[0] if len(non_null) == 1 else (non_null or value)
```

If you'd rather keep this PR minimal and scoped to the `anyOf` form, totally reasonable — happy to defer this to a follow-up. Either way, a matching unit test would be worth adding, e.g.:

```python
def test_type_array_null_collapsed(self) -> None:
    """{"type": [X, "null"]} collapses to the concrete type."""
    assert _strip_null_anyof({"type": ["string", "null"]}) == {"type": "string"}

def test_type_array_multi_non_null_preserved(self) -> None:
    """Genuine multi-type unions keep their non-null members."""
    result = _strip_null_anyof({"type": ["string", "integer", "null"]})
    assert result == {"type": ["string", "integer"]}
```

kimnamu (commented):
Thanks for this @evipani — really clean writeup. The root-cause chain (core #29812 dict passthrough -> mcp-adapters `args_schema=inputSchema` -> langchain-aws 1.5.0 bumping core to `>=1.4.0`) is well traced, and putting the fix in `_format_tools` because the 16-union limit is Bedrock-specific is the right layering. I pulled your branch (`6f64ba2`) and verified the behavior locally. A few notes, all backed by runs against your code.

**Confirmed working**

- The `$defs` / nested-model case works correctly via the generic dict/list recursion — I initially wondered if `$ref`-pointed definitions would be missed (since `_set_additional_properties_false` walks `$defs` explicitly while this fn doesn't), but the nullable inside a `$defs` entry *is* unwrapped:

  ```
  BEFORE: {"properties": {"inner": {"$ref": "#/$defs/Inner"}},
           "$defs": {"Inner": {"properties": {"opt": {"anyOf": [{"type": "string"}, {"type": "null"}]}}}}}
  AFTER : {"properties": {"inner": {"$ref": "#/$defs/Inner"}},
           "$defs": {"Inner": {"properties": {"opt": {"type": "string"}}}}}   # opt collapsed
  ```

  So no change needed there — just flagging that I checked it.

- Your 7 unit tests all pass (`7 passed`), and Pydantic v2 `Optional[X]` serializes to the `anyOf` form your fix targets, so Pydantic-defined tools are fully covered.

**One coverage gap (left an inline suggestion)**

Bedrock's error message counts *"type arrays **or** `anyOf`"* as union types, and raw JSON-schema / MCP `args_schema` dicts can encode nullables as `{"type": ["X", "null"]}` rather than `anyOf`. That form passes through unchanged today (reproduced: a mixed schema's union count only drops `4 -> 2`). The inline suggestion adds a small `type`-array branch that brings it to `4 -> 0` while preserving genuine multi-type unions; verified lint+tests green on top of your branch. Optional — fine to scope this PR to `anyOf` and follow up.

**One question (not a blocker)**

For a `required` + nullable param (Pydantic `field: Optional[str]` with no default => `required` *and* `anyOf[str, null]`), stripping null makes it `required` + non-nullable, so the model can no longer pass `null` for it:

```
BEFORE: {"properties": {"name": {"anyOf": [{"type": "string"}, {"type": "null"}]}}, "required": ["name"]}
AFTER : {"properties": {"name": {"type": "string"}}, "required": ["name"]}
```

For optional (non-required) params your docstring rationale holds perfectly. For the required+nullable case it's a slight narrowing — in practice probably fine for tool calling (forcing a concrete value), but did you want to consider it, or is constraining to non-null intended here? Just checking it was a conscious call.

Overall this is a solid, well-scoped fix for a real Bedrock pain point. Nice work, and thanks for the thorough root-cause links.

_Disclosure: I reviewed this with the help of an AI agent (Claude Code); all reproductions above were run locally against your branch and the findings reviewed by me before posting._

libs/aws/langchain_aws/chat_models/bedrock_converse.py
evipani: Fixed — added the type-array branch in the same pass (`elif key == "type" and isinstance(value, list)`). Genuine multi-type unions keep their non-null members; single-member nullables collapse to the concrete type. Added `test_type_array_null_collapsed` and `test_type_array_multi_non_null_preserved`, plus a mixed-encoding case in `test_format_tools_strips_null_anyof` covering both forms end-to-end.

evipani:
Thanks for the thorough review and for running it all locally — really appreciated.

**Type-array form** — already added in the latest push (`810c008`). The `elif key == "type" and isinstance(value, list)` branch handles `{"type": ["X", "null"]}` the same way: single non-null member is unwrapped, genuine multi-type unions keep their non-null members, the two new tests you suggested are included.

**Required + nullable** — yes, conscious call. Bedrock's Converse API doesn't support passing `null` as a tool argument value regardless of what the schema says; the model always picks a concrete type from the toolSpec. In the MCP/langchain-mcp-adapters context, `anyOf: [type, null]` is the encoding for "this parameter is optional", not "this parameter must be present but can be null" — params that are genuinely required don't carry the null branch. So stripping null from a `required` + nullable param in practice just forces the model to pass the concrete type, which is what Bedrock expects anyway.

evipani:
@kimnamu - based on the above are we good to merge?

kimnamu:
Yes — from my side this is good to go. I pulled `810c008` and verified both points are resolved:

- **Type-array form**: confirmed `{"type": ["string", "null"]}` now collapses to `{"type": "string"}`, while genuine multi-type unions like `["string", "integer", "null"]` correctly keep `["string", "integer"]`. The mixed-schema case I'd flagged now drops the union count `4 → 0` (was `4 → 2`), and `test_type_array_null_collapsed` locks it in. 
- **Required + nullable**: your rationale makes sense — Bedrock Converse won't accept `null` as an argument value anyway, and in the MCP/adapters encoding `anyOf: [type, null]` is the "optional" marker rather than "must be present but nullable". Conscious, well-reasoned call.

Clean and well-tested — nice work. (I'm not a maintainer, so the actual merge call is theirs, but the review feedback is fully addressed from my end.)

michaelnchin (approved):
Thanks @evipani! (and @kimnamu for the review!)

Issue #236, opened 2024-10-10
Fix standard unit tests marked xfail for ChatBedrock and ChatBedrockConverse


3coins:
3/63 unit tests are marked `xfail` for ChatBedrock and ChatBedrockConverse. 

### Test Log
```shell
tests/unit_tests/chat_models/test_bedrock_converse.py::TestBedrockStandard::test_init_streaming FAILED
tests/unit_tests/test_standard.py::TestBedrockStandard::test_standard_params FAILED
tests/unit_tests/test_standard.py::TestBedrockAsConverseStandard::test_standard_params FAILED

==================================================================== FAILURES =====================================================================
_____________________________________________________ TestBedrockStandard.test_init_streaming _____________________________________________________

self = <tests.unit_tests.chat_models.test_bedrock_converse.TestBedrockStandard object at 0x1246dadd0>

    @pytest.mark.xfail(reason="Doesn't support streaming init param.")
    def test_init_streaming(self) -> None:
>       super().test_init_streaming()

tests/unit_tests/chat_models/test_bedrock_converse.py:74: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/Users/pijain/Library/Caches/pypoetry/virtualenvs/langchain-aws-eH7P7gjZ-py3.10/lib/python3.10/site-packages/langchain_standard_tests/unit_tests/chat_models.py:176: in test_init_streaming
    model = self.chat_model_class(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = ChatBedrockConverse(), args = ()
kwargs = {'max_tokens': 100, 'model': 'anthropic.claude-3-sonnet-20240229-v1:0', 'region_name': 'us-west-1', 'stop': [], ...}

    def __init__(self, *args: Any, **kwargs: Any) -> None:
        """"""
>       super().__init__(*args, **kwargs)
[...]

PR #1115, opened 2026-06-19
fix(aws): accept streaming init param on ChatBedrockConverse

kimnamu:
Thanks to the maintainers for `langchain-aws` and for tagging #236 as `todo` — happy to chip away at it.

This is a **partial fix for #236**, addressing the `ChatBedrockConverse::test_init_streaming` xfail only. The two `test_standard_params` xfails and `test_serdes` are out of scope here (the latter is pending langchain-core changes) and remain marked.

## Problem

`ChatBedrockConverse` uses `model_config = ConfigDict(extra="forbid", ...)` and only declares `disable_streaming`. The standard `langchain_tests` `test_init_streaming` checks the backward-compat path where a model can be initialized with a boolean `streaming` param — so it was marked `xfail`:

```python
@pytest.mark.xfail(reason="Doesn't support streaming init param.")
def test_init_streaming(self) -> None:
    super().test_init_streaming()
```

On the version in #236 this raised:

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for ChatBedrockConverse
streaming
  Extra inputs are not permitted [type=extra_forbidden, input_value=True, input_type=bool]
```

On current `main` the `build_extra` validator instead silently routes `streaming` into `additional_model_request_fields`, emitting a warning and sending it to the Bedrock API as an unknown model parameter:

```
UserWarning: WARNING! streaming is not default parameter.
                streaming was transferred to model_kwargs.
```

So today `streaming=True` is a misleading no-op for the streaming behavior the user intended.

## Fix

Add a `streaming: bool = False` field, matching the established sibling convention — `langchain-anthropic` (`streaming: bool = False`) and `langchain-aws`'s own `BedrockBase` (`llms/bedrock.py`) both declare it. LangChain Core already honors a `streaming` attribute **when the subclass declares it**: `BaseChatModel._should_stream` / `_streaming_disabled` gate on `"streaming" in self.model_fields_set`. So this is no longer a silent no-op, and `disable_streaming` (per-model capability auto-config) still takes precedence and is untouched.

## Before / After

| Item | Before | After |
|------|--------|-------|
| `ChatBedrockConverse(..., streaming=True)` | ⚠️ swallowed into `additional_model_request_fields` + `UserWarning` (or `extra_forbidden` on older versions) | ✅ captured as a first-class `streaming` field Core honors |
| `test_init_streaming` | ❌ `xfail` marker | ✅ passes (marker removed) |
| `disable_streaming` auto-config (`set_disable_streaming`) | unchanged | ✅ unchanged |
| Public API / arguments / ordering | — | ✅ unchanged (new field defaults to `False`) |
| `test_serdes` xfail | xfail | ✅ still xfail (out of scope, pending core) |

## Tests

Removed the `test_init_streaming` xfail marker and added a regression test `test_streaming_init_param` that asserts `streaming=True` becomes a real field (in `model_fields_set`, not in `additional_model_request_fields`) and that the default leaves `disable_streaming` auto-config intact.

Red (fix reverted, test kept):

```
FAILED tests/unit_tests/chat_models/test_bedrock_converse.py::test_streaming_init_param
E   AttributeError: 'ChatBedrockConverse' object has no attribute 'streaming'
```

Green (with fix):

```
275 passed, 1 xfailed in 3.72s    # full test_bedrock_converse.py (1 xfailed = test_serdes)
909 passed, 1 skipped, 1 xfailed  # full unit suite
```

`make lint` (ruff + `ruff format --diff` + mypy) passes clean.

## Areas worth a careful look

- Whether you'd prefer this minimal `streaming` field vs. mapping `streaming` → `disable_streaming` internally. I went with the field to match the sibling convention and to keep Core's existing handling authoritative, but happy to adjust to your preference.

Addresses #236 (the `ChatBedrockConverse` `test_init_streaming` xfail).

---

This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, its rationale, and the test results before submitting.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -557,6 +557,14 @@ class ChatBedrockConverse(BaseChatModel):
 
     """
 
+    streaming: bool = False
+    """Whether to stream the results or not.
+
+    Kept for backward compatibility with ``ChatBedrock``. When ``True``, LangChain
+    Core will use the streaming path for ``invoke``/``ainvoke`` (subject to
+    :attr:`disable_streaming`, which still governs per-model capability).
+    """
+
     endpoint_url: Optional[str] = Field(default=None, alias="base_url")
     """Needed if you don't want to default to us-east-1 endpoint"""
 
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -558,12 +558,7 @@ class ChatBedrockConverse(BaseChatModel):
     """
 
     streaming: bool = False
-    """Whether to stream the results or not.
-
-    Kept for backward compatibility with ``ChatBedrock``. When ``True``, LangChain
-    Core will use the streaming path for ``invoke``/``ainvoke`` (subject to
-    :attr:`disable_streaming`, which still governs per-model capability).
-    """
+    """Whether to stream the results or not."""
 
     endpoint_url: Optional[str] = Field(default=None, alias="base_url")
     """Needed if you don't want to default to us-east-1 endpoint"""
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -786,61 +786,22 @@ def test_set_disable_streaming(
 
 
 def test_streaming_init_param() -> None:
-    model_id = "anthropic.claude-fable-5"
+    model_id = "anthropic.claude-sonnet-4-6"
 
-    def _make(**kwargs: Any) -> Tuple[ChatBedrockConverse, Any]:
-        client = mock.MagicMock()
-        client.converse.return_value = {
-            "output": {"message": {"role": "assistant", "content": [{"text": "hi"}]}},
-            "stopReason": "end_turn",
-            "usage": {"inputTokens": 1, "outputTokens": 1, "totalTokens": 2},
-            "ResponseMetadata": {},
-        }
-        client.converse_stream.return_value = {
-            "stream": [
-                {"messageStart": {"role": "assistant"}},
-                {
-                    "contentBlockDelta": {
-                        "delta": {"text": "hi"},
-                        "contentBlockIndex": 0,
-                    }
-                },
-                {"contentBlockStop": {"contentBlockIndex": 0}},
-                {"messageStop": {"stopReason": "end_turn"}},
-                {
-                    "metadata": {
-                        "usage": {
-                            "inputTokens": 1,
-                            "outputTokens": 1,
-                            "totalTokens": 2,
-                        }
-                    }
-                },
-            ]
-        }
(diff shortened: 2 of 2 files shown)


libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
```diff
@@ -789,6 +785,31 @@ def test_set_disable_streaming(
...
+    assert "streaming" in llm.model_fields_set
+    assert "streaming" not in (llm.additional_model_request_fields or {})
+
+    # Defaults to False and leaves disable_streaming auto-config intact.
+    default = ChatBedrockConverse(
+        model="anthropic.claude-3-sonnet-20240229-v1:0",
+        region_name="us-west-2",
+    )
+    assert default.streaming is False
+    assert default.disable_streaming is False
```
michaelnchin: In general, it's preferable for the tests to check the executable behavior rather than just looking at the plumbing (i.e. we're testing what we already know what BaseModel sets when given the `streaming` param). `_should_stream` would be a better choice to check the output of as it's called at invoke time.

```suggestion
    model_id = "anthropic.claude-fable-5"

    streaming = ChatBedrockConverse(
        model=model_id, region_name="us-west-2", streaming=True
    )
    assert streaming._should_stream(async_api=False) is True

    default = ChatBedrockConverse(model=model_id, region_name="us-west-2")
    assert default._should_stream(async_api=False) is False

    disabled = ChatBedrockConverse(
        model=model_id,
        region_name="us-west-2",
        streaming=True,
        disable_streaming=True,
    )
    assert disabled._should_stream(async_api=False) is False
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -557,6 +557,14 @@ class Joke(BaseModel):
 
     """
 
+    streaming: bool = False
+    """Whether to stream the results or not.
+
+    Kept for backward compatibility with ``ChatBedrock``. When ``True``, LangChain
+    Core will use the streaming path for ``invoke``/``ainvoke`` (subject to
+    :attr:`disable_streaming`, which still governs per-model capability).
+    """
```
michaelnchin: Remove/make same as BedrockBase
```suggestion
    """Whether to stream the results or not."""
```

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
```diff
@@ -789,6 +785,31 @@ def test_set_disable_streaming(
     assert llm.disable_streaming == disable_streaming
 
 
+def test_streaming_init_param() -> None:
+    """``streaming=True`` is accepted as a backward-compatible init param.
+
+    It must be captured as a first-class field that LangChain Core honors (i.e.
+    present in ``model_fields_set``), not silently routed into
+    ``additional_model_request_fields`` as an unknown model parameter.
+    """
```
michaelnchin: Remove docstring
```suggestion
```

michaelnchin (commented):
Thanks @kimnamu ! Adding a couple changes.

michaelnchin (approved):
Thanks @kimnamu !

Issue #29, opened 2024-04-26
Exposing more boto3 client options to ChatBedrock and doc example?


P1llus:
With the addition of the opus model support on Bedrock, which generally tends to always hit the default timeout options, would it be okay to add the different retry/timeout options from boto3 as common options to ChatBedrock?

Currently we have support for a few common keys here: https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/llms/bedrock.py#L300.

However just changing from `Claude 2/2.1` or `Claude 3 Sonnet` to `Opus` increases the runtime mostly by 10x (mine are consistently around 20-35 seconds to **200+** seconds)..

I would assume that once more people start using this new library, and depending on the popularity of Opus, a few changes are recommended:

1. Adding a higher default timeout, to prevent common issues to appear.
2. Adding more boto3 client options available as arguments to ChatBedrock/BedrockBase, so that developers can instantly see them as a supported list of arguments.
3. A small documentation example that links to the available boto3 options that would work with Bedrock, and an example on how to create a custom `Botocore.config.Config()` object that can be passed to the existing `config` argument would be greatly appreciated. I did get it working through trial and error, but only from boto3 examples directly and not from langchain issues or docs.

3coins:
@P1llus 
All great ideas, there is plan to abstract out the client creation to a separate module with more options. Note that users can create their own client and pass in to the Bedrock classes, which supports all options available with boto3. Defaulting to a higher timeout might make sense for Opus, but not applicable universally. If there is a config that you have worked out which can work universally, open a PR with that change, and we can review it. 
Agree that we need more examples and documentation, we welcome contributions from the community, and let us know if you need any help with this.

P1llus:
@3coins Ah yeah that is true about the defaults, might be a bit overkill to change it like that.

I think something in-between would be if its possible for us to detect the underlying issue with some sort of logging.
As mentioned above, the only message received from Chat bedrock is that there was a timeout. 
However it was actually hitting a throttle and 504's with a backoff.
There was no indication on that from the client perspective, and only the logs from the bedrock service itself would highlight it.

I am unsure if this is how boto3 logs it as well or if its the ChatBedrock class that don't surface information like that?

3coins:
Yes, totally makes sense, exception handling is not the best at the moment and we can improve that. I have to check if there is a separate exception thrown by the API that we can surface here.

PR #1123, opened 2026-06-23
docs: add Bedrock timeout and retry example

WOLIKIMCHENG:
## Summary
- add a README example for configuring Bedrock request timeouts and retries
- show both `timeout` / `max_retries` and a `botocore.config.Config` option
- clarify that these settings apply only to the configured model instance

Related to #29.

## Testing
- `git diff --check`
- `python .github/scripts/check_diff.py README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,32 @@ response = model.invoke("Hello! How are you today?")
 print(response)
 ```
 
+You can configure Bedrock request timeouts and retries for longer-running calls.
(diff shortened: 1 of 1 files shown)


michaelnchin (commented):
Thanks for the submission @WOLIKIMCHENG - IMO Boto3 config is too specific for the README, where we should only have very simple examples.

PR #1152, opened 2026-07-07
fix(langgraph-checkpoint-aws): resolve actor_id for nested subgraph get_state (#733)

### LINKED ISSUE #733: AgentCoreMemorySaver: Exception when using get_state(subgraphs=True) with nested graphs
When using `AgentCoreMemorySaver` and calling `get_state(config, subgraphs=True)` on a graph that has nested graphs, I get the following error. Note that without subgraphs=True it works as expected.

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File ".venv/lib/python3.12/site-packages/langgraph/pregel/main.py", line 1264, in get_state
    return self._prepare_state_snapshot(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/langgraph/pregel/main.py", line 1066, in _prepare_state_snapshot
    task_states[task.id] = subgraphs[task.name].get_state(
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/langgraph/pregel/main.py", line 1263, in get_state
    saved = checkpointer.get_tuple(config)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/langgraph_checkpoint_aws/agentcore/saver.py", line 75, in get_tuple
    checkpoint_config = CheckpointerConfig.from_runnable_config(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".venv/lib/python3.12/site-packages/langgraph_checkpoint_aws/agentcore/models.py", line 40, in from_runnable_config
    raise InvalidConfigError(
langgraph_checkpoint_aws.agentcore.constants.InvalidConfigError: RunnableConfig must contain 'actor_id' for AgentCore Checkpointer
```

Config:
```
RunnableConfig(
    configurable={
        "thread_id": thread_id,
        "actor_id": ACTOR_ID,
    },
    recursion_limit=RECURSION_LIMIT,
)
```

islamelkadi:
## Summary

`AgentCoreMemorySaver.get_state(config, subgraphs=True)` raised `InvalidConfigError: RunnableConfig must contain 'actor_id'` on any graph containing nested subgraphs, so subgraph state could not be inspected. LangGraph derives a per-subgraph `RunnableConfig` that carries `thread_id` (and `checkpoint_ns`) but **not** `actor_id`, and the saver required `actor_id` on every call. After this change, a subgraph read inherits the `actor_id` from the current request's parent read, so `get_state(subgraphs=True)` works on nested (and deeply nested) graphs.

Resolves #733.

## How it works

> **Terminology.** *Fail closed* means: if the identity (`actor_id`) can't be established, the operation is **refused** (raises `InvalidConfigError`) rather than proceeding with a guess.

Resolution order for a read is: explicit config `actor_id` → request-scoped captured actor (same `thread_id`) → raise.

1. `models.py` — `CheckpointerConfig.from_runnable_config` gains a keyword-only `default_actor_id`. It resolves `actor_id` as `configurable.get("actor_id") or default_actor_id`, still checking `thread_id` first. This is a pure parser mechanism; it does not decide where the fallback comes from.
2. `saver.py` — a module-level `ContextVar` (`_request_actor`) records the `(thread_id, actor_id)` seen on a read. `_resolve_actor_id` captures that pair when a read supplies both, and lets a later read that omits `actor_id` inherit it **only when** (a) the `thread_id` matches and (b) the read is a derived subgraph read (non-empty `checkpoint_ns`). Otherwise it returns `None` and the parser fails closed.
3. Reads (`get_tuple`, `list`) pass `default_actor_id=self._resolve_actor_id(config)`. Writes (`put`, `put_writes`, `put_with_writes`) pass no fallback — they stay fail-closed, exactly as in the released version.
4. Async reads (`aget_tuple`, `alist`) prime the capture on the event loop, then dispatch via `langchain_core`'s `run_in_executor`, which copies the context into the worker thread so the sync body resolves the captured actor.

## Design decision: request-scoped captures the `(thread_id, actor_id)` instead of a saver-level fallback

The capture lives in a `ContextVar` (isolated per thread / async task), never on the saver instance, so a single saver can serve many actors concurrently without one request's actor leaking into another's. Inheritance is gated on both a matching `thread_id` and a non-empty `checkpoint_ns`, so only genuine subgraph reads can inherit; a top-level read that omits `actor_id` fails closed.

**Alternatives considered.** The first approach added an `actor_id` argument to the saver's constructor and used it as the fallback whenever a config lacked one. It fixed the bug, but the fallback was stored on the saver *object*, so it was a single fixed value for
the life of that saver. That breaks down when one saver instance serves multiple users (a multi-tenant setup where the user is passed per request): the fixed value is only correct for one user, so a subgraph read for user B — which arrives without an `actor_id` — would fall back to whatever the saver was built with, potentially user A, and read user A's data. The request-scoped approach drops the constructor argument entirely and instead takes the fallback from the *current request* (captured per request via a `ContextVar`), so it is always the requesting user. And because nothing is stored on the saver, there is no saver-wide `actor_id` that a deployment could set once and have wrongly applied to every other user's requests (see the ccomments for further details).

## Scope

- Fixes the read path (`get_tuple` / `aget_tuple`, `list` / `alist`) — the path `get_state(subgraphs=True)` uses. Writes are unchanged from the release.

## Trade-offs / Areas needing careful review

- **Coupling to a LangGraph internal (please review):** the guard assumes "non-empty `checkpoint_ns` ⇒ derived subgraph read" (and that top-level reads carry their own `actor_id`). This matches how LangGraph derives per-subgraph configs today — [`_prepare_state_snapshot` always builds a non-empty `task_ns`](https://github.com/langchain-ai/langgraph/blob/1.2.5/libs/langgraph/langgraph/pregel/main.py#L1203-L1221) (langgraph 1.2.5). If a future LangGraph changed that convention, the failure mode is safe (a subgraph read would stop inheriting and fail closed, i.e. #733 would resurface) rather than a leak, but it's worth re-checking on LangGraph upgrades.

## Test coverage

**Unit tests** (`tests/unit_tests/agentcore/test_saver.py`, tests pass) :

Created:
- `TestCheckpointerConfig.test_from_runnable_config_resolves_default_actor_id` — fallback resolution
- `TestCheckpointerConfig.test_config_actor_id_takes_precedence_over_default` — explicit config wins
- `TestCheckpointerConfig.test_missing_thread_id_checked_before_actor_id` — validation order
- `test_get_tuple_resolves_fallback_actor_id` — subgraph read inherits the parent's actor
- `test_get_tuple_resolves_fallback_for_deeply_nested_subgraphs` — 3 levels
- `test_get_tuple_session_mismatch_does_not_leak` — different `thread_id` fails closed
- `test_get_tuple_top_level_read_does_not_inherit_actor_id` — empty-ns read fails closed
- `test_aget_tuple_resolves_fallback_actor_id` — async inheritance across the executor hop
- `test_request_actor_isolated_across_threads` — concurrent same-`thread_id` requests don't share actors
- `test_get_tuple/list/put/put_writes/put_with_writes_without_actor_id_raises` — fail-closed paths
- autouse fixture resetting the `ContextVar` between tests

**Integration tests** (`tests/integration_tests/agentcore/test_saver.py`; live AgentCore
Memory, skipped unless `AGENTCORE_MEMORY_ID` is set, tests pass):

Created:
- `test_get_state_with_subgraphs_resolves_actor_id` — end-to-end #733 repro, single level
- `test_get_state_with_deeply_nested_subgraphs_resolves_actor_id` — 3 nested levels
- `test_aget_state_with_subgraphs_resolves_actor_id` — async variant (skipif Python < 3.11, where LangGraph `interrupt()` is unsupported in an async context).

## Final callouts

`AgentCoreValkeySaver` has the same structural exposure (`actor_id` is a required field and part of every storage key) and should probably adopt the same mechanism. I'll do an additional offline analysis, and if warranted, I'll track it as a follow-up (out of scope here for now).

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore/models.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore/models.py
@@ -48,13 +48,16 @@ class CheckpointerConfig(BaseModel):
             config: RunnableConfig whose ``configurable`` dict supplies
                 ``thread_id`` and (normally) ``actor_id``.
             default_actor_id: Fallback actor id used only when ``config`` omits
-                ``actor_id`` (e.g. a derived subgraph config). Not yet honored.
+                ``actor_id``. Lets a derived subgraph config, which carries
+                ``thread_id`` but not ``actor_id``, inherit the actor from the
+                saver instead of failing.
 
         Returns:
             The parsed CheckpointerConfig.
 
         Raises:
-            InvalidConfigError: If ``thread_id`` or ``actor_id`` is missing.
+            InvalidConfigError: If ``thread_id`` is missing, or if ``actor_id`` is
+                missing and no fallback resolves it.
         """
         from .constants import InvalidConfigError
 
@@ -65,15 +68,15 @@ class CheckpointerConfig(BaseModel):
                 "RunnableConfig must contain 'thread_id' for AgentCore Checkpointer"
             )
 
-        # TODO(#733): resolve a missing actor_id from default_actor_id.
-        if not configurable.get("actor_id"):
+        actor_id = configurable.get("actor_id") or default_actor_id
+        if not actor_id:
             raise InvalidConfigError(
                 "RunnableConfig must contain 'actor_id' for AgentCore Checkpointer"
             )
 
         return cls(
             thread_id=configurable["thread_id"],
-            actor_id=configurable["actor_id"],
+            actor_id=actor_id,
             checkpoint_ns=configurable.get("checkpoint_ns", ""),
             checkpoint_id=configurable.get("checkpoint_id"),
         )
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore/models.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore/models.py
@@ -39,8 +39,23 @@ class CheckpointerConfig(BaseModel):
         return self.thread_id
 
     @classmethod
-    def from_runnable_config(cls, config: dict[str, Any]) -> "CheckpointerConfig":
-        """Create CheckpointerConfig from RunnableConfig."""
+    def from_runnable_config(
+        cls, config: dict[str, Any], *, default_actor_id: str | None = None
+    ) -> "CheckpointerConfig":
+        """Create a CheckpointerConfig from a RunnableConfig.
+
+        Args:
+            config: RunnableConfig whose ``configurable`` dict supplies
+                ``thread_id`` and (normally) ``actor_id``.
+            default_actor_id: Fallback actor id used only when ``config`` omits
+                ``actor_id`` (e.g. a derived subgraph config). Not yet honored.
+
+        Returns:
+            The parsed CheckpointerConfig.
+
+        Raises:
+            InvalidConfigError: If ``thread_id`` or ``actor_id`` is missing.
+        """
         from .constants import InvalidConfigError
 
         configurable = config.get("configurable", {})
@@ -50,6 +65,7 @@ class CheckpointerConfig(BaseModel):
                 "RunnableConfig must contain 'thread_id' for AgentCore Checkpointer"
             )
 
+        # TODO(#733): resolve a missing actor_id from default_actor_id.
         if not configurable.get("actor_id"):
             raise InvalidConfigError(
                 "RunnableConfig must contain 'actor_id' for AgentCore Checkpointer"
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore/saver.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore/saver.py
@@ -4,7 +4,6 @@ AgentCore Memory Checkpoint Saver implementation.
 
 from __future__ import annotations
 
-import asyncio
 import random
 from collections.abc import AsyncIterator, Iterator, Sequence
 from contextvars import ContextVar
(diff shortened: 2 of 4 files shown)


islamelkadi:
### Design pivot details: from a static fallback to request-scoped resolution

Sharing how this design evolved, since the final approach is deliberately more involved than the first cut.

**1. The bug.** `get_state(config, subgraphs=True)` failed because LangGraph derives a per-subgraph config that keeps `thread_id` but drops `actor_id`, and the saver required `actor_id` on every call. Each nested lookup raised `InvalidConfigError`.

**2. First implementation — a saver-level `actor_id` fallback.** My minimal fix added an `actor_id` parameter to the saver and used it as the fallback for any config missing one. It resolved the bug and was backward compatible (default `None` → unchanged, fail-closed).

**3. Why I eventually moved away from that approach — a multi-tenant security concern.** The fallback lived on the saver *instance*. For a per-user/per-session saver that's fine. But for a **shared, multi-tenant** saver (one instance serving many users, identity supplied per request), a single fixed fallback is correct for only one user. During subgraph reads — which lack `actor_id` — the fallback would resolve to that one fixed actor regardless of who is actually asking, so a request for user B could read under user A's identity. Crucially, this isn't caught by the default AWS posture: a typical deployment uses one application IAM identity with access to the whole memory resource, and AWS's shared-responsibility model puts session-to-user mapping on the application. So the saver-level fallback approach turns a safe-but-annoying error into a silent wrong-actor read.

**4. Final implementation — request-scoped `ContextVar`.** Instead of storing the actor on the saver, I capture `(thread_id, actor_id)` in a `ContextVar` that is isolated per thread / async task. A subgraph read inherits the actor **only** from its own request's parent read, and **only** when the `thread_id` matches. One shared saver can now serve many actors concurrently with no cross-request leakage (covered by a thread-isolation test and a `thread_id`-mismatch test).

`ContextVar` is also a natural fit for the data already in play. The identity to carry is just `(thread_id, actor_id)` — already an immutable `tuple` in the implementation — and that pairs cleanly with how `ContextVar` works. A `ContextVar` isolates the per-context *binding*, but `copy_context()` (used to carry the capture into the executor thread) copies value *references*, not deep copies; a mutable value (a `dict`/`list`) would be shared across the copied contexts, so a mutation in one thread/task could surface in another — a subtle cross-context leak. Because the captured value is an immutable tuple, there is nothing to mutate: the only way to change it is `_request_actor.set(...)` with a brand-new tuple, which rebinds the variable in the current context alone. So storing the existing tuple in a `ContextVar` gives request isolation and closes the aliasing hazard by construction — no extra wrapping or defensive copying needed.

**5. One more hardening — the `checkpoint_ns` guard.** I then restricted inheritance to reads with a non-empty `checkpoint_ns` (i.e. genuine derived subgraph reads — in the current LangGraph, [`_prepare_state_snapshot` builds each per-subgraph config with a non-empty `task_ns`](https://github.com/langchain-ai/langgraph/blob/1.2.5/libs/langgraph/langgraph/pregel/main.py#L1203-L1221) (langgraph 1.2.5), so they always carry a namespace). A *top-level* read that omits `actor_id` — only possible if a caller leaves it out — now fails closed instead of borrowing a possibly stale captured actor, narrowing the residual surface further. (This relies on that LangGraph behavior; the PR body flags it as a coupling to re-check on upgrades.)

PR #1153, opened 2026-07-08
fix(aws): Correctly handle omitted-display thinking blocks on Claude 4.7+

michaelnchin:
While working on migrating `ChatBedrock` thinking integration tests from Claude 4.x to Claude 5, Invoke API would sometimes throw errors noting that the model returned empty thinking blocks:
```
`ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: messages.1.content.0.thinking.thinking: Field required`   
```

This is because adaptive-only thinking Claude models (specifically 4.7+) default to `omitted` [thinking display mode](https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking#summarized-thinking), which intentionally returns thinking blocks empty (except for the signature). However, we can't pass these empty thinking blocks back to Invoke API as-is, at it considers them invalid without the `thinking` field.

Fixing this by updating the thinking block processing logic in `_format_anthropic_messages` to ensure that we add a dummy `"thinking": ""` field to these empty thinking blocks before the round-trip back to Invoke API.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -638,9 +638,16 @@ def _format_anthropic_messages(
                             )
                     elif item["type"] in ["thinking", "redacted_thinking"]:
                         # Store thinking blocks separately
-                        thinking_blocks.append(
-                            {k: v for k, v in item.items() if k != "index"}
-                        )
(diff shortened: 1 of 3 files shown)


michaelnchin (commented):
Integ CI (from main) passing: https://github.com/langchain-ai/langchain-aws/actions/runs/28975481471

PR #1157, opened 2026-07-11
fix(aws): Make ChatBedrock streaming with Claude 5 produce consistent block content

michaelnchin:
Fixing a stochastic (~75% success rate) failure with `TestBedrockStandard::test_astream` in live integ CI runs after #1154 (https://github.com/langchain-ai/langchain-aws/actions/runs/29048020685/job/86221390451):
```
          text_blocks = [b for b in full.content_blocks if b["type"] == "text"]
  >       assert len(text_blocks) == 1
  E       assert 0 == 1
```

As is, ChatBedrock's stream response processor will skip string coercion of text delta blocks if thinking is enabled, which it was able to positively determine with older models [by directly checking the request parameters](https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/llms/bedrock.py#L1298-L1299). 

HOWEVER, for Claude 5 models, the presence of thinking is nondeterministic with the default adaptive thinking mode, which can't be disabled. So, as is, we might end up with mixed plain string and dict thinking blocks in the final processed stream content when these blocks dodge the check that's supposed to prevent this from happening.

Fixing this by adding an extra conditional to run the `thinking_on_by_default` util check for adaptive-thinking-only models, ensuring that Bedrock streaming with Claude 5 models always returns a consistent set of dict-format blocks.

### DIFF
--- a/libs/aws/langchain_aws/llms/bedrock.py
+++ b/libs/aws/langchain_aws/llms/bedrock.py
@@ -38,6 +38,7 @@ from langchain_aws.utils import (
     enforce_stop_tokens,
     get_num_tokens_anthropic,
     get_token_ids_anthropic,
+    thinking_disabled_in_params,
     thinking_in_params,
     thinking_on_by_default,
 )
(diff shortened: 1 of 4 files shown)


michaelnchin (commented):
Integ CI passing on branch: https://github.com/langchain-ai/langchain-aws/actions/runs/29295965024/job/86969418952

Issue #1158, opened 2026-07-13
ChatBedrockConverse: fragile provider parsing + silent disable_streaming fallback keep regenerating per-model streaming bugs


hanjoonchoe:
### Submission checklist

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Package (Required)

langchain-aws

### Related Issues / PRs

#354, #395, #489, #544, #573, #1139 — all instances of the same bug class, each closed with a narrow allowlist/matcher edit. This issue is about the mechanism that keeps producing them.

### Reproduction Steps / Example Code (Python)

No AWS credentials needed — the misparse and the silent flag are pure client-side logic:

```python
from langchain_aws import ChatBedrockConverse

# Problem 1: provider parsing breaks on dots in the version segment
model_id = "minimax.minimax-m2.5"
parts = model_id.split(".")
print(parts[-2] if len(parts) > 1 else parts[0])  # "minimax-m2" — wrong; should be "minimax"

# Problem 2: the misparsed/unknown provider silently disables streaming
m = ChatBedrockConverse(model_id=model_id, region_name="us-east-1")
print(m.disable_streaming)  # True — no warning was emitted
# .astream() now "works" but yields exactly ONE chunk with the full response
```

### Error Message and Stack Trace (if applicable)

```shell
No error is raised — that is the core of the problem. .stream() / .astream()
silently degrade to a single chunk containing the fully-generated response
(complete content, full usage_metadata, stopReason "end_turn"). For SSE-based
products this is a severe, hard-to-diagnose regression: tokens stop flowing and
users wait in silence, with no exception or log line pointing at the cause.
```

### Description

[...]

PR #1160, opened 2026-07-14
feat(aws): Log helpful warning when ChatBedrockConverse auto-disables streaming for unsupported models

michaelnchin:
Addresses point 2 from #1158.

Updating `ChatBedrockConverse` to add a clarifying logger warning when streaming is auto-disabled for models that haven't been whitelisted. The warning will also point users towards the `disable_streaming=False` override parameter (which should reduce the need to manually provide this guidance on issues related to new model release support).

As part of this, also refactored the streaming configuration logic to have both regular model IDs and AIPs use a shared `_resolve_disable_streaming` helper method to set the final support level.


michaelnchin (commented):
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/29363203863/job/87188220698

PR #1162, opened 2026-07-14
feat(aws): Update ChatBedrockConverse for simplified Anthropic cache management, preserve cache_control across bind_tools

billcai:
## Why

Two prompt-caching bugs in `ChatBedrockConverse` (Bedrock Converse API), both of which silently reduce cache hit rates rather than erroring — so they're easy to miss until a bill or latency review surfaces them.

### 1. Cache checkpoint placed on the wrong message (fixed-instruction shape)

`_apply_cache_points` placed a `cachePoint` on the **last** message. That's fine when the last message is a fresh user question, but it breaks a common agentic pattern: a **fixed instruction that always trails a growing history** — e.g. an intent-classifier node whose final turn is always *"classify the conversation above"*.

In that shape the checkpoint caches a prefix ending **with the instruction**. The next turn inserts new history *before* the instruction, so the cached prefix diverges right after the old history — only the system prompt gets reused, and the entire history is re-written every turn. Cache reads flatline.

This is subtle to observe: with a **large system prompt** the legacy placement also caches the system block on its own, which masks the lost history reuse. The bug only shows clearly when the reusable bulk lives in the **history** (small system + large history). Measured live (call-2 `cacheRead`, small system, growing history, fixed trailing instruction):

| Model | Legacy `multi` | New `auto` |
|---|---|---|
| Opus 4.8 | **0** | 8052 |
| Sonnet 4.6 | **0** | 5373 |
| Haiku 4.5 | **0** | 5381 |

**Fix:** for Anthropic Claude models, the new default `cache_strategy="auto"` places a **single** checkpoint at the **end of the conversation history** (the penultimate message), which is also what enables Bedrock's [simplified cache management](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-simplified) (auto lookback across prior content-block boundaries). This keeps `system + history` cache-readable as turns accumulate, and is neutral-to-positive for the standard "new question last" shape. Amazon Nova and other providers keep the existing multi-checkpoint placement unchanged.

A new keyword-only field controls this:

```python
cache_strategy: Literal["auto", "single", "multi"] = "auto"
```

- `"auto"` (default): single end-of-history checkpoint for Claude; legacy multi-point for Nova/others.
- `"single"`: force single-checkpoint placement for any provider.
- `"multi"`: restore the previous multi-checkpoint placement exactly (escape hatch).

### 2. `bind(cache_control=...).bind_tools(...)` silently drops `cache_control`

```python
# Before this PR: cache_control is lost, zero cachePoints sent
model.bind(cache_control={"type": "ephemeral", "ttl": "5m"}).bind_tools([...])
```

`.bind(...)` returns a `RunnableBinding`, which has no `bind_tools`. The call is proxied via `__getattr__` to the **underlying model's** `bind_tools`, whose `self` is the bare model — it rebuilds a fresh binding and never carries the binding's kwargs. (The reverse order, `.bind_tools(...).bind(cache_control=...)`, already worked.)

**Fix:** `ChatBedrockConverse.bind()` now returns a small `_ChatBedrockConverseBinding` whose `bind_tools` delegates to the model and then re-merges the binding's own kwargs (tool kwargs win on conflict). The two now compose in **any order**, and *all* previously-bound kwargs are preserved, not just `cache_control`.

## Compatibility

- Additive, keyword-only field — no signature changes. Default `"auto"` ships the placement fix transparently; `"multi"` restores prior behavior byte-for-byte.
- Non-Anthropic / non-Nova providers are unchanged under `"auto"`.
- `bind()` override returns a `_ChatModelBinding` subclass, preserving `lc_id`/serialization (`"RunnableBinding"`) and the typed `stream_events`/`astream_events` overloads.

## Testing

- **Unit** (`tests/unit_tests/chat_models/test_bedrock_converse.py`): placement across shapes (fixed-instruction vs new-question-last), strategy overrides, idempotency, Nova/other-provider preservation, and the `bind`/`bind_tools` composition in both orders + conflict resolution + end-to-end cachePoint emission. The `bind_tools` regression tests fail if the `bind()` override is removed.
- **Integration** (VCR, `tests/integration_tests/middleware/test_prompt_caching.py`): recorded live against **Opus 4.8, Sonnet 4.6, Haiku 4.5** and replay offline — request-shape assertions, cache reuse for both the fixed-instruction and standard growing-conversation shapes, a direct `multi`-vs-`auto` contrast proving the asymmetry, and `bind(cache_control).bind_tools()` reuse. Cassettes are region-agnostic (per #1156).
- `make lint`, `ruff format`, and `mypy` clean — **zero new** mypy errors vs the base.

## Review focus

- `_apply_single_cache_point` placement preference (penultimate → system → tools) and the idempotency short-circuit.
- The `bind()` override / `_ChatBedrockConverseBinding` — it depends on the private `langchain_core...._ChatModelBinding` (present since the pinned `langchain-core>=1.4.7`, guarded with a scoped `# type: ignore[attr-defined]`).

---

*This PR was developed with the assistance of an AI agent*


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -143,6 +147,37 @@ def _infer_region_name_from_client(client: Optional[Any]) -> Optional[str]:
...
+    conflict, matching the behavior of chaining ``bind`` calls.
+    """
+
+    def bind_tools(
+        self,
+        tools: Sequence[Union[Dict[str, Any], TypeBaseModel, Callable, BaseTool, str]],
+        **kwargs: Any,
+    ) -> Runnable[LanguageModelInput, AIMessage]:
+        """Bind tools while preserving kwargs already bound to this binding."""
+        tool_binding = self.bound.bind_tools(tools, **kwargs)
```
michaelnchin:
```suggestion
        model = cast("ChatBedrockConverse", self.bound)
        tool_binding = model.bind_tools(tools, **kwargs)
```

libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
```diff
@@ -5329,6 +5330,18 @@ def test_response_format_stream_without_response_format() -> None:
     assert "outputConfig" not in call_kwargs
 
 
+def _count_cache_points(*sections: list) -> int:
```
michaelnchin:
```suggestion
def _count_cache_points(*sections: Iterable[Any]) -> int:
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -678,14 +746,106 @@ def create_cache_point(cls, cache_type: str = "default") -> Dict[str, Any]:
...
+            any(_is_cache_point(b) for b in system)
+            or any(_is_cache_point(t) for t in (tools or []))
+            or any(
+                isinstance(m.get("content"), list)
+                and any(_is_cache_point(b) for b in m["content"])
+                for m in bedrock_messages
+            )
+        )
+        if already_cached:
+            return
```
michaelnchin, line 809: `cache_control` is getting discarded silently here, should log a warning for visibility (ex. user left a manual checkpoint in accidentally):
```suggestion
        if already_cached:
            logger.debug(
                "Ignoring cache_control settings, as the request already contains "
                "a cachePoint."
            )
            return
```

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -678,14 +746,106 @@ def create_cache_point(cls, cache_type: str = "default") -> Dict[str, Any]:
...
+
+        # Anchor the checkpoint at the end of the conversation history: the
+        # penultimate message, just before the final (most dynamic) message.
+        if len(bedrock_messages) >= 2:
+            penultimate_content = bedrock_messages[-2].get("content")
```
michaelnchin, line 814: One possible case worth mentioning is that the message history could end with a tool call -> tool result, which together are almost certainly stable (i.e. the agent loop's not going to insert any content in between). In this case we would definitely want to move the checkpoint to the very back instead of at -2 position, so that the tool result is included in the cache prefix:

Should we add a check for this here, i.e. :
```suggestion
        def _has_tool_result(message: Dict[str, Any]) -> bool:
            content = message.get("content")
            return isinstance(content, list) and any(
                isinstance(b, dict) and "toolResult" in b for b in content
            )
        # Anchor the checkpoint at the end of the conversation history: the
        # penultimate message, just before the final (most dynamic) message.
        # If we end with a tool result, instead place the cache at the very
        # end, since it's assumed that nothing will be inserted between it
        # and the corresponding toolUse block.
        if bedrock_messages and _has_tool_result(bedrock_messages[-1]):
            bedrock_messages[-1]["content"].append(cache_block)
            return
        if len(bedrock_messages) >= 2:
            penultimate_content = bedrock_messages[-2].get("content")
            if isinstance(penultimate_content, list):
                penultimate_content.append(cache_block)
                return
```

michaelnchin (commented):
Thanks for working on this @billcai ! Tested manually and looks great overall - couple of minor comments.

billcai:
hey @michaelnchin, thanks for the review. I was on the same train of thought with you re testing more comprehensive and common cases for prompt caching, and found regressions. The original commit made `auto` place a **single** cachePoint at the end of history. Head-to-head benchmarking on real Bedrock showed that regresses the common cases. `auto` now **keeps the legacy points (`system` + last message + `tools`) and *adds* an end-of-history point** — best of both, never worse than the old `multi` on cache reads.

## What changed vs. the first commit

| | First commit (`07b008f`) | This commit (`c8ae6cd`) |
|---|---|---|
| `auto` (Claude) | single point at end of history | `system` + last msg + `tools` **+** end-of-history (≤ 4, Claude's max) |
| `single` | — (was the `auto` behavior) | explicit opt-in: end-of-history only |
| `multi` | legacy 3-point | unchanged |
| Nova / other providers | unchanged | unchanged |
| Manual cachePoint present | short-circuit | `auto` degrades to legacy 3-point (stay ≤ 4) |

`bind_tools`/`cache_control` fix from the first commit is untouched.

## Why the single-point default was wrong

Benchmarked `auto` (single) vs. legacy `multi` on real Bedrock, cumulative `cacheRead` over a growing conversation, per-run nonce so caches don't cross-contaminate. Single-point loses in 4 of 5 scenarios because it **never caches the last message** (so the newest turn is re-processed next turn) and **caches nothing on a single-message request** (no penultimate → falls back to a small system prompt):

| Scenario | `multi` | `auto` (single) | verdict |
|---|---|---|---|
| A — standard chat, new question last | 4663 | 1875 | single −60% |
| B — big final user message each turn | 5491 | 1840 | single −66% |
| C — single message, context in user turn (RAG/classify) | 1840 | **0** | single caches nothing |
| D — huge system + tools, short agent chat | 3584 | 3239 | single −10% |
| E — fixed-instruction last (the TCB case) | **0** | 3048 | single wins |

Single only won the specific shape it was built for.

## Why Option 2 is the fix

Keep the legacy points (they're what win A–D) and *add* the end-of-history point (the only thing needed for E). Re-benchmarked with the real implementation — `auto` is now the read frontier: **`auto` ≥ `multi` in every scenario**, and it recovers history reuse in E where `multi` reads zero.

| Scenario | `multi` | **`auto` (Option 2)** |
|---|---|---|
| A — standard chat | 4656 | **4656** |
| B — big final message | 5506 | **5506** |
| C — single message | 1838 | **1838** |
| D — system + tools | 3586 | **3586** |
| E — fixed-instruction last | **0** | **3046** |

Confirmed on Opus 4.8, Sonnet 4.6, and Haiku 4.5.

## The one tradeoff

In the pure fixed-instruction shape (E), `auto` writes more than a single point would, because it keeps the last-message checkpoint that caches the (constant, never-reused) instruction tail each turn. Writes bill at 1.25× vs. reads at 0.1×, so reads dominate net savings and `auto` still beats `multi` on E overall — but a workload that is *only* fixed-instruction classification is marginally more write-efficient under `cache_strategy="single"`, which is why `single` stays as an opt-in.

## Implementation notes

- `_apply_cache_points`: `single` → end-of-history-only helper; `auto` (Anthropic, non-Nova) → legacy placement + penultimate point; `multi` / Nova / other providers → legacy placement, no end-of-history point.
- 4-checkpoint cap: when the request already carries a manual cachePoint, `auto` skips the extra end-of-history point and emits the legacy 3-point layout.
- `_apply_single_cache_point` is retained for `single` (penultimate → `system` → `tools` fallback).

## Testing

- Unit tests rewritten for the 4-point `auto` layout (single-message, forward-chat, fixed-instruction shapes; `single`/`multi` overrides; Nova/other preservation; idempotency; 4-cap when a manual point exists). `981 passed`.
- VCR integration tests re-recorded live for all three Claude models (request shape, reuse in the standard growing-conversation and fixed-instruction shapes, the `multi`-vs-`auto` contrast, and `bind(cache_control).bind_tools()` reuse). Replay offline; region-agnostic per #1156. The request-shape test was renamed `test_converse_single_cache_point_request_shape` → `test_converse_auto_cache_point_request_shape`, so its old cassettes were replaced.
- `mypy --all-groups`, `ruff check`, `ruff format` clean.

michaelnchin (commented):
@billcai, thanks for the revisions, and good catch on the single message edge case. 👍 I think my revision comments are moot now, especially for the last-message tool result since this will be covered by the retention of the old checkpoint at the very end of messages. 

Couple more thoughts...

1. `single` as an `cache_strategy` seems of limited use now, should we remove it? I ran the benchmarks myself with the single checkpoint vs. auto 4-checkpoints mode for Claude, and single performs same or worse for cache hits:
```
mode                multi   single    auto

fixed-instruction       0     2281    2281
standard chat        2292     2279    2289
single message       1235        0    1234
```
There's also the fact that it puts the one checkpoint at -2 by default, which might confuse users when it doesn't work for the single message edge-case that we just revised for. It's trivial enough that I think we should users manually insert the single checkpoint if they really want to.

2. More generally, do we really need the `cache_strategy` parameter? Preferably, ChatBedrockConverse should only expose (apart from BaseChatModel defaults) fields that correspond directly to Bedrock APIs or Boto configurations. I can't see a use case where the `auto` mode for Claude with the 4th checkpoint would be worse than the `multi` 3-point configuration. Should we just hardcode the new `auto` to optimize for Claude (and keep the current 3-point behavior the same for all other models)?

michaelnchin:
Hi @billcai, I've opened a new PR with a minimal implementation, based on my last review: https://github.com/langchain-ai/langchain-aws/pull/1192

Will close your PR and add you as a co-author on mine. Thanks!

PR #1164, opened 2026-07-14
fix(aws): escape stop sequences before building regex in enforce_stop_tokens

chuenchen309:
## Problem

`enforce_stop_tokens()` (in both `langchain_aws/utils.py` and the duplicated copy in `langchain_aws/llms/sagemaker_endpoint.py`) joins raw stop strings into a regex alternation pattern without escaping them:

```python
return re.split("|".join(stop), text, maxsplit=1)[0]
```

Any stop sequence containing regex metacharacters is silently misinterpreted instead of matched literally. This is especially damaging because the most common real-world stop sequences are LLM special tokens containing `|`, e.g. Llama 3's `<|eot_id|>` or `<|im_end|>` — these get parsed as regex alternation instead of a literal string, so output can be truncated at the wrong position. Stop sequences containing unbalanced parentheses (e.g. `"AI)"`) raise `re.error: unbalanced parenthesis`, crashing the call outright.

`BedrockLLM._call` (`llms/bedrock.py:1199`) uses this as its **sole** stop-sequence enforcement mechanism for the non-streaming path, so this is reachable directly from `BedrockLLM.invoke()`. The identical bug is independently duplicated in `sagemaker_endpoint.py`.

## Fix

`re.escape()` each stop string before joining into the pattern. Fixed both copies since they're the same defect.

## Testing

- `tests/unit_tests/test_utils.py` and `tests/unit_tests/llms/test_sagemaker_endpoint.py`: added tests covering (a) a stop sequence containing regex metacharacters (`|`) truncating at the correct literal position rather than an earlier metacharacter match, and (b) a stop sequence with unbalanced parentheses no longer raising `re.error`.
- Confirmed red→green: reverting only the two source files reproduces both failure modes (wrong truncation point; `re.error: unbalanced parenthesis`); reapplying the fix passes.
- Full `tests/unit_tests/` suite: 970 passed, 1 skipped, 1 xfailed (matches baseline).
- `ruff check` and `mypy` clean on changed files.

## AI-Generated disclosure

Found via an AI-assisted code review pass (Claude Code) over `langchain_aws/`. I personally reproduced both failure modes with standalone repros, verified the fix, traced the reachability from `BedrockLLM.invoke()`, and ran the full unit test suite plus lint/mypy before submitting.

### DIFF
--- a/libs/aws/langchain_aws/llms/sagemaker_endpoint.py
+++ b/libs/aws/langchain_aws/llms/sagemaker_endpoint.py
@@ -27,7 +27,7 @@ logger = logging.getLogger(__name__)
 
 def enforce_stop_tokens(text: str, stop: List[str]) -> str:
     """Cut off the text as soon as any stop words occur."""
-    return re.split("|".join(stop), text, maxsplit=1)[0]
+    return re.split("|".join(re.escape(s) for s in stop), text, maxsplit=1)[0]
 
 
 class LineIterator:
--- a/libs/aws/langchain_aws/utils.py
+++ b/libs/aws/langchain_aws/utils.py
@@ -86,7 +86,7 @@ class ContentHandlerBase(Generic[INPUT_TYPE, OUTPUT_TYPE]):
 
 def enforce_stop_tokens(text: str, stop: List[str]) -> str:
     """Cut off the text as soon as any stop words occur."""
-    return re.split("|".join(stop), text, maxsplit=1)[0]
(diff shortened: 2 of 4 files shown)


michaelnchin (commented):
Please review the [pull request requirements](https://docs.langchain.com/oss/python/contributing/overview#pull-request-requirements) in the contributing guide. Thank you!

> Mass automated contributions represent a denial-of-service attack on our human effort ... We will close pull requests and issues that appear to be low-effort spam.

PR #1167, opened 2026-07-14
fix(aws): Improve provider identification for model IDs with dotted versions

michaelnchin:
Addressing point 1 from #1158.

Updates `ChatBedrock` and `ChatBedrockConverse` to use a new `parse_model_provider` utility. 

This will now correctly handle base model IDs that contain non-demarcating `.` (ex. for model minor versions, like in ChatGPT 5.6: `us.openai.gpt-5.6-sol`)


michaelnchin (commented):
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/29375259860/job/87227263795

PR #1168, opened 2026-07-15
fix(aws): map search_result blocks for Bedrock Converse

### LINKED ISSUE #1159: ChatBedrockConverse raises ValueError on Anthropic search_result content blocks — Converse has a native searchResult block but no mapping exists
### Submission checklist

- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Package (Required)

langchain-aws

### Related Issues / PRs

#770 (Anthropic `document` custom-content citations — confirmed already supported; this issue is the symmetric gap for `search_result`), #537 (response-side `CitationsContentBlock` handling), #613 (document content type in tool results)

No AWS credentials or network needed — the failure is client-side, before any request is built:

```python
from langchain_aws.chat_models.bedrock_converse import _messages_to_bedrock
from langchain_core.messages import AIMessage, HumanMessage, ToolMessage

# Anthropic search_result block, exactly per the official spec:
# https://platform.claude.com/docs/en/build-with-claude/search-results
block = {
    "type": "search_result",
    "source": "https://wiki.example.com/policy",
    "title": "Vacation Policy",
    "content": [{"type": "text", "text": "Annual leave is 15 days."}],
    "citations": {"enabled": True},
}
msgs = [
    HumanMessage("What is the vacation policy?"),
    AIMessage("", tool_calls=[{"name": "retrieval", "args": {}, "id": "c1"}]),
    ToolMessage(content=[block], tool_call_id="c1"),
]
_messages_to_bedrock(msgs)
```

The same failure occurs end-to-end via `ChatBedrockConverse(...).invoke(msgs)` — any RAG agent whose retrieval tool returns citation-enabled search results dies on the first tool round-trip.

```shell
ValueError: Unsupported content block type:
{'type': 'search_result', 'source': 'https://wiki.example.com/policy', 'title': 'Vacation Policy', 'content': [{'type': 'text', 'text': 'Annual leave is 15 days.'}], 'citations': {'enabled': True}}
```

### Description

`ChatBedrockConverse` supports Anthropic-style `document` content blocks for citations (confirmed in #770), but the other citations input type — [`search_result`](https://platform.claude.com/docs/en/build-with-claude/search-results) — is not handled at all. `_lc_content_to_bedrock` has no branch for it, so any conversation containing one hits the unknown-block fallback and raises `ValueError` (currently line ~2704 on `main`).

This matters for RAG agents: `search_result` is the citations format designed for **tool results** (Anthropic `document` blocks are user-message content, not valid inside tool results), so it cannot be avoided by switching block types.

Note this is not an inherent Converse limitation — the Bedrock Converse API has a native equivalent for both directions:

- input: `searchResult` ([SearchResultBlock](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_SearchResultBlock.html))
- output: `citationsContent` ([CitationsContentBlock](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ContentBlock.html))

Both ends of the pipe speak citations; only the translation layer is missing the mapping.

**Expected behavior**: `search_result` blocks are translated to Converse `searchResult` blocks (field mapping is nearly 1:1: source, title, content, citations config), preserving citations on the Bedrock path — symmetrical with the existing `document` support.

**Current workaround**: we monkeypatch `_lc_content_to_bedrock` to flatten `search_result` into a text block (`<document source="..." title="...">...</document>`). This keeps production alive but destroys the citation metadata, so the model cannot emit structured citations via Bedrock.

Happy to submit a PR implementing the `searchResult` mapping.

### System Info

- langchain-aws 1.5.0 (branch also verified absent on `main` as of 2026-07-13)
- langchain-core 1.4.0, langchain 1.3.2
- Python 3.13, macOS

WOLIKIMCHENG:
## Summary

- Map Anthropic-style `search_result` content blocks to Bedrock Converse `searchResult` blocks.
- Preserve `source`, `title`, text-only `content`, and optional `citations` in the request payload.
- Add unit coverage for direct conversion, ToolMessage nesting, and rejection of non-text search result content.

Fixes #1159

## Testing

- `libs/aws/.venv/bin/python -m pytest libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py -q -k "search_result"`
- `python -m py_compile libs/aws/langchain_aws/chat_models/bedrock_converse.py libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py`
- `git diff --check`

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -2671,6 +2671,23 @@ def _lc_content_to_bedrock(
                     }
                 }
             )
+        elif block["type"] == "search_result":
+            search_result_content = _lc_content_to_bedrock(block["content"])
+            if not all(
+                set(search_result_content_block) == {"text"}
+                for search_result_content_block in search_result_content
+            ):
+                msg = "search_result content blocks must be text-only"
+                raise ValueError(msg)
+
+            search_result: Dict[str, Any] = {
+                "source": block["source"],
+                "title": block["title"],
+                "content": search_result_content,
+            }
+            if block.get("citations") is not None:
+                search_result["citations"] = block["citations"]
+            bedrock_content.append({"searchResult": search_result})
         # Only needed for tool_result content blocks.
         elif block["type"] == "json":
             bedrock_content.append({"json": block["json"]})
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -639,6 +639,53 @@ def test_messages_to_bedrock_with_cache_point() -> None:
     assert [] == actual_system
 
 
+def test_messages_to_bedrock_with_tool_message_search_result() -> None:
+    messages = [
+        ToolMessage(
+            content=[
+                {
+                    "type": "search_result",
+                    "source": "web",
+                    "title": "Weather report",
+                    "content": [{"type": "text", "text": "It is sunny in Seattle."}],
+                    "citations": {"enabled": True},
+                }
+            ],
+            tool_call_id="tool-123",
+            status="success",
+        )
+    ]
+
+    actual_messages, actual_system = _messages_to_bedrock(messages)
+
+    assert actual_messages == [
+        {
+            "role": "user",
+            "content": [
+                {
+                    "toolResult": {
+                        "content": [
+                            {
+                                "searchResult": {
+                                    "source": "web",
+                                    "title": "Weather report",
+                                    "content": [
+                                        {"text": "It is sunny in Seattle."}
+                                    ],
+                                    "citations": {"enabled": True},
+                                }
+                            }
+                        ],
+                        "status": "success",
+                        "toolUseId": "tool-123",
+                    }
+                }
+            ],
+        }
+    ]
+    assert [] == actual_system
+
+
 def test__messages_to_bedrock_empty_list() -> None:
     messages: List[BaseMessage] = []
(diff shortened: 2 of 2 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -2671,6 +2671,23 @@ def _lc_content_to_bedrock(
                     }
                 }
             )
+        elif block["type"] == "search_result":
+            search_result_content = _lc_content_to_bedrock(block["content"])
```
michaelnchin, line 2675: Using `_lc_content_to_bedrock` to convert the internal block content will cause some issues here:
1. It mutates whitespace-only text (ex. `" "` get converted to `"."`), which we use to work around a Converse limitation for message text blocks. But this should not be used in a way that would potentially corrupt the sourceContent that citations directly reference as the original material
2.  Any text block carrying citations (e.g. a previously-cited answer archived and served back as a retrieval result) gets wrapped into citationsContent by the text branch, and then ends up failing the following text type check with a misleading error.

michaelnchin (changes requested):

michaelnchin (commented):
Thanks for the submission @WOLIKIMCHENG - closing as we've opened a separate PR (#1175) which will cover the mentioned conversion gap, as well as handle a separate blocking issue for search results in multi-turn conversations.

PR #1173, opened 2026-07-20
fix: replace EOL Bedrock Meta Llama 3.2 90B with Llama 3.3 70B

mdrxy:
The `TestBedrockMetaStandard` integration tests were failing with `ResourceNotFoundException: This model version has reached the end of its life` because `us.meta.llama3-2-90b-instruct-v1:0` was retired on Amazon Bedrock. This swaps the test model to the current, generally available `us.meta.llama3-3-70b-instruct-v1:0`, which is the natural successor and shares the same Converse/tool-calling behavior the test suite relies on.

Made by [Open SWE](https://openswe.vercel.app/agents/aa870ff9-6dad-890d-7340-3a177a74e367)

### DIFF
--- a/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
@@ -160,7 +160,7 @@ class TestBedrockMetaStandard(ChatModelIntegrationTests):
 
     @property
     def chat_model_params(self) -> dict:
-        return {"model": "us.meta.llama3-2-90b-instruct-v1:0"}
+        return {"model": "us.meta.llama3-3-70b-instruct-v1:0"}
 
     @property
     def standard_chat_model_params(self) -> dict:
--- a/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/integration_tests/chat_models/test_bedrock_converse.py
@@ -160,7 +160,7 @@ class TestBedrockMetaStandard(ChatModelIntegrationTests):
 
     @property
     def chat_model_params(self) -> dict:
-        return {"model": "us.meta.llama3-3-70b-instruct-v1:0"}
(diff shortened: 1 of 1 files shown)


mdrxy:
hmm https://github.com/langchain-ai/langchain-aws/actions/runs/29748097804/job/88371159235

michaelnchin:
Tried rerunning, the deprecation warning is gone now but we're getting some new assertion errors on 3.3: https://github.com/langchain-ai/langchain-aws/actions/runs/29773233213/job/88456327286

Taking a look 👀

michaelnchin (commented):
@mdrxy thanks for taking the initial look into the Meta tests migration. I did a bit of extra testing to see how we can unblock this:

- Running the MetaStandard suite with Llama 3.3 70B locally, I saw the same failures with the structured output tests pretty consistently(usually 4+ on any given run, out of 20 of which only one saw every test pass). It looks like Llama 3.3 decides to return native toolUse blocks less than half the time, the rest of the time they just come back as raw JSON instead and so the assertions fail.

Subsequently, moved onto checking Llama4 instead:
- Llama4 Maverick fares slightly better than 3.3, but is still incredibly flaky (only 3/20 MetaStandard runs fully succeeded). Cause of this is the same duplicate tool calls issue as you previously noted in the Maverick tests revert last month: https://github.com/langchain-ai/langchain-aws/pull/1126
- Interestingly, OTOH, Llama 4 Scout **does** appear to work consistently (at least for all 50 of my runs), and always returns the correct `toolUse` format like Maverick, just without the duplication issue.

Going to update the PR to `us.meta.llama4-scout-17b-instruct-v1:0`. Also, I noticed that the `test_tool_calling*` tests that we currently have as `xfail` actually pass now (since Scout returns integer tool input as the same type in the response instead of str) so I'll remove those overrides also.

michaelnchin:
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/29801534818/job/88543403576

michaelnchin (approved):

PR #1178, opened 2026-07-23
feat(aws): support Bedrock Guardrails in ChatAnthropicBedrock

### LINKED ISSUE #1177: ChatAnthropicBedrock: no supported way to attach Amazon Bedrock Guardrails
### Description

`ChatAnthropicBedrock` (added in #891) wraps the Anthropic SDK's `AnthropicBedrock` client, which calls Bedrock's native `InvokeModel` / `InvokeModelWithResponseStream` APIs. On that path, Amazon Bedrock Guardrails are attached via HTTP request headers (AWS docs: *"Include a guardrail with a model invocation"*):

- `X-Amzn-Bedrock-GuardrailIdentifier`
- `X-Amzn-Bedrock-GuardrailVersion`
- `X-Amzn-Bedrock-Trace`

The class currently exposes no parameter for this — there is no `guardrail_config` (or equivalent), and nothing translates guardrail settings into those headers. Verified against `langchain-aws` 1.6.3 and current `main`: the class's AWS surface is credentials/region/retries/timeout only, and [`_client_params` (1.6.3, L139-L155)](https://github.com/langchain-ai/langchain-aws/blob/langchain-aws%3D%3D1.6.3/libs/aws/langchain_aws/chat_models/anthropic.py#L139-L155) passes `default_headers` through but builds nothing guardrail-related. (The docstring's "unsupported params are passed to `messages.create`" escape hatch doesn't help here — guardrail parameters on this API are transport *headers*, not body fields.)

For contrast, in the same release `ChatBedrockConverse` has [`guardrail_config` (L584)](https://github.com/langchain-ai/langchain-aws/blob/langchain-aws%3D%3D1.6.3/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L584) and `guard_last_turn_only` (L655).

By contrast, `ChatBedrockConverse` supports `guardrail_config` (plus `guard_last_turn_only`), so Anthropic models invoked through the Converse API can be guarded — but users choosing `ChatAnthropicBedrock` for its native-SDK features currently cannot attach a guardrail through the class API.

### Why this matters

Beyond feature parity, this interacts badly with **IAM-enforced guardrails**: organizations can require guardrails at the IAM level via the `bedrock:GuardrailIdentifier` condition key on `bedrock:InvokeModel`. That condition key is only populated when the request actually carries a guardrail — so under such a policy, **all `ChatAnthropicBedrock` traffic is denied**, with no way to comply from the class API. We hit exactly this while rolling out org-wide guardrail enforcement: `ChatBedrockConverse`-based agents comply fine; `ChatAnthropicBedrock`-based ones cannot.

### Reproduction

```python
from langchain_aws import ChatAnthropicBedrock, ChatBedrockConverse

# Converse path — guardrail attaches fine:
ChatBedrockConverse(
    model="anthropic.claude-sonnet-4-5",
    guardrail_config={"guardrailIdentifier": "gr-abc123", "guardrailVersion": "1"},
)

# Anthropic SDK path — no equivalent:
ChatAnthropicBedrock(
    model="claude-sonnet-4-5",
    # guardrail_config=...   # no such parameter; nothing emits the
    #                        # X-Amzn-Bedrock-Guardrail* headers
)
```

### Workaround

The class already plumbs `default_headers` through to the underlying client, and the guardrail parameters are plain request headers on this API, so this works today:

```python
ChatAnthropicBedrock(
    model="claude-sonnet-4-5",
    default_headers={
        "X-Amzn-Bedrock-GuardrailIdentifier": "gr-abc123",
        "X-Amzn-Bedrock-GuardrailVersion": "1",
        "X-Amzn-Bedrock-Trace": "ENABLED",
    },
)
```

It's just undiscoverable, and easy to get silently wrong — a typo'd header name means no guardrail and no error.

### Proposal

Accept `guardrail_config` on `ChatAnthropicBedrock` with the same shape `ChatBedrockConverse` uses (`{guardrailIdentifier, guardrailVersion, trace}`), and translate it into the three `X-Amzn-Bedrock-*` headers when building the client (merged with any user-supplied `default_headers`). Small, contained change — happy to contribute the PR with tests if the approach sounds right.

### System info

- `langchain-aws` 1.6.3 (also checked `main` as of this writing)
- Python 3.12

pablodg12:
## Why

`ChatAnthropicBedrock` wraps the Anthropic SDK's `AnthropicBedrock` client, which calls Bedrock's **native `InvokeModel` APIs** — where guardrails are attached via HTTP request headers, not a request body field. The class exposes no parameter for them, so Amazon Bedrock Guardrails cannot be attached through the class API at all. This also interacts badly with IAM-enforced guardrails (`bedrock:GuardrailIdentifier` condition key on `InvokeModel`): under such a policy, all traffic from this class is denied with no way to comply.

Fixes #1177

## What

- New `guardrail_config` field — same shape and `guardrails` alias as `ChatBedrockConverse`, for parity:
  `{"guardrailIdentifier": "...", "guardrailVersion": "...", "trace": "enabled"}`
- `_guardrail_headers()` translates it into `X-Amzn-Bedrock-GuardrailIdentifier` / `X-Amzn-Bedrock-GuardrailVersion` / `X-Amzn-Bedrock-Trace`, merged over `default_headers`.
- Incomplete config (missing identifier or version) raises a `ValueError` instead of silently sending unguarded requests.
- 6 hermetic unit tests. `make format` / `make lint` (ruff + mypy) / full unit suite (986 passed) are green.

## Verified end-to-end against live Bedrock

Using this branch's code (us-east-1, Claude Sonnet 4.5, disposable guardrail with a word filter):

| Case | Result |
|---|---|
| No `guardrail_config`, trigger word | Model answers normally |
| `guardrail_config`, benign prompt | Normal answer, `guardrailAction: NONE` in `response_metadata` |
| `guardrail_config`, trigger word | Guardrail's blocked message returned, `guardrailAction: INTERVENED` |
| `guardrails=` alias, trigger word | Same intervention — alias works |
| `.stream()` with `guardrail_config`, trigger word | Intervention arrives as regular stream chunks (blocked message), no exception |
| `.ainvoke()` (async client) with `guardrail_config`, trigger word | `guardrailAction: INTERVENED` — async client inherits the headers |

The intervened response parses as a normal message through the Anthropic SDK — no shape issues, on both the sync and streaming paths.

## Review notes

- **Merge precedence**: explicit `guardrail_config` wins over conflicting keys in a user-supplied `default_headers`. Flagging in case the reverse is preferred.
- `trace` is upper-cased to the header enum (`"enabled"` → `ENABLED`).
- Headers are applied at client construction (`_client_params`), so the sync, async and streaming clients all inherit them — no streaming-specific handling needed (verified empirically for streaming and async, see table above).

---

*Disclaimer (per the repo's contribution guidelines): this contribution was developed with the assistance of an AI coding agent, with human direction and review, and validated end-to-end against a live Bedrock guardrail.*

### DIFF
--- a/libs/aws/langchain_aws/chat_models/anthropic.py
+++ b/libs/aws/langchain_aws/chat_models/anthropic.py
@@ -110,6 +110,19 @@ class ChatAnthropicBedrock(ChatAnthropic):
     If not provided, will be read from 'AWS_SESSION_TOKEN' environment variable.
     """
 
+    guardrail_config: dict[str, Any] | None = Field(default=None, alias="guardrails")
+    """Configuration for an Amazon Bedrock Guardrail applied to every request.
+
+    Same shape as `ChatBedrockConverse`, e.g.
+    `{"guardrailIdentifier": "gr-abc123", "guardrailVersion": "1", "trace": "enabled"}`.
+
+    The native Bedrock `InvokeModel` APIs used by the Anthropic SDK attach
+    guardrails as HTTP request headers rather than a request body field, so this
+    config is translated into the `X-Amzn-Bedrock-GuardrailIdentifier`,
+    `X-Amzn-Bedrock-GuardrailVersion` and `X-Amzn-Bedrock-Trace` headers and
+    merged over `default_headers` (guardrail values win on key conflicts).
+    """
+
     @property
     def _llm_type(self) -> str:
         """Return type of chat model."""
@@ -135,6 +148,40 @@ class ChatAnthropicBedrock(ChatAnthropic):
         """
         return ["langchain", "chat_models", "anthropic_bedrock"]
 
+    def _guardrail_headers(self) -> dict[str, str]:
+        """Translate `guardrail_config` into Bedrock guardrail request headers.
+
+        The native `InvokeModel` / `InvokeModelWithResponseStream` APIs attach
+        guardrails via HTTP headers instead of a request body field, so the
+        Converse-style config is mapped onto those headers here.
+
+        Returns:
+            Header dict to merge into the client's default headers; empty when
+            no guardrail is configured.
+
+        Raises:
+            ValueError: If `guardrail_config` is missing `guardrailIdentifier`
+                or `guardrailVersion` — failing loudly instead of silently
+                sending unguarded requests.
+        """
+        if not self.guardrail_config:
+            return {}
+        identifier = self.guardrail_config.get("guardrailIdentifier")
+        version = self.guardrail_config.get("guardrailVersion")
+        if not identifier or not version:
+            msg = (
+                "guardrail_config requires both 'guardrailIdentifier' and "
+                "'guardrailVersion'."
+            )
+            raise ValueError(msg)
+        headers = {
+            "X-Amzn-Bedrock-GuardrailIdentifier": str(identifier),
+            "X-Amzn-Bedrock-GuardrailVersion": str(version),
+        }
+        if trace := self.guardrail_config.get("trace"):
+            headers["X-Amzn-Bedrock-Trace"] = str(trace).upper()
+        return headers
+
     @cached_property
     def _client_params(self) -> dict[str, Any]:
         """Get client parameters for AnthropicBedrock."""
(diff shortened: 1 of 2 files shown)


michaelnchin (approved):
Thanks for the issue and pull request submissions @pablodg12, looks good. 

One thing worth mentioning, this only allows passing guardrail config on instantiation of `ChatAnthropicBedrock`. `ChatBedrockConverse` is more flexible as it also allows passing the guardrail config [as a kwarg per model invocation](github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L1197-L1207). We should also be able to add the same into `ChatAnthropicBedrock` since the Anthropic SDK accepts `extra_headers` for individual calls. Will take this in a follow up PR.

PR #1182, opened 2026-07-26
fix(langgraph-checkpoint-aws): make ValkeySaver writes atomic and deduped

### LINKED ISSUE #1174: AsyncValkeySaver.aput_writes / ValkeySaver.put_writes are not atomic or deduped, causing lost writes for parallel tool calls under Send fan-out
## Environment
- `langgraph-checkpoint-aws`: 1.2.0 (bug present on `main` as of 2026-07 as well)
- `langgraph`: 1.2.9
- `langchain` / `langchain-core`: 1.3.14 / 1.4.9
- `valkey` (client): 6.1.1
- Python 3.12

## Summary

`AsyncValkeySaver.aput_writes` (and the sync `ValkeySaver.put_writes`) store every task's pending writes for a checkpoint as one JSON blob under a single `writes_key`, keyed only by `(thread_id, checkpoint_ns, checkpoint_id)` — **not** by `task_id`. The write path is a plain, unprotected read-modify-write:

```python
# checkpoint/valkey/async_saver.py, aput_writes()
existing_data = await self.client.get(writes_key)
...
new_writes = self._serialize_writes_data(writes, task_id)
existing_writes.extend(new_writes)          # (1) no dedup
pipe = self.client.pipeline()
pipe.set(writes_key, orjson.dumps(existing_writes))  # (2) no atomicity
...
await pipe.execute()
```

This has two independent bugs:

1. **No dedup.** Writes are appended with `list.extend`, never deduped by `(task_id, idx)`. `langgraph.checkpoint.memory.InMemorySaver.put_writes` treats `(task_id, WRITES_IDX_MAP.get(channel, idx))` as a key and only writes it once (`langgraph/checkpoint/memory/__init__.py`), specifically so that replaying a task around an `interrupt()`/resume doesn't duplicate or corrupt writes already recorded for that task. The Valkey backend has no equivalent guard.

2. **No atomicity across concurrent callers.** The GET and SET are two separate round trips with no `WATCH`/`MULTI`/`EXEC` transaction and no Lua script. When two LangGraph Pregel tasks call `aput_writes` concurrently for the *same checkpoint* (same `writes_key`, different `task_id`s), both can `GET` the same (stale) snapshot before either `SET`s — whichever `SET` lands last **silently discards** the other task's write entirely. This is a classic lost-update race.

## Why this is reachable in practice, not just in theory

Agents built via `langchain.agents.factory.create_agent` fan each parallel tool call out into its own independent Pregel task rather than invoking `ToolNode` once with all pending calls:

```python
# langchain/agents/factory.py
if pending_tool_calls:
    return [Send("tools", [tool_call]) for tool_call in pending_tool_calls]
```

For *N* parallel tool calls, that's *N* separate `Send`s, each becoming its own Pregel task with its own `task_id`. LangGraph flushes each task's writes via a fire-and-forget call, with no serialization across tasks writing to the same checkpoint (`langgraph/pregel/_runner.py: PregelRunner.commit()`, `langgraph/pregel/_loop.py: SyncPregelLoop.put_writes`, `langgraph/pregel/_executor.py: AsyncBackgroundExecutor.submit` — scheduled via `run_coroutine_threadsafe`, no lock/queue).

So: 2+ parallel tool calls → 2+ concurrent `aput_writes` calls for the same `writes_key` → real, observed lost updates.

## Observed symptom

In an application using Bedrock's Converse API (via `langchain-aws`), a turn with 2 parallel tool calls followed by a later `interrupt()`/resume produces:

```
ValidationException: Expected toolResult blocks at messages.N.content for the following Ids: tooluse_XXXXXXXX
```

Across repeated runs of the identical scenario, a *different* tool_use ID goes missing each time — the non-determinism is the signature of the race, not a fixed logic bug.

## Minimal reproduction

No real Valkey server needed — the mock client models the same GET-then-SET round trip a real network call would make, which is exactly where the race lives. Reproduces reliably (5/5 runs in our testing).

```python
"""Reproduces a lost-update race in AsyncValkeySaver.aput_writes.

Two LangGraph Pregel tasks (e.g. two tool calls fanned out via
Send("tools", [tool_call]) per call, as langchain.agents.factory does for
parallel tool calls) call aput_writes concurrently for the *same checkpoint*.
Each does: GET existing writes -> append its own write in Python -> SET the
whole list back. Because the two GETs can both return before either SET lands,
whichever SET finishes last overwrites the other task's write - it is not
merged, it is gone.

Run: python3 repro_valkey_lost_write.py
Requires: pip install langgraph-checkpoint-aws[valkey]
"""

import asyncio
from unittest.mock import AsyncMock, MagicMock

import orjson
from langgraph_checkpoint_aws import AsyncValkeySaver

WRITES_KEY = "writes:{t1}::c1"


def make_mock_client(stored: dict, network_delay: float) -> MagicMock:
    """A fake Valkey client whose SET only lands after `network_delay` seconds,
    modeling the real gap between a GET returning and a subsequent SET landing
    over the network - the exact window the race exploits."""

    async def fake_get(key):
        return stored.get(key)

    def fake_pipeline():
        pipe = MagicMock()
        pending = {}

        def fake_set(key, value):
            pending["key"], pending["value"] = key, value

        async def fake_execute():
            await asyncio.sleep(network_delay)
            if "key" in pending:
                stored[pending["key"]] = pending["value"]

        pipe.set = MagicMock(side_effect=fake_set)
        pipe.expire = MagicMock()
        pipe.execute = AsyncMock(side_effect=fake_execute)
        return pipe

    client = MagicMock()
    client.get = AsyncMock(side_effect=fake_get)
    client.pipeline = MagicMock(side_effect=fake_pipeline)
    return client


async def main() -> None:
    stored: dict = {}
    saver = AsyncValkeySaver(client=make_mock_client(stored, network_delay=0.02), ttl=None)
    config = {"configurable": {"thread_id": "t1", "checkpoint_ns": "", "checkpoint_id": "c1"}}

    # Two independent Pregel tasks (task-a, task-b) writing to the same
    # checkpoint concurrently - exactly what happens when a single AI turn
    # makes 2+ parallel tool calls under langchain.agents.factory's
    # Send("tools", [tool_call])-per-call fan-out.
    await asyncio.gather(
        saver.aput_writes(config, [("messages", "task-a-result")], task_id="task-a"),
        saver.aput_writes(config, [("messages", "task-b-result")], task_id="task-b"),
    )

    stored_writes = orjson.loads(stored[WRITES_KEY])
    task_ids_present = sorted(w["task_id"] for w in stored_writes)

    print(f"Writes actually stored for the checkpoint: {task_ids_present}")
    print(f"Writes expected:                           ['task-a', 'task-b']")

    if task_ids_present == ["task-a", "task-b"]:
        print("\nOK - both writes present (bug not reproduced this run; rerun, it's a race).")
    else:
        missing = {"task-a", "task-b"} - set(task_ids_present)
        print(f"\nBUG REPRODUCED: {sorted(missing)}'s write was silently lost.")


if __name__ == "__main__":
    asyncio.run(main())
```

Example output:
```
Writes actually stored for the checkpoint: ['task-b']
Writes expected:                           ['task-a', 'task-b']

BUG REPRODUCED: ['task-a']'s write was silently lost.
```

## How other checkpointers in the ecosystem handle this correctly

- **Postgres** (`langgraph-checkpoint-postgres`): `checkpoint_writes` table has `PRIMARY KEY (thread_id, checkpoint_ns, checkpoint_id, task_id, idx)`, written via `INSERT ... ON CONFLICT (...) DO UPDATE` / `DO NOTHING`. Per-row atomicity, DB-enforced.
- **SQLite** (`langgraph-checkpoint-sqlite`): same key columns, `INSERT OR REPLACE`/`INSERT OR IGNORE`, executed under a lock with `conn.commit()`.
- **This same package's own DynamoDB backend** (`checkpoint/dynamodb/unified_repository.py`, `put_single_write_item`): stores each write as its own item with `ConditionExpression="attribute_not_exists(PK)"` on `put_item` to prevent silent overwrites.

Only the Valkey backend collapses everything into one shared blob with no per-row keying and no conditional write.

## Suggested fix

Either:
- **(a)** Store each write as its own key/hash-field keyed by `(task_id, idx)` (e.g. a Valkey `HSET writes_key task_id:idx value`, which is atomic per-field and naturally dedupes on re-write of the same field), removing the need for a Python-side read-modify-write entirely; or
- **(b)** Keep the JSON-blob design but wrap the read-modify-write in a `WATCH`/`MULTI`/`EXEC` optimistic transaction (retrying on `WatchError`) or a Lua script (`EVAL`) that performs the get-merge-set atomically server-side, plus dedup the in-memory merge by `(task_id, idx)` to match `InMemorySaver`'s semantics.

Happy to open a PR with option (a) or (b) if that's useful — we've already validated a working fix (dedup + a process-local lock as an interim mitigation) in our own fork, but a proper fix belongs server-side in Valkey for correctness across multiple worker processes.

apoorva-01:
Fixes #1174

`put_writes`/`aput_writes` did a plain read-modify-write on one shared JSON blob. When two tasks flush writes for the same checkpoint at once (parallel tool calls each become their own `Send` task), both read the stale blob and the last `SET` drops the other's write. Shows up as a Bedrock `ValidationException` about a missing `toolResult`.

Wrapped it in a WATCH/MULTI/EXEC transaction (retry on `WatchError`) and deduped the merge by `(task_id, idx)`, same as `InMemorySaver`. I kept the JSON-blob format instead of moving to per-field `HSET`. That version would be atomic for free but changes the on-disk layout, so I went with the smaller change. Easy to switch if you'd rather.

Tests use fakeredis: 8 tasks writing one checkpoint at once with nothing lost, plus the dedup cases. They fail on `main`, pass with the fix.

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/valkey/base.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/valkey/base.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import base64
 import json
+import logging
 import random
 from collections.abc import Sequence
 from typing import Any, cast
@@ -26,6 +27,8 @@ from .utils import set_client_info, set_client_name
 # before giving up. Contention is bounded by the number of tasks flushing writes
 # for the same checkpoint concurrently, so this is only ever reached pathologically.
 _PUT_WRITES_MAX_ATTEMPTS = 100
+
+logger = logging.getLogger(__name__)
 
 
 class BaseValkeySaver(BaseCheckpointSaver[str]):
@@ -254,8 +257,18 @@ class BaseValkeySaver(BaseCheckpointSaver[str]):
         try:
             parsed = orjson.loads(existing_data)
         except (orjson.JSONDecodeError, TypeError, ValueError):
+            logger.warning(
+                "Discarding un-decodable checkpoint writes blob; existing "
+                "writes for this checkpoint will be replaced."
+            )
             return []
-        return parsed if isinstance(parsed, list) else []
+        if not isinstance(parsed, list):
+            logger.warning(
+                "Checkpoint writes blob was not a JSON list; existing "
+                "writes for this checkpoint will be replaced."
+            )
+            return []
+        return parsed
 
     @staticmethod
     def _merge_writes(
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/valkey/base.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/valkey/base.py
@@ -8,6 +8,7 @@ import random
 from collections.abc import Sequence
 from typing import Any, cast
 
+import orjson
 from langchain_core.runnables import RunnableConfig
 from langgraph.checkpoint.base import (
     WRITES_IDX_MAP,
@@ -20,6 +21,11 @@ from langgraph.checkpoint.base import (
 )
 
 from .utils import set_client_info, set_client_name
+
+# Number of optimistic-lock attempts for the writes read-modify-write transaction
+# before giving up. Contention is bounded by the number of tasks flushing writes
+# for the same checkpoint concurrently, so this is only ever reached pathologically.
+_PUT_WRITES_MAX_ATTEMPTS = 100
 
 
 class BaseValkeySaver(BaseCheckpointSaver[str]):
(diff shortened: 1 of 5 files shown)


michaelnchin (commented):
Thanks for the fix submission @apoorva-01 - @ulag could you kindly pull down the changes from this branch, and confirm if they fully resolve your issue?

libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/valkey/base.py
```diff
@@ -224,6 +230,64 @@ def _serialize_writes_data(
...
+        elif not isinstance(existing_data, (bytes, bytearray, memoryview)):
+            try:
+                existing_data = orjson.dumps(existing_data)
+            except (TypeError, ValueError):
+                return []
+        try:
+            parsed = orjson.loads(existing_data)
+        except (orjson.JSONDecodeError, TypeError, ValueError):
+            return []
+        return parsed if isinstance(parsed, list) else []
```
michaelnchin: I think we should add log warnings here for the "corrupted" stored writes that are potentially getting tossed without the user being aware of the actual failure point. Otherwise they'll just see the same error from the race condition you're fixing here

michaelnchin (approved):
Thanks @apoorva-01 !

PR #1183, opened 2026-07-27
feat(aws): Support per-request guardrail configuration on ChatAnthropicBedrock

michaelnchin:
Related: #1177, addresses follow-up noted here: https://github.com/langchain-ai/langchain-aws/pull/1178#pullrequestreview-4782812273

Moving the logic for guardrail header injection out of `_client_params` (which is frozen at construction) and into a new `_get_request_payload` method (called per-request) that instead sets the guardrail config via the Anthropic SDK's `extra_headers` parameter. 

With this, users of `ChatAnthropicBedrock` can now set per-invoke guardrail overrides, bringing it in line with what `ChatBedrockConverse` supports.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/anthropic.py
+++ b/libs/aws/langchain_aws/chat_models/anthropic.py
@@ -28,6 +28,45 @@ def _get_default_model_profile(model_name: str) -> ModelProfile:
     return default.copy()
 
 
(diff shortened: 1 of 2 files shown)


michaelnchin (commented):
Demo of this working with 2 different guardrails, one on model instantiation and one on invoke:
Image

PR #1186, opened 2026-07-28
feat(aws): Support Claude Opus 5 default thinking in ChatBedrock

michaelnchin:
Mainly updating ChatBedrock to ensure that we correctly handle Opus 5's potential output formats while streaming on the default adaptive thinking mode (as was previously done for Sonnet and Fable 5 in https://github.com/langchain-ai/langchain-aws/pull/1157.)

Also, (per the [AWS docs](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-5.html)) Opus 5 re-instated the ability to disable thinking entirely, which is more flexible than Sonnet 5 (accepts `thinking: {"type": "disabled"}` but it doesn't have an effect) and Opus 5 (rejects `disabled` with a hard exception). So, I've also added Opus 5 into the streaming integration tests matrices to ensure that its forced plaintext streaming mode is represented.

### DIFF
--- a/libs/aws/langchain_aws/utils.py
+++ b/libs/aws/langchain_aws/utils.py
@@ -489,7 +489,9 @@ def thinking_disabled_in_params(params: dict) -> bool:
 
 def thinking_on_by_default(model: str) -> bool:
     """Check if the model runs adaptive thinking without a thinking parameter."""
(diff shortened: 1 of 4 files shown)


michaelnchin (commented):
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/30323330966

PR #1188, opened 2026-07-29
feat(aws): support for `reasoning_effort` param for `ChatBedrockConverse`

imnishitha:
Mirrors the standardized `reasoning_effort` chat-model parameter introduced in langchain-ai/langchain#38887 for `ChatBedrockConverse`. Today, configuring extended `thinking`/`reasoning` on Bedrock means hand-building provider-specific dicts inside `additional_model_request_fields`. This PR adds a single `reasoning_effort`: `Literal["low","medium","high","xhigh","max"]` field/call-time kwarg that translates to the right wire shape per model family:

| Model family | Request shape |
|--------------|---------------|
| Claude (Opus 5 / Sonnet 5) | `{"thinking": {"type": "adaptive"}, "output_config": {"effort": ...}}` |
| Amazon Nova 2 | `{"reasoningConfig": {"type": "enabled", "maxReasoningEffort": ...}}` |
| OpenAI (native GPT-5.x and open-weight gpt-oss), Moonshot Kimi K2 | `{"reasoning_effort": ...}` |
| Anything else | warnings and no-op |

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -614,18 +614,12 @@ class ChatBedrockConverse(BaseChatModel):
 
     """
 
-    reasoning_effort: Optional[
-        Literal["none", "low", "medium", "high", "xhigh", "max"]
-    ] = None
+    reasoning_effort: Optional[Literal["low", "medium", "high", "xhigh", "max"]] = None
     """Reasoning effort level for models that support configurable reasoning.
 
     Translated into the appropriate provider-specific request format based on the
     model family. Only applied to models whose profile declares supported
     ``reasoning_effort_levels``; unsupported values raise ``ValueError``.
-
-    ``"none"`` disables reasoning outright for native GPT-5.x models, it is
-    not equivalent to omitting the argument, since these models otherwise
-    default to ``"medium"``.
 
     Explicit values already present in ``additional_model_request_fields`` take
     precedence. If the model has no known reasoning-effort translation, a warning
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -614,12 +614,18 @@ class ChatBedrockConverse(BaseChatModel):
 
     """
 
-    reasoning_effort: Optional[Literal["low", "medium", "high", "xhigh", "max"]] = None
+    reasoning_effort: Optional[
+        Literal["none", "low", "medium", "high", "xhigh", "max"]
+    ] = None
     """Reasoning effort level for models that support configurable reasoning.
 
     Translated into the appropriate provider-specific request format based on the
     model family. Only applied to models whose profile declares supported
     ``reasoning_effort_levels``; unsupported values raise ``ValueError``.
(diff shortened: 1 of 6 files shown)


libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -595,6 +614,24 @@ class Joke(BaseModel):
...
+    """Reasoning effort level for models that support configurable reasoning.
+
+    Translated into the appropriate provider-specific request format based on the
+    model family. Only applied to models whose profile declares supported
+    ``reasoning_effort_levels``; unsupported values raise ``ValueError``.
+
+    ``"none"`` disables reasoning outright for native GPT-5.x models, it is
+    not equivalent to omitting the argument, since these models otherwise
+    default to ``"medium"``.
+
```
michaelnchin: The `none` option should be excluded since GPT 5.x models are only supported on `bedrock-mantle` endpoint (ChatBedrockConverse uses `bedrock-runtime`): https://docs.aws.amazon.com/bedrock/latest/userguide/models-endpoint-availability.html#model-endpoints-openai

libs/aws/langchain_aws/data/_profiles.py
michaelnchin, line 1: We should avoid manually updating `_profiles.py` as it's auto-synced directly to upstream models.dev: https://github.com/anomalyco/models.dev/tree/dev/providers/amazon-bedrock/models

Override fields should be added here instead: https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/data/profile_augmentations.toml, this should look something like:
```python
[overrides."anthropic.claude-sonnet-5"]
reasoning_effort_levels = ["low", "medium", "high", "xhigh", "max"]
```

(note that you will also have to override the regional profiles for each model)

michaelnchin (changes requested):
Thanks @imnishitha! Couple of observations -

libs/aws/langchain_aws/chat_models/bedrock_converse.py
```diff
@@ -595,6 +614,18 @@ class Joke(BaseModel):
 
     """
 
+    reasoning_effort: Optional[Literal["low", "medium", "high", "xhigh", "max"]] = None
```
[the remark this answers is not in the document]
imnishitha: these are not supported only by bedrock-mantle

libs/aws/langchain_aws/data/_profiles.py
imnishitha: done

libs/aws/langchain_aws/data/profile_augmentations.toml
```diff
@@ -5,3 +5,60 @@ image_url_inputs = true
 pdf_inputs = true
 pdf_tool_message = true
 image_tool_message = true
+
+[overrides."anthropic.claude-opus-5"]
+reasoning_effort_levels = ["low", "medium", "high", "xhigh", "max"]
```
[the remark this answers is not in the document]
imnishitha: it will be regenerated when cli runs to pull data from models.dev

ccurme (approved):

michaelnchin (approved):
LGTM!

PR #1190, opened 2026-07-30
feat(aws): add ChatAnthropicMantle for Bedrock Mantle Anthropic Messages API

3coins:
## Summary

Adds `ChatAnthropicMantle`, a thin `ChatAnthropic` subclass targeting the Amazon Bedrock **Mantle** Anthropic Messages endpoint (`bedrock-mantle.{region}.api.aws/anthropic`). Because Mantle speaks the native Anthropic Messages API, this integration is a small subclass (mirroring the `AzureChatOpenAI` pattern): only the endpoint + auth are overridden, so tool calling, structured output, streaming, thinking, tracing, and multimodal are inherited from `ChatAnthropic` and stay in sync with `langchain-anthropic`.

This is the Anthropic counterpart to `ChatOpenAIMantle` (#1189) and follows the same design.

## What's included
* `ChatAnthropicMantle(ChatAnthropic)` — region → `base_url` resolution (`https://bedrock-mantle.{region}.api.aws/anthropic`; the Anthropic SDK appends `/v1/messages`) and a Bedrock API key (`bedrock_api_key` / `AWS_BEARER_TOKEN_BEDROCK`) routed into the Anthropic `api_key`, sent as the standard `x-api-key` header.
* Uses the existing `langchain-aws[anthropic]` optional extra (no new dependency).
* Exports wired into `langchain_aws` and `langchain_aws.chat_models` (lazy-imported).
* `is_lc_serializable()` overridden to `False` (the partner class isn't registered in `langchain-core`'s deserialization allowlist).
* Standard `ChatModelUnitTests` suite + focused unit tests.
* Skip-gated standard `ChatModelIntegrationTests` (skips unless `AWS_BEARER_TOKEN_BEDROCK` is set; unskip in CI once the account has `bedrock-mantle` permissions).

## Difference from `ChatAnthropicBedrock`

`ChatAnthropicBedrock` talks to `bedrock-runtime` via the `AnthropicBedrock` SigV4 client (AWS credentials, invocation logging). `ChatAnthropicMantle` talks to the `bedrock-mantle` endpoint and authenticates with a Bedrock API key (bearer/`x-api-key`) — no AWS SigV4 credential chain required.

## Naming

`ChatAnthropicMantle` — provider-SDK-first (like `ChatAnthropicBedrock` / `ChatOpenAIMantle`), with `Mantle` making the endpoint explicit.

## Testing
* Unit: standard suite + focused tests pass (19 passed, 1 skipped — `test_serdes` skips because the class is intentionally not lc-serializable). Full `libs/aws` unit suite green with no regressions.
* Integration (live, `anthropic.claude-sonnet-5`, `us-east-1`): passing. `temperature` is intentionally omitted from `standard_chat_model_params` because Claude Sonnet 5 / Opus 5 deprecate it (they use adaptive thinking / `output_config.effort`); sending it returns a 400. The model id is overridable via `ANTHROPIC_MANTLE_MODEL`.
* ruff + mypy clean.

<details>
  <summary>Integration test results:</summary>

```

tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_no_overrides_DO_NOT_OVERRIDE PASSED                                                                                      [  2%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_invoke PASSED                                                                                                            [  4%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_ainvoke PASSED                                                                                                           [  6%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_stream[model0] PASSED                                                                                                    [  8%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_stream[model1] PASSED                                                                                                    [ 10%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_astream[model0] PASSED                                                                                                   [ 12%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_astream[model1] PASSED                                                                                                   [ 14%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_stream_events_v3 PASSED                                                                                                  [ 16%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_astream_events_v3 PASSED                                                                                                 [ 18%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_invoke_with_model_override SKIPPED (model_override_value not specified.)                                                 [ 20%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_ainvoke_with_model_override SKIPPED (model_override_value not specified.)                                                [ 22%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_stream_with_model_override SKIPPED (model_override_value not specified.)                                                 [ 24%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_astream_with_model_override SKIPPED (model_override_value not specified.)                                                [ 26%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_batch PASSED                                                                                                             [ 28%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_abatch PASSED                                                                                                            [ 30%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_conversation PASSED                                                                                                      [ 32%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_double_messages_conversation PASSED                                                                                      [ 34%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_usage_metadata PASSED                                                                                                    [ 36%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_usage_metadata_streaming PASSED                                                                                          [ 38%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_stop_sequence PASSED                                                                                                     [ 40%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_tool_calling[model0] PASSED                                                                                              [ 42%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_tool_calling[model1] PASSED                                                                                              [ 44%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_tool_calling_async PASSED                                                                                                [ 46%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_bind_runnables_as_tools PASSED                                                                                           [ 48%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_tool_message_histories_string_content PASSED                                                                             [ 50%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_tool_message_histories_list_content PASSED                                                                               [ 52%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_tool_choice PASSED                                                                                                       [ 54%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_tool_calling_with_no_arguments PASSED                                                                                    [ 56%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_tool_message_error_status PASSED                                                                                         [ 58%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_few_shot_examples PASSED                                                                                      [ 60%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output[pydantic] PASSED                                                                                       [ 62%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output[typeddict] PASSED                                                                                      [ 64%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output[json_schema] PASSED                                                                                    [ 66%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output_async[pydantic] PASSED                                                                                 [ 68%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output_async[typeddict] PASSED                                                                                [ 70%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output_async[json_schema] PASSED                                                                              [ 72%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output_pydantic_2_v1 PASSED                                                                                   [ 74%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output_optional_param PASSED                                                                                  [ 76%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_json_mode SKIPPED (Test requires json mode support.)                                                                     [ 78%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_pdf_inputs SKIPPED (Model does not support PDF inputs.)                                                                  [ 80%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_audio_inputs SKIPPED (Model does not support audio inputs.)                                                              [ 82%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_image_inputs PASSED                                                                                                      [ 84%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_image_tool_message SKIPPED (Model does not support image tool message.)                                                  [ 86%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_pdf_tool_message SKIPPED (Model does not support PDF tool message.)                                                      [ 88%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_anthropic_inputs PASSED                                                                                                  [ 90%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_message_with_name PASSED                                                                                                 [ 92%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_agent_loop[model0] PASSED                                                                                                [ 94%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_agent_loop[model1] PASSED                                                                                                [ 96%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_stream_time SKIPPED (VCR not set up.)                                                                                    [ 98%]
tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_unicode_tool_call_integration PASSED                                                                                     [100%]

======================================================================================================= slowest 5 durations =======================================================================================================
9.20s call     tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_image_inputs
8.37s call     tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output_optional_param
8.17s call     tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output_pydantic_2_v1
5.31s call     tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_anthropic_inputs
4.98s call     tests/integration_tests/chat_models/test_anthropic_mantle.py::TestAnthropicMantleIntegration::test_structured_output_async[json_schema]
=========================================================================================== 40 passed, 10 skipped in 128.71s (0:02:08) ============================================================================================


```

</details>

## Scope / follow-ups

This is intentionally the **basic** class (static bearer key only). Planned follow-ups:
1. **Credential helper** — port the auto-refreshing short-term bearer-token provider (`_BedrockApiKeyProvider`, as tracked for `ChatOpenAIMantle`) so short-lived Bedrock API keys refresh transparently. The current class accepts a pre-minted static key only.
2. Per-model capability flags / structured-output + tool-choice validation as more Mantle Claude models are exercised.
3. Optional Bedrock Guardrails support via `extra_headers`, mirroring `ChatAnthropicBedrock`.

> Note: AI agent assistance was used in drafting this change.


michaelnchin (commented):
Thanks Piyush! Still having a look at testing but quick note from eyeballing:

> ChatAnthropicMantle talks to the bedrock-mantle endpoint and authenticates with a Bedrock API key (bearer/x-api-key) - no AWS SigV4 credential chain required.

We only support Bedrock API keys right now which is OK, but is there any reason not to support SigV4 as well? The [`AnthropicBedrockMantle`](https://github.com/anthropics/anthropic-sdk-python/blob/f5c30d0490fb7bcd8e0b65d8d8e63c0e7d1bfe59/src/anthropic/lib/bedrock/_mantle.py#L164) SDK client accepts both - for quick support, we could use it as a drop-in override for the default Anthropic client, similar to `ChatAnthropicBedrock`'s use of the`BedrockAnthropic` client:
```python
    # ... set URL (optional, AnthropicBedrockMantle, sets correct endpoint by default: https://github.com/anthropics/anthropic-sdk-python/blob/f5c30d0490fb7bcd8e0b65d8d8e63c0e7d1bfe59/src/anthropic/lib/bedrock/_mantle.py#L152)
    #  along with SigV4 / api key / anything else into _client_params 

    @cached_property
    def _client(self) -> Any:  # type: ignore[type-arg]
        return AnthropicBedrockMantle(**self._client_params)

    @cached_property
    def _async_client(self) -> Any:  # type: ignore[type-arg]
        return AsyncAnthropicBedrockMantle(**self._client_params)
```

Only real difference with the `AnthropicBedrockMantle`'s request creation is that it [sets the API key header](https://github.com/anthropics/anthropic-sdk-python/blob/f5c30d0490fb7bcd8e0b65d8d8e63c0e7d1bfe59/src/anthropic/lib/bedrock/_mantle.py#L246) as `{"Authorization": f"Bearer {api_key}"}` instead of the `x-api-key` that the base Anthropic client uses (but Mantle seems to accept both anyway). 

It also handles other aspects like creds validation and explicit response [error code mapping](https://github.com/anthropics/anthropic-sdk-python/blob/f5c30d0490fb7bcd8e0b65d8d8e63c0e7d1bfe59/src/anthropic/lib/bedrock/_client.py#L97) (which is useful as Mantle's error set is slightly different from that of Runtime). Might be other things I missed, but in general, the Mantle client that Anthropic provides looks fairly robust.

3coins:
@michaelnchin 
Ty for looking into this. Feel free to submit these updates on the PR directly.

libs/aws/langchain_aws/chat_models/anthropic.py
```diff
@@ -341,3 +341,158 @@ def _make_message_chunk_from_anthropic_event(
...
+        )
+
+        # Default the base URL to the region's Mantle endpoint unless the caller
+        # supplied one explicitly (either alias form).
+        if (
+            not values.get("base_url")
+            and not values.get("anthropic_api_url")
+            and region
+        ):
+            values["base_url"] = _MANTLE_BASE_URL_TEMPLATE.format(region=region)
```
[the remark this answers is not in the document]
michaelnchin: This will be handled with the switch to `AnthropicBedrockMantle` client (requires either region or base_url to be provided): https://github.com/anthropics/anthropic-sdk-python/blob/f5c30d0490fb7bcd8e0b65d8d8e63c0e7d1bfe59/src/anthropic/lib/bedrock/_mantle.py#L143-L152

Will update the PR branch soon

libs/aws/langchain_aws/chat_models/anthropic.py
```diff
@@ -443,8 +443,11 @@ def _set_mantle_defaults(cls, values: Any) -> Any:
         return values
 
     @model_validator(mode="after")
-    def _stamp_version(self) -> Self:
-        """Record the langchain-aws version in tracing metadata."""
+    def _set_model_profile(self) -> Self:
+        """Resolve the model profile and record the langchain-aws version."""
+        if self.profile is None:
+            model = re.sub(r"^[A-Za-z]{2}\.", "", self.model)
```
michaelnchin: Quick callout, this handles 2-char region IDs, but not others like `global`/`apac`/etc.

Should check against this static `MODEL_ID_GEO_PREFIXES` list that we maintain in utils: https://github.com/langchain-ai/langchain-aws/blob/main/libs/aws/langchain_aws/utils.py#L39-L41

michaelnchin: I'll update for this directly on the branch @ccurme

ccurme: thank you!

libs/aws/langchain_aws/chat_models/anthropic.py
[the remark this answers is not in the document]
michaelnchin: Done in https://github.com/langchain-ai/langchain-aws/pull/1190/commits/078b12be5487ace960912ac7bed7b0aed04e91ff

michaelnchin:
The stream/astream integ standard tests are flaking (~10% of the time) due to Sonnet 5 randomly activating adaptive thinking and returning signature-only thinking blocks. Going to update these two tests to use Haiku 4.5 instead.

libs/aws/langchain_aws/chat_models/anthropic.py
michaelnchin: Done in https://github.com/langchain-ai/langchain-aws/pull/1190/commits/d11279d6708715b609c8f3e1d60880a00f00457a

michaelnchin (commented):
Fixed - integ CI w/new Mantle tests is passing now: https://github.com/langchain-ai/langchain-aws/actions/runs/30591192733

gouveags:
We are actively integrating Claude on Bedrock Mantle into a LangGraph application, and this is exactly the upstream integration we need. We tested the same `AnthropicBedrockMantle` client with Sonnet 5 and Opus 5, including streaming, forced tool calls, cold and warm prompt caching, and SigV4 using task role credentials.

We did run into one authentication edge case. When bearer token environment variables are present, the SDK can choose bearer authentication even when the application should use task role SigV4. Right now, the only way around that is forcing the private `_use_sigv4` state. Having an explicit authentication mode in the SDK or integration would let us remove that workaround.

I also noticed that the PR is conflicting now. I am happy to help rebase it, run the current test suite, or work on a focused follow up if that would help, just let me know what would be most useful.

AlcebiadesFilho:
We've been running this same shape in production-bound code for a few weeks — a `ChatAnthropic` subclass over `AnthropicBedrockMantle`, SigV4 from a task role — so it's good to see it land properly upstream. Two things we measured against the live endpoint that might save someone else the trip.

**Bedrock Guardrails don't work on this endpoint, and they fail quietly.**

We assumed they would, honestly. #1178 landed guardrail support for `ChatAnthropicBedrock` a few days before this PR opened, and since that's also a non-Converse path where guardrails ride as `X-Amzn-Bedrock-Guardrail*` headers, the obvious guess is that Mantle takes the same headers. It doesn't:

| | result |
|---|---|
| `ApplyGuardrail(source=INPUT)` on the payload | `GUARDRAIL_INTERVENED` |
| Mantle `/anthropic/v1/messages`, no headers | `200`, model answers |
| Mantle, **with** the headers (present in `SignedHeaders`) | `200`, answers identically |
| Converse, same guardrail, same `DRAFT` version | `stopReason=guardrail_intervened` |

The first and last rows are there so the middle one means something — the payload really is blockable, and the guardrail really does work with that exact version, just not here. We tried it both signed and unsigned in case the service was dropping an unsigned header; same result either way.

What makes this worth a comment rather than a shrug is the failure mode. There's no `400`, no warning, and no `amazon-bedrock-guardrailAction` in the body. You get a clean `200` and a normal answer, which is indistinguishable from a guardrail that ran and found nothing. Anyone who reads #1178 and assumes parity here will believe they're covered.

We don't think the library can fix this — the endpoint is the one ignoring the headers — but a line in the class docstring would probably stop a few people from shipping a compliance claim they can't back. Might also be worth flagging to the Mantle service team.

**Small related thing:** the API rejects whitespace-only text blocks too, and the message is different, so a guard written against one will miss the other.

```
{"type": "text", "text": ""}      -> 400  text content blocks must be non-empty
{"type": "text", "text": "   "}   -> 400  text content blocks must contain non-whitespace text
{"type": "text", "text": "\n"}    -> 400  text content blocks must contain non-whitespace text
```

**On the ~10% stream flake** — we see a related shape on this endpoint, where `content` comes back as a mixed list of raw strings and dicts, in 9 of 30 real Sonnet 5 turns. We can't tell from out here whether it's your trigger or a cousin of it. In our case the interesting part is that it's harmless in the turn that produces it and only bites on the *next* one, when the raw `""` gets wrapped into an empty text block and the API rejects the whole request.

If a client-side normalization would be welcome upstream, we have one running and I'm happy to open it as a separate PR. If the library's position is that it shouldn't reshape provider output, that's a perfectly good answer and we'll keep ours local — just say which.

michaelnchin:
@AlcebiadesFilho thanks for the callout. Guardrails indeed are not yet supported on Mantle endpoint (see the Bedrock feature availability table here: https://docs.aws.amazon.com/bedrock/latest/userguide/endpoints.html). I was able to reproduce the same behavior with Mantle silently discarding the valid headers for `bedrock-runtime`, and I agree that this is fairly confusing so I'll add some pre-validation to both `ChatAnthropicMantle` and `ChatOpenAIMantle` in a follow-up.

michaelnchin:
@gouveags thanks, I actually ran into some similar difficulties with local unit test, related to the Anthropic SDK defaulting to env API key. Will also follow up separately to see if we can add an explicit SigV4 vs API key toggle.

michaelnchin (approved):
LGTM, thanks @3coins !

gouveags:
Great to see this merged, thanks @3coins and @michaelnchin!

@michaelnchin, I’d be happy to take the explicit SigV4-vs-API-key follow-up. We already have the SigV4-forced path running with task-role credentials, so I can contribute focused sync/async tests covering ambient API-key environment variables and an explicit auth-mode API that preserves today’s automatic behavior by default.

If you’d prefer an issue before code, I’ll open it through this repository’s Feature Request form and wait for maintainer assignment. Otherwise, let me know the API shape you’d prefer and I can prepare the PR.

gouveags:
Opened #1217 to track the auth-mode follow-up.

PR #1192, opened 2026-08-01
feat(aws): use Anthropic simplified prompt cache management in ChatBedrockConverse

michaelnchin:
Minimal PR distilled from the review discussion on #1162 - please see @billcai's analysis in the original PR for full context.

**TL;DR: For Anthropic models, we're updating ChatBedrockConverse's `cache_control` behavior to add a fourth cachePoint on the penultimate message, at the end of the static content.** 

This takes advantage of Bedrock's [simplified cache management feature for Claude](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-simplified), allowing Bedrock to search back through the message history for the longest cached prefix, and keep the conversation largely cache-readable as it grows after each turn. 

Additionally, since this extra cachePoint would hit Bedrock's hard limit of 4 per request, we check for any manual checkpoint already in the message history, and retain it over the 4th auto-checkpoint.

The existing 3-checkpoint `cache_control` behavior for non-Anthropic models remains unchanged.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -88,7 +88,7 @@ from langchain_aws.utils import (
 
 logger = logging.getLogger(__name__)
 
-
+_MAX_CACHE_POINTS = 4
 _MODEL_PROFILES = cast("ModelProfileRegistry", _PROFILES)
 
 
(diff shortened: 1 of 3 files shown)


michaelnchin (commented):
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/30682789938

PR #1193, opened 2026-08-01
fix(aws): include model name in invocation_params for Bedrock chat models

josepmafe:
## Summary

`ChatBedrock`, `ChatBedrockConverse`, and `ChatBedrockNovaSonic` never populate a `"model"` key in `invocation_params` (chat models) / `_identifying_params`, unlike `ChatOpenAI` (`langchain_openai`) and `ChatAnthropic` (`langchain_anthropic`), which both include `"model"` directly.

This breaks any tracing/observability integration that reads `invocation_params["model"]` (or the equivalent `_identifying_params`) directly to identify the model, such as MLflow. MLflow's LangChain autolog computes per-call token cost by looking up `SpanAttributeKey.MODEL`, which it sets from `kwargs["invocation_params"].get("model")`. For all three Bedrock chat classes, this is always `None`, so cost is silently never computed, even when the model has a known price in MLflow's (or LiteLLM's) pricing catalog.

Notably, each of these classes' own `_get_ls_params()` already anticipates this key — e.g. in `bedrock_converse.py`:
```python
ls_model_name=params.get("model") or self.base_model_id or self.model_id,
```
— but nothing ever populated `params["model"]`, so the fallback was silently always taken.

## Fix

- `ChatBedrock`, `ChatBedrockConverse`: add a `_get_invocation_params` override that merges in `"model"` (resolving `base_model_id` first where applicable), following the same override pattern used by `BaseChatOpenAI._get_invocation_params`.
- `ChatBedrockNovaSonic`: add `"model"` to the existing `_identifying_params` dict, which already had `"model_id"` under a Bedrock-specific key name but not the generic `"model"` key that tracing conventions expect.

No existing key is renamed or removed — `model_id`/`base_model_id` remain available as before, so this doesn't affect any other consumer (verified: `_identifying_params`/`_get_invocation_params` output only feeds `asdict()`/repr and `_get_ls_params`/LangSmith tracing for these classes; the LLM response cache key uses `_serialized`/`to_json()` instead, since all three classes have `is_lc_serializable() = True`, so caching is unaffected).

`ChatAnthropicBedrock` already inherits a correct `_identifying_params` from `ChatAnthropic` and needs no change.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -980,7 +980,7 @@ class ChatBedrock(BaseChatModel, BedrockBase):
     ) -> Dict[str, Any]:
         """Get the parameters used to invoke the model, for tracing purposes."""
         return {
-            "model": self.model_id,
+            "model": self.base_model_id or self.model_id,
             **super()._get_invocation_params(stop=stop, **kwargs),
         }
 
--- a/libs/aws/langchain_aws/chat_models/bedrock.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock.py
@@ -975,6 +975,15 @@ class ChatBedrock(BaseChatModel, BedrockBase):
         populate_by_name=True,
     )
 
+    def _get_invocation_params(
+        self, stop: Optional[List[str]] = None, **kwargs: Any
+    ) -> Dict[str, Any]:
+        """Get the parameters used to invoke the model, for tracing purposes."""
+        return {
+            "model": self.model_id,
+            **super()._get_invocation_params(stop=stop, **kwargs),
+        }
+
     def _get_ls_params(
         self, stop: Optional[List[str]] = None, **kwargs: Any
     ) -> LangSmithParams:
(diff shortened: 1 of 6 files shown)


libs/aws/langchain_aws/chat_models/bedrock.py
```diff
@@ -975,6 +975,15 @@ def lc_attributes(self) -> Dict[str, Any]:
         populate_by_name=True,
     )
 
+    def _get_invocation_params(
+        self, stop: Optional[List[str]] = None, **kwargs: Any
+    ) -> Dict[str, Any]:
+        """Get the parameters used to invoke the model, for tracing purposes."""
+        return {
+            "model": self.model_id,
```
michaelnchin: Like ChatBedrockConverse, ChatBedrock also sets `base_model_id`. Should prioritize it for better accuracy over raw `model_id`:
```suggestion
            "model": self.base_model_id or self.model_id,
```

michaelnchin (approved):
Thanks @josepmafe !

josepmafe:
> Thanks Josep Martinez (@josepmafe) !

Happy to help

PR #1195, opened 2026-08-05
fix(langgraph-checkpoint-aws): Support custom Boto config in AgentCore memory classes

### LINKED ISSUE #1194: AgentCoreMemorySaver: cannot configure botocore timeouts — `config` in `boto3_kwargs` raises TypeError
### Summary

`AgentCoreEventClient.__init__` hardcodes `config=` before splatting `**boto3_kwargs` into `boto3.client()`, so there is no supported way to set botocore timeouts or retries on the AgentCore Memory client. Passing `config=` raises `TypeError`.

This matters because `AgentCoreMemorySaver.aget_tuple()` offloads the synchronous `list_events` call to the default executor:

```python
async def aget_tuple(self, config: RunnableConfig) -> CheckpointTuple | None:
    return await run_in_executor(None, self.get_tuple, config)
```

An `asyncio.wait_for()` around that coroutine cancels the _await_, not the worker thread. With botocore's default `read_timeout` (60s) and default retries, a stalled AgentCore backend keeps a `ThreadPoolExecutor` thread occupied for minutes after the caller has given up. The default executor is bounded at `min(32, cpu_count + 4)`, so under concurrent load repeated stalls exhaust it and block unrelated executor work.

Bounding the botocore timeouts is the natural fix, but the seam is closed.

### Repro

```python
from botocore.config import Config
from langgraph_checkpoint_aws.agentcore.saver import AgentCoreMemorySaver

AgentCoreMemorySaver(
    "mem-abc",
    region_name="us-west-2",
    config=Config(connect_timeout=5, read_timeout=10),
)
```

```
TypeError: boto3.client() got multiple values for keyword argument 'config'
```

Version: `langgraph-checkpoint-aws==1.0.7`

### Cause

`langgraph_checkpoint_aws/agentcore/helpers.py` (1.0.7, ~L250):

```python
config = Config(
    user_agent_extra="x-client-framework:langgraph_agentcore_memory"
)
raw_client = boto3.client("bedrock-agentcore", config=config, **boto3_kwargs)
```

The docstring advertises `**boto3_kwargs: Additional arguments to pass to boto3.client`, but `config` is not actually forwardable.

### Suggested fix

Merge a caller-supplied `Config` instead of overwriting it, so `user_agent_extra` is preserved either way:

```python
user_agent_config = Config(
    user_agent_extra="x-client-framework:langgraph_agentcore_memory"
)
caller_config = boto3_kwargs.pop("config", None)
config = user_agent_config.merge(caller_config) if caller_config else user_agent_config
raw_client = boto3.client("bedrock-agentcore", config=config, **boto3_kwargs)
```

`botocore.config.Config.merge()` is the documented way to combine configs, and the caller's values take precedence for keys they set.

### Workaround

Constructing the saver, then replacing the underlying boto3 client on the private attribute. It works, but depends on internals that can move between releases — hence this issue.

michaelnchin:
Fixes #1194.

Users can now pass a custom boto `Config` on initialization of `AgentCoreMemorySaver` and `AgentCoreMemoryStore`, overriding any hardcoded values (with the exception of `user_agent_extra`, which gets prepended to the existing default `x-client-framework:langgraph_agentcore_memory`.)

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore/helpers.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/checkpoint/agentcore/helpers.py
@@ -53,6 +53,17 @@ MAX_PAYLOAD_ITEMS_PER_EVENT = 100  # max items in payload array
 MAX_PAYLOAD_BYTES_PER_EVENT = 10_000_000  # 10 MB max event size
 
 
+def merge_client_config(base_config: Config, boto3_kwargs: dict[str, Any]) -> Config:
+    """Handle merging caller-supplied Config into the final boto config."""
+    caller_config: Config | None = boto3_kwargs.pop("config", None)
+    if not isinstance(caller_config, Config):
+        return base_config
+    base_ua = getattr(base_config, "user_agent_extra", "") or ""
+    caller_ua = getattr(caller_config, "user_agent_extra", "") or ""
+    merged = base_config.merge(caller_config)
+    return merged.merge(Config(user_agent_extra=f"{caller_ua} {base_ua}".strip()))
+
+
 class BedrockAgentCoreClientWithRetry:
     """Wrapper around bedrock-agentcore client with retry logic.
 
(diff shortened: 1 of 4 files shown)


michaelnchin (commented):
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/30971699100

PR #1196, opened 2026-08-05
fix(langgraph-checkpoint-aws): Fix silent drops of custom config values in DynamoDB and Bedrock Sessions integrations

michaelnchin:
Similar to #1194, which fixed the AgentCore memory classes.

Updating the `BedrockSessionSaver`, `BedrockAgentRuntimeSessionClient`, `DynamoDBSaver`, `DynamoDBStore` Boto client creation utilities to fix a bug where nearly all custom boto Config values were being silently discarded in favor of a hardcoded configuration preset.


michaelnchin (commented):
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/30973673632

PR #1199, opened 2026-08-05
feat(langgraph-checkpoint-aws): add vector search to DynamoDBStore

### LINKED ISSUE #1198: Support DynamoDB native vector search
### Package

langgraph-checkpoint-aws

### Checked other resources

### Feature Description

AWS recently announced native real-time vector search capabilities for Amazon DynamoDB.
(https://aws.amazon.com/blogs/aws/amazon-dynamodb-now-supports-real-time-vector-search-at-any-scale/)

We would like to request native vector search integration within `langchain-aws`.

 Specifically:
**Vector Query Support in `DynamoDBStore`**: Native semantic search integration for `DynamoDBStore` (added in #764 for `langgraph-checkpoint-aws`), allowing LangGraph agents to perform vector-based memory retrieval directly on top of DynamoDB tables.

### Use Case

Previously, building RAG or agentic memory on top of DynamoDB data required complex secondary setups—such as DynamoDB Streams, S3 exports, and OpenSearch Service or third-party vector databases. This pattern introduced extra infrastructure cost, operational overhead, and latency sync issues.

With DynamoDB's native vector search, users can store both transactional operational data and vector embeddings in a single serverless table. Adding native support in `langchain-aws` will allow developers to build single-database RAG systems and LangGraph long-term memory solutions with zero-ETL, lower cost, and millisecond latencies on AWS.

```python

```

### Additional Context

_No response_

LeeroyHannigan:
Adds semantic search to DynamoDBStore using the DynamoDB vector index, so a LangGraph store can hold an item and its embedding in one table and query by similarity without a second vector database or a sync pipeline.

Opt in by passing an `index` config (embedding model, dims, optional distance function and fields). `setup()` then creates the vector index alongside the table and waits for it to become ACTIVE; `search(namespace, query="...")` runs nearest-neighbour search scoped to the namespace partition. Stores created without `index` behave exactly as before.

Design notes:

- Scores. SearchVectors returns a DISTANCE (lower is closer), while LangGraph's SearchItem.score convention is higher-is-more-relevant, so the distance is converted per distance function. Ranking order from the service is preserved.
- TopK. SearchVectors is a single top-K query with no pagination token, and the service accepts TopK only in 1..100. limit/offset from the BaseStore contract are satisfied by one fetch plus a client-side slice; a request for limit+offset beyond 100 raises ValueError naming the limit rather than silently returning a truncated tail. Filtered searches over-fetch (capped at 100) because value filters are applied client-side, so a filtered search can return fewer than `limit` matches once the cap truncates the over-fetch.
- Consistency. The index is maintained from the same write path as the item and is eventually consistent, the same model as a global secondary index.
- Errors. A failed search raises rather than returning an empty list: for a memory store, silently swallowing throttling or auth errors is indistinguishable from "no memories" and quietly degrades recall.
- Writes. The item and its embedding are written in a single request so an item can never be indexed without its vector, or vice versa.
- SDK floor. Vector index operations (SearchVectors, VectorIndexUpdates) are in the public service model from botocore 1.43.64, so boto3 is pinned to
  >= 1.43.64; on older SDKs the feature would fail at runtime with an
  unknown-operation error.

Testing:

- 64 unit tests covering score conversion per distance function, TopK bounds and the filter clamp, namespace scoping, the UpdateTable payload shape for index creation, waiter behaviour including timeout, and error propagation.
- Live integration tests against real DynamoDB (opt in with DDB_VECTOR_INTEG=1): table creation with a vector index, writes with embeddings, semantic search round trip, and partition scoping. 2 passed against us-east-1.
- ruff, ruff format and mypy clean.


michaelnchin (approved):
LGTM - thanks for your work on this @LeeroyHannigan !

PR #1200, opened 2026-08-05
feat(aws): add DynamoDBVectorStore backed by DynamoDB vector indexes

LeeroyHannigan:
Documents are stored as DynamoDB items and searched via a table-level vector index (SearchVectors, approximate nearest neighbor). No separate vector database: the same table serves operational data and retrieval.

- add_texts embeds via embed_documents, writes through BatchWriteItem with unprocessed-item retry and the 25-item service batch limit
- similarity_search / _with_score / _by_vector via SearchVectors; raw distance surfaced as the score, TopK bounds (1..100) enforced loudly
- relevance score conversion per distance function (COSINE, EUCLIDEAN, DOT_PRODUCT) for similarity_search_with_relevance_scores
- setup provisions the table + index, or retrofits the index onto an existing table via UpdateTable VectorIndexUpdates; no SearchSchema so search is global over the corpus
- vector index is eventually consistent (GSI model), documented
- 19 unit tests, fully mocked; ruff/format/mypy clean

### DIFF
--- a/libs/aws/langchain_aws/vectorstores/dynamodb/base.py
+++ b/libs/aws/langchain_aws/vectorstores/dynamodb/base.py
@@ -36,6 +36,9 @@ _VECTOR_ATTR = "embedding"
 _CONTENT_ATTR = "page_content"
 _METADATA_ATTR = "metadata"
 
+# Attribute names this store owns; a partition attribute must not collide.
+_RESERVED_ATTRS = frozenset({_PK_ATTR, _VECTOR_ATTR, _CONTENT_ATTR, _METADATA_ATTR})
+
 
 def _to_dynamodb_compatible(value: Any) -> Any:
     """Recursively convert floats to Decimal for DynamoDB serialization.
@@ -86,10 +89,20 @@ class DynamoDBVectorStore(VectorStore):
     To use, provide a table name and an embedding function. The table and
     vector index are created on first write if they don't exist.
 
-    Limitations: metadata filtering is not supported (a ``filter`` argument
-    is rejected rather than silently ignored), and maximal marginal
-    relevance (MMR) search is not implemented. SearchVectors returns at
-    most the top 100 matches per query.
+    Pass ``partition_attribute`` to scope searches. The index is then created
+    with a ``SearchSchema`` partition key on that document metadata field, and
+    each search examines only one value of it rather than the whole corpus,
+    which is how the index scales search throughput. The partition value is
+    supplied per search via ``filter={partition_attribute: value}``, or once
+    via ``default_partition_value``.
+
+    Limitations: inline filter attributes are not yet supported, so ``filter``
+    accepts only the partition key (any other key is rejected rather than
+    silently ignored), and maximal marginal relevance (MMR) search is not
+    implemented. SearchVectors returns at most the top 100 matches per query.
+    ``Dimensions``, ``DistanceFunction``, ``SearchSchema`` and the projection
+    are all fixed when the index is created; this store validates the first
+    three against an existing index rather than proceeding with a mismatch.
 
     Example:
         ```python
@@ -112,6 +125,8 @@ class DynamoDBVectorStore(VectorStore):
         embedding: Embeddings,
         index_name: str = "documents-vector-index",
         distance_function: Literal["COSINE", "EUCLIDEAN", "DOT_PRODUCT"] = "COSINE",
+        partition_attribute: Optional[str] = None,
+        default_partition_value: Optional[str] = None,
         create_table_if_not_exist: bool = True,
         relevance_score_fn: Optional[Callable[[float], float]] = None,
         region_name: Optional[str] = None,
@@ -129,6 +144,21 @@ class DynamoDBVectorStore(VectorStore):
             index_name: Name of the vector index on the table.
             distance_function: Distance function for the vector index.
                 One of "COSINE" (default), "EUCLIDEAN", "DOT_PRODUCT".
+            partition_attribute: Name of a document metadata field to use as
+                the vector index partition key. When set, the index is created
+                with a ``SearchSchema`` HASH element on this attribute and each
+                search is scoped to a single value of it, which lets the index
+                scale search throughput horizontally instead of examining the
+                whole corpus. Prefer low-to-medium cardinality values such as
+                a collection, category or tenant. This is part of the index
+                schema and cannot be changed after the index is created.
+                Every search must then supply the value, via
+                ``filter={partition_attribute: value}`` or
+                ``default_partition_value``.
+            default_partition_value: Value to use for ``partition_attribute``
+                when a document's metadata omits it and when a search supplies
+                no ``filter``. Only valid together with
+                ``partition_attribute``.
             create_table_if_not_exist: Create the table and vector index on
                 first use if they don't exist. Default is True.
             relevance_score_fn: Override for the relevance score conversion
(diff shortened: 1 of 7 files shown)


libs/aws/langchain_aws/vectorstores/dynamodb/base.py
```diff
@@ -0,0 +1,567 @@
...
+        )
+        self._wait_for_vector_index()
+
+    def _vector_index_spec(self, dimensions: int) -> dict:
+        # No SearchSchema: the index spans the whole table so searches are
+        # global over the corpus (unlike a partition-scoped memory store).
+        return {
+            "IndexName": self.index_name,
+            "VectorAttribute": {"AttributeName": _VECTOR_ATTR},
+            "Projection": {"ProjectionType": "ALL"},
```
michaelnchin, line 379: Most of the index spec fields are user configurable, but we're specifically hardcoding `ProjectionType` here. I'm curious, do we require it to be `ALL` for something, or can we add an optional `projection_type` class arg too?

Specifically mentioning this because the reqs and limitations [docs page](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/VectorSearch.Requirements.html) has this point:
> SearchVectors responses are limited to 16 MB. Pagination is not supported. Because each result includes the projected attributes, using ProjectionType: ALL with large items and a high TopK value can approach this limit. If your items are large, use a narrower projection or reduce TopK.

Given that DDB doesn't allow changing the `ProjectionType` on the index after creation, I think it would be good to allow users to configure `KEYS_ONLY`/`INCLUDE` to avoid SearchVectors request failures on the hard cap.

libs/aws/langchain_aws/vectorstores/dynamodb/base.py
```diff
@@ -0,0 +1,567 @@
...
+        """Retrofit the vector index onto an existing table."""
+        self.client.update_table(
+            TableName=self.table_name,
+            VectorIndexUpdates=[{"Create": self._vector_index_spec(dimensions)}],
+        )
+        self._wait_for_vector_index()
+
+    def _vector_index_spec(self, dimensions: int) -> dict:
+        # No SearchSchema: the index spans the whole table so searches are
+        # global over the corpus (unlike a partition-scoped memory store).
```
michaelnchin: Are we omitting `SearchSchema` config entirely because of the current lack of support for filtering? (Is filtering planned to be added at some point?)

Main concern is that this is also blocking use of the `HASH` partition key that can be used to improve scaling on large indices, this has to be submitted inside `SearchSchema`: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/VectorSearch.html#VectorSearch.SearchSchema

Would it make sense to accept the partition key on its own as a user arg and use it to create/submit `SearchSchema` without filtering for now?

libs/aws/langchain_aws/vectorstores/dynamodb/base.py
```diff
@@ -0,0 +1,567 @@
...
+                    f"DynamoDB table '{self.table_name}' does not exist and "
+                    "create_table_if_not_exist is False."
+                )
+            self._create_table(dimensions=dimensions)
+            return
+        for idx in indexes:
+            if idx.get("IndexName") == self.index_name:
+                existing_dims = idx.get("Dimensions")
+                if existing_dims is not None and existing_dims != dimensions:
+                    raise ValueError(
```
michaelnchin, line 428: It's good that we validate `Dimensions` against the existing index here. Since we have this already, do we want to add a similar check for the `DistanceFunction`?

Example from testing where this check might have been useful:  

I used `DynamoDBVectorStore` to create a table/index with the default `COSINE` distance_function. Then, I created a second DDBVectorStore set to `EUCLIDEAN` and pointed at the same table as the first store. 

The Euclidean store was able to run `add_texts` and searches against the initial cosine index without any errors, BUT the searches actually ran under the index's cosine. So when `_select_relevance_score_fn` applied the Euclidean calculation anyway based on `distance_function`,  `similarity_search_with_relevance_scores` ended up returning incorrect scores (0.69 vs 0.62 for the same doc+query run on each store).

michaelnchin (commented):
Thanks @LeeroyHannigan ! Tested out the new vector store and seems to be working great overall. I have a couple of questions about configuration and validation (given the largely immutable nature of the indexes):

libs/aws/langchain_aws/vectorstores/dynamodb/base.py
LeeroyHannigan: Great call out, adding that to the Store also.

libs/aws/langchain_aws/vectorstores/dynamodb/base.py
LeeroyHannigan: I went a different way than you suggested, and I think it gets you more. The docs are explicit that attributes not projected into the index can never be returned by SearchVectors, so narrowing the index projection is the choice that removes capability, not ALL. With KEYS_ONLY this store could not build a Document from a search response at all and would need a BatchGetItem hydration round trip on every query.

SearchVectors accepts a ProjectionExpression. We were passing none, which is the gap your comment exposed. b243b57 now sends one covering exactly the three attributes _item_to_document reads, so a response is bounded to the id, page content and metadata regardless of what else an item carries, which also helps anyone sharing the table with another workload. Names are aliased through ExpressionAttributeNames so a document field colliding with a reserved word cannot break the request. Keeping the index at ALL means adding an attribute later stays a one-line change instead of an index rebuild.

If you still want the index-level knob for people who need a physically smaller index, I am happy to add projection_type and an include list on top. My read is that it is not needed for the 16MB concern now, but say the word.

libs/aws/langchain_aws/vectorstores/dynamodb/base.py
LeeroyHannigan: partition_attribute names a metadata field to use as the vector index partition key, with an optional default_partition_value. The index is created with the HASH element, each search sends SearchConditionExpression, and because the schema is immutable I also validate it against an existing index in all three mismatch directions rather than proceeding toward a search the index cannot satisfy.

Worth flagging the contract you implied: once a HASH element exists the service requires the condition on every call, so similarity_search(query) raises unless default_partition_value is set. I did not see a cleaner way to express that through the VectorStore interface.

On filtering, yes it is planned. Inline filters are equality-only in the service today, and they use the same SearchConditionExpression plumbing, so filter stays the entry point: it currently accepts the partition key only, and any other key is rejected rather than silently dropped.

michaelnchin (commented):
@LeeroyHannigan thanks for the quick turnaround!

Just one issue I found testing the update for partition key support. It looks like DDB `CreateTable` and `UpdateTable` both require the key to be declared as an attribute if we add it to the index. Getting this error when trying to run `add_texts` with a `partition_attribute` specified on the current build:
> botocore.exceptions.ClientError: An error occurred (ValidationException) when calling the CreateTable operation: One or more parameter values were invalid: One element in SearchSchema is not defined in attribute definitions

I was able to get it working by appending `{"AttributeName": self.partition_attribute, "AttributeType": "S"` to `AttributeDefinitions` in both the `_create_table` and `_create_vector_index` methods. Will add this in a commit, but let me know if you have any concerns with this 👍

libs/aws/langchain_aws/vectorstores/dynamodb/base.py
michaelnchin: Sounds good to me - thanks Lee

michaelnchin (approved):
^ @LeeroyHannigan let me know if the last change is okay, but otherwise this PR LGTM now!

LeeroyHannigan:
> ^ Lee (@LeeroyHannigan) let me know if the last change is okay, but otherwise this PR LGTM now!

Thanks @michaelnchin , your fix is 100% correct, thanks for catching it.

PR #1203, opened 2026-08-06
feat(aws): credential-derived auth for ChatOpenAIMantle

3coins:
## Summary

Adds a shared `_BedrockApiKeyProvider` utility (`langchain_aws.utils`) that mints and transparently refreshes **short-term Bedrock API keys from the standard AWS credential chain** (via `aws-bedrock-token-generator`), and wires it into `ChatOpenAIMantle`.

When no static Bedrock API key is provided, authentication is now derived from ordinary AWS credentials — explicit keys, a named profile, or the default chain (env / assumed role / IRSA) — instead of requiring a manually-minted `AWS_BEARER_TOKEN_BEDROCK`. Keys are cached and refreshed near expiry, with the TTL capped by the underlying credential lifetime.

Auth precedence (fully backward compatible):
`api_key` (explicit) → static `bedrock_api_key` / `AWS_BEARER_TOKEN_BEDROCK` → credential-derived provider.

The `_BedrockApiKeyProvider` implementation is derived from @rishabhagrawal1's work in the closed PR #1018, abstracted here into a shared utility for reuse across the Bedrock Mantle chat models. It is the credential-refresh follow-up noted in #1189.

## Reviewer verification

The credential-derived path has a dedicated, opt-in live integration test — no static bearer token needed, just AWS credentials + a region for a Mantle-enabled account. From `libs/aws/`:

```
MANTLE_CREDS_E2E=1 AWS_REGION=us-east-1 \
  uv run --group test --group test_integration \
  pytest tests/integration_tests/chat_models/test_openai.py -k credential_derived -v
```

It asserts the credential-derived provider is installed, mints a real short-term key from your AWS credentials, and invokes `openai.gpt-oss-120b`. The static-key standard suite is unchanged; unit tests (mocked) cover generation, caching, and refresh behavior.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/openai.py
+++ b/libs/aws/langchain_aws/chat_models/openai.py
@@ -31,6 +31,10 @@ from typing_extensions import Self
 
 from langchain_aws._version import _add_langchain_aws_version
 from langchain_aws.data._profiles import _PROFILES
+from langchain_aws.utils import (
+    _BEDROCK_API_KEY_MAX_TTL_SECONDS,
+    _BedrockApiKeyProvider,
+)
 
 _MANTLE_BASE_URL_TEMPLATE = "https://bedrock-mantle.{region}.api.aws/v1"
 
@@ -43,18 +47,29 @@ def _get_default_model_profile(model_name: str) -> ModelProfile:
     return default.copy()
 
 
+def _plain_secret(value: Any) -> str | None:
+    """Return the plain string for a ``SecretStr``/``str``/``None`` value."""
+    if value is None:
+        return None
+    if isinstance(value, SecretStr):
+        return value.get_secret_value()
+    return str(value)
+
+
 class ChatOpenAIMantle(BaseChatOpenAI):
     """OpenAI-compatible GPT/open-weight models via the Amazon Bedrock Mantle endpoint.
 
     Talks to the ``bedrock-mantle`` OpenAI-compatible endpoint
     (``bedrock-mantle.{region}.api.aws``) using the OpenAI Python SDK.
     Authentication uses an Amazon Bedrock API key (bearer token) rather than
-    AWS SigV4.
+    AWS SigV4. Provide it directly (``bedrock_api_key`` /
+    ``AWS_BEARER_TOKEN_BEDROCK``), or omit it and let short-term keys be derived
+    from your AWS credentials and refreshed transparently.
 
     See the [LangChain docs for `ChatOpenAI`](https://docs.langchain.com/oss/python/integrations/chat/openai)
     for tutorials and feature walkthroughs — the same features apply here.
 
-    Example:
+    Example (static bearer token):
         ```python
         # pip install "langchain-aws[openai]"
         # export AWS_BEARER_TOKEN_BEDROCK="your-bedrock-api-key"
@@ -65,6 +80,21 @@ class ChatOpenAIMantle(BaseChatOpenAI):
         model = ChatOpenAIMantle(
             model="openai.gpt-oss-120b",
             region_name="us-east-1",
+        )
+        model.invoke("What is 2 + 2?")
+        ```
+
+    Example (derive short-term keys from AWS credentials):
+        ```python
+        # pip install "langchain-aws[openai]"
+        # Uses the standard boto3 credential chain (env, profile, role, IRSA).
+
+        from langchain_aws import ChatOpenAIMantle
+
+        model = ChatOpenAIMantle(
+            model="openai.gpt-oss-120b",
+            region_name="us-east-1",
+            # optionally: credentials_profile_name="my-profile"
         )
         model.invoke("What is 2 + 2?")
         ```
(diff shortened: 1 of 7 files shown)


rishabhagrawal1:
Thanks @3coins for putting this PR out and glad to see the `_BedrockApiKeyProvider` approach from #1018 land as a shared utility. One architectural choice we should discuss - the reason the Anthropic side (#1190) doesn't need this change is that auth there is owned entirely by Anthropic SDK.
[`ChatAnthropic`](https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/langchain_anthropic/chat_models.py) lets subclasses override how the client is built (`_client` / `_async_client` are overridable factories), so `ChatAnthropicMantle` (#1190) simply returns [`AnthropicBedrockMantle`](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/bedrock/_mantle.py) - a client that picks bearer or SigV4 on its own (explicit key → bearer; AWS keys or profile → SigV4; env token → bearer; otherwise the default AWS chain → SigV4) and signs every request with botocore's auto-refreshing credentials. No short-term key handling required at LangChain layer.

The OpenAI SDK has since caught up: as of v2.44.0 ([openai/openai-python#3398](https://github.com/openai/openai-python/pull/3398)), `OpenAI(provider=bedrock(...))` gives the same coverage natively - bearer keys and full SigV4 with auto-refreshing credentials ([`providers/bedrock.py`](https://github.com/openai/openai-python/blob/main/src/openai/providers/bedrock.py), [`lib/_bedrock_auth.py`](https://github.com/openai/openai-python/blob/main/src/openai/lib/_bedrock_auth.py)), signed per request against the Mantle endpoint (the base URL is overridable). The reason we can't use that here is a limitation in how `langchain-openai` constructs its client. Anthropic makes Bedrock a separate client class, so a subclass can just swap it in. OpenAI instead configures Bedrock on the regular client at construction time (`provider=bedrock(...)`) but [`BaseChatOpenAI`](https://github.com/langchain-ai/langchain/blob/master/libs/partners/openai/langchain_openai/chat_models/base.py) always constructs the client itself with `base_url` and `api_key` in `validate_environment`, and the SDK doesn't allow provider option together with those.
The callable `api_key` this PR uses works within that limit. It refreshes the token, but it's key-based auth. The upstream fix is small - move the two inline client constructions into overridable builder methods (defaults unchanged), plus one hook to skip the api_key requirement when the client owns auth. With that, `ChatOpenAIMantle` would just return the Bedrock-configured client which will handle SigV4 auth. This will leave both Mantle integrations with the same architecture: the vendor SDK owns authentication. So this PR should be the right path today, with a future path of `langchain-openai` change. Happy to file the upstream proposal and own the changes if there's interest.

For concreteness, the entire `langchain-openai` change (in [`chat_models/base.py`](https://github.com/langchain-ai/langchain/blob/master/libs/partners/openai/langchain_openai/chat_models/base.py)) is three redirected lines plus three default-preserving methods:

```diff
         if not self.client:
-            if sync_api_key_value is None:
+            if self._requires_api_key() and sync_api_key_value is None:
                 ...
-                self.root_client = openai.OpenAI(**client_params, **sync_specific)
+                self.root_client = self._build_sync_root_client(client_params, sync_specific)
                 self.client = self.root_client.chat.completions
         ...
-            self.root_async_client = openai.AsyncOpenAI(**client_params, **async_specific)
+            self.root_async_client = self._build_async_root_client(client_params, async_specific)
             self.async_client = self.root_async_client.chat.completions

+    def _requires_api_key(self) -> bool:
+        return True
+
+    def _build_sync_root_client(self, client_params, sync_specific) -> openai.OpenAI:
+        return openai.OpenAI(**client_params, **sync_specific)      # old line, verbatim
+
+    def _build_async_root_client(self, client_params, async_specific) -> openai.AsyncOpenAI:
+        return openai.AsyncOpenAI(**client_params, **async_specific)  # old line, verbatim
```

After which `ChatOpenAIMantle` reduces to roughly:

```python
class ChatOpenAIMantle(BaseChatOpenAI):
    def _requires_api_key(self) -> bool:
        return False   # the client owns auth

    def _build_sync_root_client(self, *_):
        return openai.OpenAI(provider=bedrock(
            region=self.region_name, base_url=self.base_url,
            api_key=...,                      # static bearer if provided
            access_key_id=..., profile=...,   # else SigV4 via keys/profile/default chain
        ))

    def _build_async_root_client(self, *_):
        return openai.AsyncOpenAI(provider=bedrock(...))
```

Existing behavior for `ChatOpenAI` and other subclasses is unchanged, and the base class still derives `client = root_client.chat.completions`, so the invariant stays owned in one place.

libs/aws/langchain_aws/chat_models/openai.py
```diff
@@ -133,11 +204,35 @@ def _set_mantle_defaults(cls, values: Any) -> Any:
...
         if not values.get("api_key") and not values.get("openai_api_key"):
-            key = values.get("bedrock_api_key") or os.getenv("AWS_BEARER_TOKEN_BEDROCK")
-            if key:
-                values["api_key"] = key
+            static_key = values.get("bedrock_api_key") or os.getenv(
+                "AWS_BEARER_TOKEN_BEDROCK"
+            )
+            if static_key:
+                values["api_key"] = static_key
+            elif region:
```
rishabhagrawal1, line 219: The no-region guard above (lines 194–203) only runs when `base_url` is not set. With an explicit `base_url`, no region, and no static key, this `elif region:` is skipped and no `api_key` will be set. `BaseChatOpenAI` then resolves the key from the `OPENAI_API_KEY` env var if present and uses it against the configured base_url; if absent, construction fails with the OpenAI SDK's "Missing credentials... set `OPENAI_API_KEY`" error, which doesn't mention the missing region.

Adding an error msg here will help avoid it - 
```python
else:
    raise ValueError(
        "ChatOpenAIMantle has no static Bedrock API key and cannot create "
        "short-term keys without a region. Set `region_name` or "
        "`AWS_REGION`/`AWS_DEFAULT_REGION`, pass a static `bedrock_api_key`, "
        "or provide an explicit `api_key` for your endpoint."
    )
```

michaelnchin (approved):
LGTM - thanks Piyush!

PR #1205, opened 2026-08-07
fix(aws): Support streaming for Kimi models on ChatBedrockConverse

### LINKED ISSUE #1204: ChatBedrockConverse: streaming silently disabled for Moonshot Kimi models (missing _get_streaming_support allowlist entry)
### Package

langchain-aws

### Checked other resources

### Example Code

```python
python

model_parts = model_id.split(".")
values["provider"] = (
    model_parts[-2] if len(model_parts) > 1 else model_parts[0]
)
```

```shell

```

## Summary

`ChatBedrockConverse` silently falls back to the non-streaming `Converse` API for **Moonshot Kimi** models (e.g. `moonshotai.kimi-k2.5`, `moonshotai.kimi-k2-thinking`), even though AWS Bedrock publishes these as valid, documented model IDs and serves `ConverseStream` for them correctly.

`disable_streaming` resolves to `True` for these models with no warning. There are actually **two independent defects** here, and only one of them is load-bearing.

## Environment

- `langchain-aws` 1.6.1
- File: `langchain_aws/chat_models/bedrock_converse.py`

## Root cause #1 (load-bearing): missing allowlist entries

`_get_streaming_support(provider, model_id_lower)` has no branch for `moonshot` / `kimi`, so it returns `False` and `set_disable_streaming` sets `disable_streaming=True`.

This is the actual bug. Even when the provider is parsed **correctly**, the model is still treated as non-streaming.

## Root cause #2 (not load-bearing, but real): provider misparse

`set_disable_streaming` (lines 874–877 in 1.6.1) extracts the provider as `model_parts[-2]`:

```python
model_parts = model_id.split(".")
values["provider"] = (
    model_parts[-2] if len(model_parts) > 1 else model_parts[0]
)
```

`[-2]` is deliberate — it strips geo / inference-profile prefixes so that `eu.anthropic.claude-...` and `us.meta.llama3-...` resolve to `anthropic` / `meta` instead of the region. It works for every model whose *name* contains no dot, but it breaks for the newer dotted-version naming convention that Moonshot and OpenAI both use:

| model id                                       | `split(".")[-2]` | `split(".")[0]` | correct provider? |
|------------------------------------------------|------------------|-----------------|-------------------|
| `anthropic.claude-sonnet-4-6`                  | `anthropic`      | `anthropic`     | OK                |
| `eu.anthropic.claude-sonnet-4-5-20250929-v1:0` | `anthropic`      | `eu`            | OK                |
| `us.meta.llama3-1-70b-instruct-v1:0`           | `meta`           | `us`            | OK                |
| `mistral.mistral-large-2407-v1:0`              | `mistral`        | `mistral`       | OK                |
| `moonshotai.kimi-k2.5`                          | `kimi-k2`        | `moonshotai`    | **WRONG**         |
| `moonshotai.kimi-k2-thinking`                  | `moonshotai`     | `moonshotai`    | OK                |
| `openai.gpt-5.6-terra`                          | `gpt-5`          | `openai`        | **WRONG**         |

### The row that isolates the real cause

`moonshotai.kimi-k2-thinking` parses **correctly** to `moonshotai` and *still* ends up with `disable_streaming=True`, because `moonshotai` isn't in the allowlist either. Same vendor, one ID parsed right and one parsed wrong — identical (wrong) outcome. That demonstrates the parse bug is not what's breaking streaming here; the missing allowlist entry is.

## Note on 1.7.0

1.7.0 fixes the parse (the `model_parts[-2]` logic is replaced by a helper that checks geo prefixes explicitly):

```python
def parse_model_provider(model_id: str) -> str:
    """Extract the provider from a Bedrock model ID."""
    parts = model_id.split(".", maxsplit=2)
    if len(parts) > 1 and parts[0].lower() in MODEL_ID_GEO_PREFIXES:
        return parts[1]
    return parts[0]
```

…but `moonshot` / `kimi` are still absent from the streaming allowlist in 1.7.0, so `disable_streaming` still resolves to `True`. 1.7.0 does add a warning that describes the manual override (`pass disable_streaming=False to override`), i.e. it converts a silent failure into a logged one with identical behaviour.

## Requested change

Add Moonshot Kimi models to `_get_streaming_support` so they get streaming (or `no_tools`, whichever AWS supports) by default, matching how the other Bedrock providers are handled. Fixing the parse alone is not sufficient.

## Reproduction

```python

llm = ChatBedrockConverse(model="moonshotai.kimi-k2.5", region_name="us-east-1")
print(llm.provider)           # -> "kimi-k2"  (misparsed)
print(llm.disable_streaming)  # -> True       (the actual problem)

llm2 = ChatBedrockConverse(model="moonshotai.kimi-k2-thinking", region_name="us-east-1")
print(llm2.provider)          # -> "moonshotai" (parsed correctly)
print(llm2.disable_streaming) # -> True         (still disabled: missing allowlist entry)
```

michaelnchin:
Fixes: #1204

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -893,6 +893,9 @@ class ChatBedrockConverse(BaseChatModel):
             or
             # Qwen3 models
             (provider == "qwen" and "qwen3" in model_id_lower)
+            or
+            # Kimi models
+            (provider in ("moonshot", "moonshotai") and "kimi" in model_id_lower)
         ):
             return True
         elif (
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -829,6 +829,8 @@ def test_invocation_params_model_prefers_base_model_id() -> None:
         ("openai.gpt-oss-120b-1:0", False),
         ("openai.gpt-oss-20b-1:0", False),
         ("qwen.qwen3-32b-v1:0", False),
+        ("moonshotai.kimi-k2.5", False),
+        ("moonshot.kimi-k2-thinking", False),
     ],
 )
 def test_set_disable_streaming(
--- a/libs/aws/tests/unit_tests/test_utils.py
+++ b/libs/aws/tests/unit_tests/test_utils.py
@@ -770,6 +770,8 @@ def test_thinking_in_params(params: dict, expected: bool) -> None:
         ("us-gov.anthropic.claude-haiku-4-5-20251001-v1:0", "anthropic"),
         ("minimax.minimax-m2.5", "minimax"),
         ("us.minimax.minimax-m2.5", "minimax"),
+        ("moonshotai.kimi-k2.5", "moonshotai"),
+        ("moonshot.kimi-k2-thinking", "moonshot"),
     ],
 )
 def test_parse_model_provider(model_id: str, expected_provider: str) -> None:


michaelnchin (commented):
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/31216476666

PR #1208, opened 2026-08-12
fix(langgraph-checkpoint-aws): validate an existing vector index in setup()

LeeroyHannigan:
`DynamoDBStore.setup()` matched an existing vector index on `IndexName` alone, so a store pointed at a table whose index was created with different geometry was accepted silently. `Dimensions` and `DistanceFunction` are both immutable after index creation, so neither can be reconciled at runtime.

Both mismatches were silent, in different ways:

- **Dimensions.** A wrong count passed `setup()` and also passed the store's put-time check, which compares the embedding model's output against the *configured* `dims` rather than against the index. It then failed later at write time with a service error that does not name the index.
- **DistanceFunction.** This never failed at all. The service scores using the index's metric, while `_distance_to_score` converts using the store's configured metric, so `search()` returns wrongly-scaled scores in the correct order. Any caller thresholding on `score` to decide whether a memory is relevant gets silently wrong answers.

`setup()` now looks the index up by name and validates both fields, raising `ValidationError` that names the existing value, the configured value, and the fix. An index whose description omits those fields is accepted unchanged.

### Testing

Five unit tests: both mismatches raise, a matching index neither raises nor recreates, a description without geometry fields is accepted, and a negative control confirms the create path still fires when the index is absent (that path was refactored away from name matching).

Gates on this branch: 1077 unit tests pass, `ruff check` clean, `ruff format --check` clean, `mypy` clean on both `langgraph_checkpoint_aws` and `tests`.

### Context

Found while addressing the same defect class in the `DynamoDBVectorStore` review (#1200), where @michaelnchin reproduced the distance-function case: two stores over one table, the second set to `EUCLIDEAN` against a `COSINE` index, returning 0.69 vs 0.62 for the same document and query. The equivalent fix for that PR is in its own branch. This one covers the `DynamoDBStore` shipped in #1199.

### DIFF
--- a/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/table.py
+++ b/libs/langgraph-checkpoint-aws/langgraph_checkpoint_aws/store/dynamodb/table.py
@@ -14,7 +14,7 @@ from typing import Any, cast
 
 from botocore.exceptions import ClientError
 
-from .exceptions import TableCreationError
+from .exceptions import TableCreationError, ValidationError
 from .vector import _VECTOR_ATTR, _VECTOR_INDEX_NAME
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


michaelnchin (approved):
LGTM - thanks @LeeroyHannigan !

PR #1216, opened 2026-08-18
fix(aws): stop zeroing cache_creation when cacheDetails is present

JonZeolla:
Bedrock only includes cacheDetails on Converse calls that write to the cache, and its entries break down the same tokens counted in cacheWriteInputTokens. Zeroing cache_creation when the breakdown is present (introduced in #1098) therefore under-reported cache writes as 0 on every write call. Keep the write count via
max(cacheWriteInputTokens, ttl sum) so partial cacheDetails entries cannot under-report it either; reads are never double-counted because cacheDetails is absent on cache-hit calls.

Verified against raw Converse API responses: a cold call reports cacheWriteInputTokens=10353 with cacheDetails=[{inputTokens:10353, ttl:"5m"}]; the warm call reports cacheReadInputTokens with no cacheDetails key.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -2335,7 +2335,11 @@ def _extract_usage_metadata(response: Dict[str, Any]) -> UsageMetadata:
         "cache_creation": cache_write_input_tokens,
     }
 
-    # Parse per-TTL cache breakdown from cacheDetails (if present)
+    # Parse per-TTL cache breakdown from cacheDetails (if present). Bedrock only
+    # includes cacheDetails on calls that write to the cache, and its entries
+    # break down the same tokens counted in cacheWriteInputTokens — so
+    # cache_creation keeps reporting the write. max() guards against a partial
+    # cacheDetails under-reporting it.
     cache_details = usage_dict.get("cacheDetails", [])
     if cache_details:
         cache_5m = sum(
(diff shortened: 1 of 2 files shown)


JonZeolla:
Closing this — after digging further, the zeroing this PR removes is intentional, not a bug.

The pattern was copied from `langchain-anthropic` (which #1098 cites as its reference), where the same logic is explicit:

```python
if specific_cache_creation_tokens > 0:
    # Remove generic key to avoid double counting cache creation tokens
    input_token_details["cache_creation"] = 0
```

The `input_token_details` keys are treated as disjoint buckets: when the per-TTL breakdown is present, cache-write tokens live in `ephemeral_5m_input_tokens` / `ephemeral_1h_input_tokens` and `cache_creation` is zeroed, so a consumer that sums keys (e.g. a cost calculator pricing the 5m and 1h tiers differently) counts each token exactly once. This PR's `max()` change would make such consumers double-count writes (`cacheDetails` breaks down the same tokens already counted in `cacheWriteInputTokens`).

The correct consumer-side read for "tokens written to cache" is `cache_creation + ephemeral_5m_input_tokens + ephemeral_1h_input_tokens` — which this repo's own integration tests already use.

If anything, the actionable item is a docs gap in `langchain-core`: `InputTokenDetails.cache_creation`'s docstring doesn't mention it's zeroed when the ephemeral per-TTL keys are populated, which is what sent me down this path.

*Disclosure: this investigation and comment were prepared with the help of an AI agent (Claude Code).*

PR #1218, opened 2026-08-19
fix(aws): ignore unsupported cross-provider AI blocks

mdrxy:
Closes #1206

When a thread switches model providers, `AIMessage.content` can retain provider-specific blocks such as OpenAI Responses `function_call` items or Google reasoning. `ChatBedrockConverse` currently tries to serialize those raw blocks and raises before it can send the portable text and tool calls, leaving persisted threads unable to return to Bedrock.

This keeps conversion strict for human, system, and tool content while ignoring unsupported blocks in AI history. Standard tool calls are still reconstructed from `AIMessage.tool_calls`, preserving OpenAI's canonical `call_id` instead of its separate Responses item `id`. If every AI content block is provider-specific, a placeholder keeps the Bedrock assistant message valid.

PR #1219, opened 2026-08-20
fix(aws): Normalize cross-provider assistant content before serialization in ChatBedrockConverse

michaelnchin:
Follow-up to #1218, this covers 2 extra edge cases with processing cross-provider messages:

**1. Signed Gemini thinking blocks**

When `ChatGoogleGenerativeAI` generates thinking blocks from the raw Gemini response, it also includes Gemini's thinking signatures. Unfortunately this happens to be the exact same format as Anthropic's blocks (i.e. `{"type": "thinking", "thinking": "...", "signature": "..."}`), so they get serialized as Anthropic thinking blocks anyway, and Bedrock subsequently throws an error when it doesn't recognize the signature format. 

Fixed this by adding a normalization step for AI messages from other providers, so that they're translated to a standard type and dropped with the other unsupported cross-provider AI blocks by the logic implemented in #1218.

**2. `tool_call` handling**

We don't currently have handling for `tool_call` content type, so these blocks get dropped and sometimes end up orphaning a corresponding `tool_result` block, which Bedrock doesn't allow. 

Fixed by adding another condition to `_messages_to_bedrock` to convert `tool_call` blocks to `toolUse`.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -2256,19 +2256,33 @@ def _messages_to_bedrock(
 ) -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]]]:
     """Handle Bedrock converse and Anthropic style content blocks"""
     for idx, message in enumerate(messages):
+        if not isinstance(message, AIMessage):
+            continue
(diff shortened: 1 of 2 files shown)


michaelnchin (commented):
Integ CI passing: https://github.com/langchain-ai/langchain-aws/actions/runs/32336901289

PR #1225, opened 2026-08-22
fix(aws): Raise error on unsupported guardrail header usage with Mantle

### LINKED ISSUE #1224: Bedrock Mantle chat models silently reject unsupported guardrail headers
We've been running this same shape in production-bound code for a few weeks — a `ChatAnthropic` subclass over `AnthropicBedrockMantle`, SigV4 from a task role — so it's good to see it land properly upstream. Two things we measured against the live endpoint that might save someone else the trip.

**Bedrock Guardrails don't work on this endpoint, and they fail quietly.**

We assumed they would, honestly. #1178 landed guardrail support for `ChatAnthropicBedrock` a few days before this PR opened, and since that's also a non-Converse path where guardrails ride as `X-Amzn-Bedrock-Guardrail*` headers, the obvious guess is that Mantle takes the same headers. It doesn't:

| | result |
|---|---|
| `ApplyGuardrail(source=INPUT)` on the payload | `GUARDRAIL_INTERVENED` |
| Mantle `/anthropic/v1/messages`, no headers | `200`, model answers |
| Mantle, **with** the headers (present in `SignedHeaders`) | `200`, answers identically |
| Converse, same guardrail, same `DRAFT` version | `stopReason=guardrail_intervened` |

The first and last rows are there so the middle one means something — the payload really is blockable, and the guardrail really does work with that exact version, just not here. We tried it both signed and unsigned in case the service was dropping an unsigned header; same result either way.

What makes this worth a comment rather than a shrug is the failure mode. There's no `400`, no warning, and no `amazon-bedrock-guardrailAction` in the body. You get a clean `200` and a normal answer, which is indistinguishable from a guardrail that ran and found nothing. Anyone who reads #1178 and assumes parity here will believe they're covered.

We don't think the library can fix this — the endpoint is the one ignoring the headers — but a line in the class docstring would probably stop a few people from shipping a compliance claim they can't back. Might also be worth flagging to the Mantle service team.

**Small related thing:** the API rejects whitespace-only text blocks too, and the message is different, so a guard written against one will miss the other.

```
{"type": "text", "text": ""}      -> 400  text content blocks must be non-empty
{"type": "text", "text": "   "}   -> 400  text content blocks must contain non-whitespace text
{"type": "text", "text": "\n"}    -> 400  text content blocks must contain non-whitespace text
```

**On the ~10% stream flake** — we see a related shape on this endpoint, where `content` comes back as a mixed list of raw strings and dicts, in 9 of 30 real Sonnet 5 turns. We can't tell from out here whether it's your trigger or a cousin of it. In our case the interesting part is that it's harmless in the turn that produces it and only bites on the *next* one, when the raw `""` gets wrapped into an empty text block and the API rejects the whole request.

If a client-side normalization would be welcome upstream, we have one running and I'm happy to open it as a separate PR. If the library's position is that it shouldn't reshape provider output, that's a perfectly good answer and we'll keep ours local — just say which.

_Originally posted by @AlcebiadesFilho in https://github.com/langchain-ai/langchain-aws/issues/1190#issuecomment-5308242972_

michaelnchin:
Fixes #1224:

Updates `AnthropicBedrockMantle` and `ChatOpenAIMantle` to catch and throw an error on user attempts to pass `X-Amzn-Bedrock-Guardrail*` headers, which are currently unsupported on Mantle endpoint.

Marked TODO removal, as this is expected to be a temporary mitigation until Mantle guardrails support is added.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/anthropic.py
+++ b/libs/aws/langchain_aws/chat_models/anthropic.py
@@ -23,7 +23,11 @@ from typing_extensions import Self
 from langchain_aws._version import _add_langchain_aws_version
 from langchain_aws.chat_models._anthropic_utils import _create_bedrock_client_params
 from langchain_aws.data._profiles import _PROFILES
-from langchain_aws.utils import MODEL_ID_GEO_PREFIXES
+from langchain_aws.utils import (
+    _MANTLE_GUARDRAILS_ERR_MSG,
+    MODEL_ID_GEO_PREFIXES,
+    _check_no_mantle_guardrail_headers,
+)
 
 _MODEL_PROFILES = cast("ModelProfileRegistry", _PROFILES)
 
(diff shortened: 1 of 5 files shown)


michaelnchin (commented):
Integ CI [failing](https://github.com/langchain-ai/langchain-aws/actions/runs/32756902545/job/97543745754), due to unrelated error on Mistral standard tests - fix pending here: https://github.com/langchain-ai/langchain-aws/pull/1228

michaelnchin (commented):
Passing after fix merge: https://github.com/langchain-ai/langchain-aws/actions/runs/32794908648

PR #1227, opened 2026-08-22
fix(aws): preserve redacted reasoning content

### LINKED ISSUE #1223: `ChatBedrockConverse.stream()` raises `IndexError` on models that emit `reasoningContent.redactedContent` deltas (GPT-5.6, Grok 4.6)
## Description

`ChatBedrockConverse.stream()` crashes with `IndexError: list index out of range` when the
model streams redacted reasoning blocks.

Amazon Bedrock's `ConverseStream` can deliver reasoning as either
`reasoningContent.text` (plain) or `reasoningContent.redactedContent` (encrypted bytes).
`_bedrock_to_lc()` handles the first form but has no branch for the second, so it returns an
empty list. `_parse_stream_event()` immediately indexes `[0]` into that result and raises.

xAI Grok 4.6 on Bedrock always returns reasoning in the redacted form, and reasoning is on by
default (`reasoning.effort` defaults to `low`), so every streamed call fails.

This is not a provider quirk. `redactedContent` is a documented member of
[`ReasoningContentBlockDelta`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlockDelta.html),
and that type is a union:

> **Important**
> This data type is a UNION, so only one of the following members can be specified when used or returned.
>
> **redactedContent** — The content in the reasoning that was encrypted by the model provider for
> safety reasons. Type: Base64-encoded binary data object

So a delta carrying only `redactedContent` is exactly what the API contract allows. Any model that
encrypts its reasoning will produce one.

## Environment

| | |
|---|---|
| langchain-aws | 1.7.2 (also reproduced on **1.7.3** and on `main`) |
| langchain-core | 1.6.0 |
| boto3 / botocore | 1.43.75 |
| Python | 3.14.6 |
| Region | `ap-northeast-2` |
| Model | `global.xai.grok-4.6` |

Code references are permalinked to [`0434c7f`](https://github.com/langchain-ai/langchain-aws/commit/0434c7f94f6c27bfa275534f8ff1b0c3f1738144) (`main` as of 2026-08-21), so line numbers stay valid as the file changes.

## Scope: who actually hits this

To be precise about impact:

- **Default configuration does not crash.** `ChatBedrockConverse(...).stream()` with no extra
  arguments falls back to the non-streaming Converse API. It yields exactly one chunk containing
  the whole message, after generation finishes (measured 3x: 6.55s / 9.13s / 15.19s to that single
  chunk). No exception — streaming just silently disappears.
- The crash requires `disable_streaming=False`, which is a public field on
  `BaseChatModel` and is what the library's own warning suggests (quoted at the bottom).
- The model does support streaming at the API layer: the same prompt through `boto3`
  `converse_stream()` yields 160-218 deltas.

So the affected users are those who try to turn streaming on — not everyone.

**Impact is not limited to Grok.** I streamed 14 models available in `ap-northeast-2` and
inspected the delta kinds each one emits:

| Model | reasoning delta | `ChatBedrockConverse` streaming |
|---|---|---|
| `global.openai.gpt-5.6-sol` | `reasoningContent.redactedContent` | **IndexError** |
| `global.xai.grok-4.6` | `reasoningContent.redactedContent` | **IndexError** |
| `global.anthropic.claude-sonnet-5` | `reasoningContent.signature` | works |
| Claude 4.x, Nova family | `text` only | works |

So it is not "reasoning models crash" — it is "models that encrypt their reasoning crash", and
that is already two providers (OpenAI and xAI). Claude Sonnet 5 works precisely because it sends
`signature`, which the branch does handle.

## How this differs from the streaming-allowlist issues

There is a family of closed issues about `_get_streaming_support()` missing an entry for a given
provider (#1204 Moonshot Kimi, #1139 Claude Sonnet 5, #1158 on the mechanism itself, and earlier
#354, #489, #544, #573). Those are about streaming being **silently disabled**.

This report is the opposite situation: streaming is **enabled** and then crashes. Adding `xai` to
the allowlist would not fix it — it would make the crash the default instead of opt-in.

Also distinct from #1218 (merged 2026-08-20), which ignores unsupported blocks when converting
`AIMessage` history **outbound** to Bedrock. This bug is on the **inbound** streaming path.

## Still present on latest

Checked on 2026-08-21:

- `langchain-aws==1.7.3` (current release on PyPI) — reproduces.
- `main` branch — `grep -in redacted libs/aws/langchain_aws/chat_models/bedrock_converse.py`
  returns a single comment line (`# Handle optional content field (may be absent in streaming
  or redacted)`). There is no code path that reads `redactedContent`.

### 1. Without AWS credentials (unit level)

```python
from langchain_aws.chat_models.bedrock_converse import _bedrock_to_lc

# Plain reasoning — handled
print(_bedrock_to_lc([{"reasoningContent": {"text": "thinking"}}]))
# [{'type': 'reasoning_content', 'reasoning_content': {'text': 'thinking'}}]

# Redacted reasoning — returns []
print(_bedrock_to_lc([{"reasoningContent": {"redactedContent": b"abc"}}]))
# []
```

`_parse_stream_event()` then does `_bedrock_to_lc([...])[0]` and raises
([L2573](https://github.com/langchain-ai/langchain-aws/blob/0434c7f94f6c27bfa275534f8ff1b0c3f1738144/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L2573)).

### 2. End to end

```python

llm = ChatBedrockConverse(
    model="global.xai.grok-4.6",
    region_name="ap-northeast-2",
    max_tokens=200,
    disable_streaming=False,   # required: xai is not in the streaming allowlist
)
for chunk in llm.stream("Say hello in two sentences."):
    print(chunk)
```

```
IndexError: list index out of range
  File "langchain_aws/chat_models/bedrock_converse.py", line 2506, in _parse_stream_event
    **_bedrock_to_lc([event["contentBlockDelta"]["delta"]])[0],
```

(Line 2506 is from the installed 1.7.2; on `main` the same statement is
[L2573](https://github.com/langchain-ai/langchain-aws/blob/0434c7f94f6c27bfa275534f8ff1b0c3f1738144/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L2573).)

The raw API streams fine — the same call through `boto3` `converse_stream()` yields 446 deltas,
all of kind `reasoningContent`.

## Root cause

[`_bedrock_to_lc()`](https://github.com/langchain-ai/langchain-aws/blob/0434c7f94f6c27bfa275534f8ff1b0c3f1738144/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L3074) has a `reasoning_content` branch
([L3180-L3210](https://github.com/langchain-ai/langchain-aws/blob/0434c7f94f6c27bfa275534f8ff1b0c3f1738144/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L3180-L3210)) that covers:

- `reasoning_text` (InvokeModel form)
- `text` (streaming form)
- `signature` (streaming form)

There is no branch for `redacted_content`. `grep -in redacted bedrock_converse.py` on `main`
returns a single comment line; the key is never read. The chain has no `else`, so when the delta
carries only `redactedContent`, nothing is appended and the function returns `[]`.

`_parse_stream_event()` assumes a non-empty result:

```python
# https://github.com/langchain-ai/langchain-aws/blob/0434c7f94f6c27bfa275534f8ff1b0c3f1738144/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L2573
elif "contentBlockDelta" in event:
    block = {
        **_bedrock_to_lc([event["contentBlockDelta"]["delta"]])[0],   # <-- IndexError
        "index": event["contentBlockDelta"]["contentBlockIndex"],
    }
```

[L2551](https://github.com/langchain-ai/langchain-aws/blob/0434c7f94f6c27bfa275534f8ff1b0c3f1738144/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L2551) applies the same `[0]` to `contentBlockStart` and carries the same risk.
The other four call sites use the list directly or `extend()` it, so an empty result is harmless
there — which is why only streaming crashes.

## What I'd expect

Either of these would be fine:

1. `_bedrock_to_lc()` emits a `reasoning_content` block carrying the redacted payload, so callers
   can pass it back on later turns (Bedrock supports returning encrypted reasoning for multi-turn
   context).
2. `_parse_stream_event()` guards against an empty result and skips unknown delta kinds instead of
   raising.

(1) preserves information; (2) alone would at least keep unknown future delta types from crashing
the stream. Both together seem safest.

## Workaround

Disable reasoning, which switches the deltas to `text`:

```python
ChatBedrockConverse(
    model="global.xai.grok-4.6",
    region_name="ap-northeast-2",
    disable_streaming=False,
    additional_model_request_fields={"reasoning": {"effort": "none"}},
)
```

| setting | delta kind | `stream()` |
|---|---|---|
| default (reasoning on) | `reasoningContent` | `IndexError` |
| `effort=none` | `text` | works (13 chunks) |

So today, reasoning and streaming are mutually exclusive for this model through
`ChatBedrockConverse`.

## Related note

[`_get_streaming_support()`](https://github.com/langchain-ai/langchain-aws/blob/0434c7f94f6c27bfa275534f8ff1b0c3f1738144/libs/aws/langchain_aws/chat_models/bedrock_converse.py#L864) is an allowlist keyed on provider/model substrings
(`ai21` jamba-1-5, `amazon` nova-*, `anthropic` claude-3+, …). `xai` is absent, so
`disable_streaming` resolves to `True` and users see:

```
Streaming disabled for model 'global.xai.grok-4.6': provider 'xai' is not verified as
streaming-capable with ChatBedrockConverse, so calls will fall back to the non-streaming
Converse API. If this model does support streaming, pass disable_streaming=False to override.
```

The message invites `disable_streaming=False`, which is exactly the path that crashes. Adding
`xai` to the allowlist (the fix pattern used in #1205, #1185) would make sense **after** the delta
handling is fixed — doing it first would turn an opt-in crash into a default one.

jackjin1997:
Fixes #1223

Preserve encrypted Bedrock reasoning deltas as `reasoning_content` blocks so streamed reasoning does not crash and the opaque payload can be replayed in later turns.

---

Bedrock ConverseStream may return `reasoningContent.redactedContent` instead of text or a signature. The inbound converter previously dropped that valid union member, leaving `_parse_stream_event` to index an empty result. This maps the payload into the existing reasoning block shape and restores it to Bedrock format on outbound conversion.

The main area to review is the provider-specific `redacted_content` field retained inside the existing `reasoning_content` block.

<details>
<summary>Test plan</summary>

- Added a failing-then-passing regression for parsing and replaying redacted reasoning.
- Added a streamed-delta regression covering the former `IndexError` path.
- Full `test_bedrock_converse.py`: 370 passed, 1 xpassed.
- Ruff check and format passed; mypy passed across 145 source files.

</details>

AI-assisted investigation and implementation; all changes were reviewed and verified locally.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -2963,7 +2963,15 @@ def _lc_content_to_bedrock(
             reasoning_content = block.get("reasoningContent") or block.get(
                 "reasoning_content", {}
             )
-            if reasoning_content.get("signature", ""):
+            if "redactedContent" in reasoning_content:
+                bedrock_content.append(
+                    {
+                        "reasoningContent": {
+                            "redactedContent": reasoning_content["redactedContent"]
+                        }
+                    }
+                )
+            elif reasoning_content.get("signature", ""):
                 bedrock_content.append(
                     {
                         "reasoningContent": {
@@ -3194,6 +3202,15 @@ def _bedrock_to_lc(content: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
                 )
             # Streaming block format
             else:
+                if "redacted_content" in reasoning_dict:
+                    lc_content.append(
+                        {
+                            "type": "reasoning_content",
+                            "reasoning_content": {
+                                "redacted_content": reasoning_dict["redacted_content"],
+                            },
+                        }
+                    )
                 if "text" in reasoning_dict:
                     lc_content.append(
                         {
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -1694,6 +1694,25 @@ def test__bedrock_to_lc_anthropic_reasoning() -> None:
     assert expected_lc == actual
 
 
+def test__bedrock_to_lc_redacted_reasoning_round_trip() -> None:
+    redacted_content = b"encrypted-reasoning"
+    expected_lc: List[Union[str, Dict[str, Any]]] = [
+        {
+            "type": "reasoning_content",
+            "reasoning_content": {"redacted_content": redacted_content},
+        }
+    ]
+
+    actual = _bedrock_to_lc(
+        [{"reasoningContent": {"redactedContent": redacted_content}}]
+    )
+
+    assert actual == expected_lc
+    assert _lc_content_to_bedrock(cast(List[Union[str, Dict[str, Any]]], actual)) == [
+        {"reasoningContent": {"redactedContent": redacted_content}}
+    ]
+
+
 def test__bedrock_to_lc_nova_reasoning_content() -> None:
     """Test parsing Nova's reasoningContent format."""
     bedrock_content: List[Dict[str, Any]] = [
@@ -5563,6 +5582,29 @@ def test_content_block_delta_tool_call_chunk_args_type() -> None:
     args = chunk.tool_call_chunks[0]["args"]
     assert isinstance(args, str)
     assert args == '{"city": "Paris"}'
+
+
+def test_content_block_delta_redacted_reasoning() -> None:
+    redacted_content = b"encrypted-reasoning"
+    event = {
+        "contentBlockDelta": {
+            "contentBlockIndex": 2,
+            "delta": {
+                "reasoningContent": {"redactedContent": redacted_content},
+            },
+        }
+    }
+
+    chunk = _parse_stream_event(event)
+
+    assert isinstance(chunk, AIMessageChunk)
+    assert chunk.content == [
+        {
+            "type": "reasoning_content",
+            "reasoning_content": {"redacted_content": redacted_content},
+            "index": 2,
+        }
+    ]
 
 
 def test_streaming_tool_use_round_trip() -> None:


jackjin1997:
The redacted reasoning delta and round-trip fix was incorporated into #1229 in commit 231b50b6737e73979dece764d07dcdf94248321e, with co-authorship retained, and #1229 has now merged. Closing this superseded PR.

PR #408, opened 2025-03-16
Prevent sending thinking blocks with no signature with ChatBedrockConverse

tahouse:
Fixes https://github.com/langchain-ai/langchain-aws/issues/401

Claude 3.7 still works because signatures are always returned with those thinking blocks. DeepSeek thinking blocks do not contain signatures and the ValidationError indicates a signature must be included with thinking blocks.

According to Converse API docs:
> "Within the reasoningText field, the text fields describes the reasoning. The signature field is a hash of all the messages in the conversation and is a safeguard against tampering of the reasoning used by the model. You must include the signature and all previous messages in subsequent Converse requests. If any of the messages are changed, the response throws an error."

https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference-call.html

@3coins I posted a separate PR https://github.com/langchain-ai/langchain-aws/pull/407 to make reviewing easier. Happy to combine if preferred.

PR #1229, opened 2026-08-25
fix(aws): parametrize reasoning block construction by model id

ccurme:
Within Converse, different models accept different forms of reasoning blocks. See below for example:
```python
import boto3

model = "openai.gpt-oss-120b-1:0"  # runs
# model = "deepseek.v3.2"          # runs
# model = "us.deepseek.r1-v1:0"    # ValidationException (reasoning content not allowed)
# model = "us.anthropic.claude-sonnet-4-5-20250929-v1:0"  # ValidationException (signature required)

payload =  {
"modelId": f"{model}",
"messages": [
    {"role": "user", "content": [{"text": "What is 2+2?"}]},
    {"role": "assistant", "content": [
        {"reasoningContent": {"reasoningText": {
            "text": "The user is asking for the sum of 2 and 2. 2+2=4."
        }}},
        {"text": "2 + 2 = 4."},
    ]},
    {"role": "user", "content": [{"text": "Now multiply that by 10."}]},
],
"system": [],
"inferenceConfig": {},
}

client = boto3.client("bedrock-runtime", region_name="us-east-1")
print(client.converse(**payload)["output"]["message"]["content"])
```

https://github.com/langchain-ai/langchain-aws/pull/408 gated reasoning blocks on the presence of a signature. This is correct for Anthropic models, and at the time also handled DeepSeek R1, which (a) doesn't return signatures, and (b) doesn't accept reasoning blocks passed back in. Now, there are models like DeepSeek 3.2 and GPT-OSS that do accept signature-less reasoning blocks.

Here we make the reasoning block construction depend on the model ID: we let unsigned reasoning be passed back to the model on subsequent turns via an allow-list.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -2788,6 +2788,9 @@ def _bedrock_reasoning_block(
     )
 
     model_id_lower = (model_id or "").lower()
+    # TODO: `_get_base_model()` returns the raw ARN when `model_id` is an ARN and
+    # `base_model_id` is unset, so this misses Nova v1. Strip the ARN there, then
+    # simplify this to use the resolved provider.
     provider = model_id_lower.partition(".")[0]
 
     if any(
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -2770,7 +2770,7 @@ def _empty_content_fallback(
 
 
 def _bedrock_reasoning_block(
-    text: str, signature: str, model_id: Optional[str]
+    reasoning: Dict[str, Any], model_id: Optional[str]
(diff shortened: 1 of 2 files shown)


jackjin1997:
One interaction to account for before this lands: a `reasoning_content` block containing only `redactedContent` / `redacted_content` has neither text nor signature, so the new `_bedrock_reasoning_block` path drops it for every model. #1227 fixes the streaming conversion and round-trip for exactly that shape (issue #1223), with focused regressions passing locally. Could we preserve redacted blocks before delegating to this helper, or sequence/rebase #1227 around this change? Otherwise the GPT-5.6/Grok redacted-content path remains unroundtrippable.

ccurme:
Hi @jackjin1997, thanks for flagging this. I've incorporated your fix into 231b50b6737e73979dece764d07dcdf94248321e and given you co-authorship.

michaelnchin (approved):
LGTM, thanks Chester!

PR #1230, opened 2026-08-25
fix(aws): align ChatAnthropicMantle auth precedence with the Anthropic SDK Mantle client

gouveags:
Partially addresses #1217.

## Summary

`ChatAnthropicMantle` currently promotes a Bedrock API key inherited from
`AWS_BEARER_TOKEN_BEDROCK` into the Anthropic SDK's explicit `api_key`
constructor argument. Because explicit API keys have the SDK's highest auth
precedence, that ambient value silently overrides SigV4 credentials supplied by
the caller.

This change omits an environment-derived API key when the caller explicitly
selects a named AWS profile or supplies either half of a complete effective AWS
access-key pair. An explicitly supplied `bedrock_api_key` still wins, and the
existing automatic behavior remains unchanged when the caller provides no
explicit SigV4 signal.

Thanks to @kimnamu for the detailed reproduction and proposed direction in
#1217.

## Changes

**`libs/aws/langchain_aws/chat_models/anthropic.py`**

- Distinguish caller-supplied authentication fields from environment defaults
  with Pydantic's `model_fields_set`.
- Prefer a caller-selected profile or complete effective access-key pair over
  an ambient Bedrock bearer token.
- Preserve explicit `bedrock_api_key` precedence and document the resulting
  authentication order.

**`libs/aws/tests/unit_tests/chat_models/test_anthropic_mantle.py`**

- Cover profiles, fully explicit key pairs, and both mixed explicit/environment
  key-pair directions.
- Verify the shared behavior reaches both synchronous and asynchronous SDK
  constructors.
- Protect explicit bearer-key precedence and the existing ambient-bearer
  default.

## Areas for careful review

- `model_fields_set` is intentionally used to distinguish caller intent from
  values populated by environment-backed default factories.
- This is the narrow precedence fix supported by the current public Anthropic
  SDK constructor. It does not add `auth_mode` or force refreshable default-chain
  SigV4 when an ambient bearer token exists; that broader part of #1217 still
  requires an SDK authentication-selection seam.

## Verification

- `make format` — clean; 147 files unchanged
- `make lint` — Ruff check/format and mypy passed; 145 source files checked
- `make test` — 1,211 passed, 3 skipped, 1 xpassed
- `git diff --check` — passed

The regression tests are hermetic and do not make network calls. A live Mantle
request was not run because no Mantle-enabled AWS credentials were available in
this environment.

## Release note

`ChatAnthropicMantle` now prefers caller-supplied AWS profile or access-key
[...]

PR #1234, opened 2026-08-26
feat(aws): Add auth_mode parameter to ChatAnthropicMantle

### LINKED ISSUE #1217: Add explicit authentication mode selection to ChatAnthropicMantle
### Package

langchain-aws

### Checked other resources

### Feature Description

ChatAnthropicMantle supports both Bedrock API-key authentication and AWS SigV4 authentication. However, authentication is currently selected implicitly.

When AWS_BEARER_TOKEN_BEDROCK or another API-key source is present in the process environment, the underlying AnthropicBedrockMantle client may select bearer authentication even when the application intends to use AWS SigV4 through a task role or another AWS credential provider.

Please add an explicit authentication-mode option while preserving the existing automatic behavior by default. For example:

auth_mode: Literal["auto", "sigv4", "api_key"] = "auto"

Expected behavior:

- auto: preserve the current credential-selection behavior.
- sigv4: always use the AWS credential chain and ignore ambient Bedrock API-key variables for this client.
- api_key: use Bedrock API-key authentication and fail clearly when no key is available.

The exact name and shape of the option can follow maintainer preference.

### Use Case

We run LangGraph workloads on AWS using task-role credentials. Some shared worker and container environments may also contain Bedrock API-key variables for unrelated clients.

In that situation, ChatAnthropicMantle can unexpectedly use bearer authentication instead of the workload's task-role SigV4 credentials. Applications currently have to remove or mutate process-wide environment variables, isolate clients in separate processes, or manipulate private SDK state.

An explicit authentication mode would make credential ownership deterministic and safe for shared and multi-tenant runtimes.

```python
from langchain_aws import ChatAnthropicMantle

model = ChatAnthropicMantle(
    model="anthropic.claude-sonnet-5",
    region_name="us-east-1",
    auth_mode="sigv4",
)

Implementation requirements:

- Preserve current behavior when auth_mode is omitted or set to auto.
- Cover synchronous and asynchronous client construction.
- Test SigV4 selection when AWS_BEARER_TOKEN_BEDROCK is present.
- Avoid mutating process-wide environment variables.
- Avoid depending on private Anthropic SDK state.

If the Anthropic SDK does not yet expose the required public authentication-selection seam, the implementation may need a coordinated upstream SDK change.
```

### Additional Context

Related merged PR:
https://github.com/langchain-ai/langchain-aws/pull/1190

Maintainer confirmation of the same authentication-selection problem:
https://github.com/langchain-ai/langchain-aws/pull/1190#issuecomment-5328398825

Original runtime report:
https://github.com/langchain-ai/langchain-aws/pull/1190#issuecomment-5276133775

Related but distinct environment-mutation issue:
https://github.com/langchain-ai/langchain-aws/issues/989

We have exercised the SigV4-forced path with task-role credentials, including streaming, tool calls, and prompt caching, and would be happy to contribute the implementation after maintainer approval and assignment.

michaelnchin:
Adds an optional `auth_mode` argument to ChatAnthropicMantle, allowing users to explicitly choose either `api_key` or `sigv4` auth sources. The default `auto` mode will continue to use the Anthropic SDK Mantle client's auth source priority order.

Apart from the extra user control, two main motivations for adding `auth_mode`:
1. The precedence fix in #1230 covers all cases where the user intent can be inferred from explicit credential or key values.  However, #1217 specifically references an example where the SigV4 credentials are pulled from the provider chain. The SDK Mantle client has the provider chain as the lowest priority of any source, so these would still be overridden by a Bedrock API key. In this scenario, `"auth_mode"="sigv4"` would allow users to directly select the provider chain as the auth source.
2. `ChatAnthropicMantle`'s auth source priority order is inconsistent with that of ChatBedrockConverse (and
most of our other boto3 based integrations), as the Converse chat model always prefers the `AWS_BEARER_TOKEN_BEDROCK` env variable if ti's present. The ``auth_mode="api_key"`` can be used to generally restore the key-over-credentials precedence that users might be more accustomed to.


libs/aws/langchain_aws/chat_models/anthropic.py
```diff
@@ -509,6 +526,29 @@ def _get_request_payload(
         _check_no_mantle_guardrail_headers(kwargs.get("extra_headers"))
         return super()._get_request_payload(input_, stop=stop, **kwargs)
 
+    @model_validator(mode="after")
+    def _validate_auth_mode(self) -> Self:
+        if self.auth_mode == "api_key" and not self.bedrock_api_key:
```
gouveags, line 534: The Mantle SDK also treats `ANTHROPIC_AWS_API_KEY` as a valid API-key source, but this rejects `auth_mode="api_key"` when that is the only key available. Could we preserve that supported source and add a regression test?

libs/aws/langchain_aws/chat_models/anthropic.py
```diff
@@ -533,8 +573,14 @@ def _client_params(self) -> dict[str, Any]:
             and self.aws_access_key_id is not None
             and self.aws_secret_access_key is not None
         )
-        if self.bedrock_api_key and (
-            "bedrock_api_key" in self.model_fields_set or not explicit_sigv4_credentials
+        if (
+            self.bedrock_api_key
+            and self.auth_mode != "sigv4"
+            and (
+                self.auth_mode == "api_key"
```
gouveags, line 583: `api_key` mode still forwards AWS credential fields below. With a valid bearer key plus only `AWS_ACCESS_KEY_ID` (or only `AWS_SECRET_ACCESS_KEY`) in the environment, the SDK raises its incomplete-pair error before selecting bearer auth. Could we omit SigV4 credential kwargs in this mode and cover both partial-pair cases?

libs/aws/langchain_aws/chat_models/anthropic.py
```diff
@@ -556,15 +602,23 @@ def _client_params(self) -> dict[str, Any]:
             client_params["timeout"] = self.default_request_timeout
         return client_params
 
+    def _pin_client_auth_mode(self, client: Any) -> Any:
+        if self.auth_mode == "sigv4" and not client._use_sigv4:
```
gouveags: This depends on and mutates the SDK-private `_use_sigv4` state, which #1217 explicitly asks us to avoid; the dependency is also unbounded above. Could we coordinate a public SDK auth selector and pass it at construction instead of patching client state afterward?

libs/aws/tests/unit_tests/chat_models/test_anthropic_mantle.py
```diff
@@ -406,3 +406,127 @@ def test_env_sigv4_credentials_do_not_outrank_ambient_api_key() -> None:
...
+        m.setenv("AWS_BEARER_TOKEN_BEDROCK", "api-key")
+        model = ChatAnthropicMantle(  # type: ignore[call-arg]
+            model_name=MODEL_NAME,
+            region_name="us-east-1",
+            auth_mode="sigv4",
+        )
+        # the env-derived key must not reach the SDK client
+        assert "api_key" not in model._client_params
+        client = model._client
+        assert client._use_sigv4 is True
```
gouveags, line 500: These checks assert the same private attributes the implementation mutates, so they do not independently prove that the request is SigV4-signed. Could we use a mocked transport/credential provider and assert the outgoing `Authorization` scheme for both sync and async clients?

gouveags (commented):
Noticed some problems, let me know if you have any questions.

libs/aws/langchain_aws/chat_models/anthropic.py
michaelnchin: Nice catch on the SDK specific field, thanks. Added an extra check

libs/aws/langchain_aws/chat_models/anthropic.py
michaelnchin: Thanks - gated passing cred fields behind non API key `auth_mode`

libs/aws/langchain_aws/chat_models/anthropic.py
michaelnchin: I wish we could set the auth mode directly during `AnthropicBedrockMantle` construction, but the only config it exposes is a nuclear `skip_auth`.

Regarding the task-role -> explicit creds workaround you mentioned at the end of #1217, the primary benefit of task-role credentials is that they refresh automatically. If we convert to string and pass them in directly, the client requests would start failing once the role's session creds auto-rotate. Keeping the auth resolution inside the SDK with `_use_sigv4` is the really the only option (for now) to allow task-role creds to work properly on `ChatAnthropicMantle`.

Valid concern that `AnthropicBedrockMantle` could lose `_use_sigv4` in a later version though, so I'll add an attribute check + user error advising downgrade if it's no longer present.

libs/aws/tests/unit_tests/chat_models/test_anthropic_mantle.py
michaelnchin: Sure! Added an extra test.

libs/aws/langchain_aws/chat_models/anthropic.py
gouveags: Sounds good

libs/aws/langchain_aws/chat_models/anthropic.py
```diff
@@ -509,6 +526,33 @@ def _get_request_payload(
         _check_no_mantle_guardrail_headers(kwargs.get("extra_headers"))
         return super()._get_request_payload(input_, stop=stop, **kwargs)
 
+    @model_validator(mode="after")
+    def _validate_auth_mode(self) -> Self:
+        if (
+            self.auth_mode == "api_key"
+            and not self.bedrock_api_key
+            and not os.getenv("ANTHROPIC_AWS_API_KEY")
```
gouveags: Thanks for adding this check. There is still a construction/use gap: validation reads `ANTHROPIC_AWS_API_KEY` now, while the lazy SDK client reads it later. I reproduced key A being replaced by key B, and deleting the variable makes declared `api_key` mode construct a SigV4 client. This is easy to avoid with the normal LangChain pattern: make `bedrock_api_key` use `secret_from_env(("AWS_BEARER_TOKEN_BEDROCK", "ANTHROPIC_AWS_API_KEY"), default=None)` and validate only the field. That snapshots either alias at model construction; could we add a mutation/removal regression test?

libs/aws/tests/unit_tests/chat_models/test_anthropic_mantle.py
```diff
@@ -406,3 +406,198 @@ def test_env_sigv4_credentials_do_not_outrank_ambient_api_key() -> None:
...
+)
+def test_auth_mode_authorization_scheme_on_the_wire(
+    auth_mode: Literal["api_key", "sigv4"], expected_scheme: str
+) -> None:
+    import boto3.session
+    import httpx
+
+    with MonkeyPatch().context() as m:
+        m.setenv("AWS_BEARER_TOKEN_BEDROCK", "api-key")
+        m.setenv("AWS_ACCESS_KEY_ID", "key-id")
```
gouveags: This test still bypasses the pin: these env vars populate the model’s explicit AWS fields, so the SDK selects SigV4 before `_pin_client_auth_mode`. I verified the test remains green if the pin is removed. A default-chain test is feasible: clear static AWS env vars, point `AWS_SHARED_CREDENTIALS_FILE` at a temporary `[default]` profile, keep `AWS_BEARER_TOKEN_BEDROCK`, assert `_client_params` contains no API-key or AWS credential kwargs, then assert `AWS4-HMAC-SHA256` on sync and async requests. That test passes here and fails with `Bearer ambient-key` when the pin is removed.

gouveags (commented):
Two follow-up gaps remain after the latest changes; details inline.

michaelnchin (commented):
New ChatAnthropicMantle auth integ tests are passing in the CI: https://github.com/langchain-ai/langchain-aws/actions/runs/33232118375/job/99046629901 (1 unrelated flake on xAI standard tests)

PR #1243, opened 2026-08-28
fix(aws): Repair model-stringified tool-call args in structured output

### LINKED ISSUE #1221: `ChatBedrock.with_structured_output()`: a `List[Model]` field is intermittently re-serialized as a JSON string instead of the declared array (Claude Sonnet on Bedrock, forced tool-calling)
### Package

langchain-aws

### Checked other resources

### Example Code

```python
class Output(BaseModel):
    schema_version: str
    items: List[Entry]

2 validation errors for Output
schema_version
  Field required [type=missing, input_value={'items': {'schema_version': 'v1', 'Onboarding Timeline': [1001, 1002, 1003], ...}}, input_type=dict]
items.schema_version
  Input should be a valid list [type=list_type, input_value='v1', input_type=str]
```

```shell

```

### System Info

```
langchain-aws==1.7.0
langchain-core==1.5.3
pydantic==2.11.7
boto3==1.43.66
```

Platform: Linux, Python 3.12
Model: `eu.anthropic.claude-sonnet-5` (Bedrock cross-region inference profile), region `eu-central-1`

### Reproduction

```python
from pydantic import BaseModel, Field
from typing import List

class Entry(BaseModel):
    label: str
    values: List[int] = Field(default_factory=list)

class Output(BaseModel):
    items: List[Entry]

llm = ChatBedrock(model_id="eu.anthropic.claude-sonnet-5", region_name="eu-central-1")

out = llm.with_structured_output(Output, include_raw=True, tool_choice="any").invoke(
    "Group these numbered discussion points into 5-8 labeled buckets by topic, "
    "covering points 1 through 40: <a long numbered list of short text items>"
)

print(out["parsed"])         # None
print(out["parsing_error"])  # ValidationError below
print(out["raw"].tool_calls) # the actual malformed args
```

Actual `parsing_error` / tool call args on a failing run (production example, field values are illustrative, not verbatim):

```
1 validation error for Output
items
  Input should be a valid list [type=list_type, input_value='[{"label": "Onboarding Timeline", "values": [1001, 1002, 1003]}, {"label": "Compensation Discussion", "values": [1004, 1005]}, {"label": "Next Steps||ACTION_ITEMS||", "values": []}]', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/list_type

attempted args: [{'name': 'Output', 'args': {'items': '[{"label": "Onboarding Timeline", "values": [1001, 1002, 1003]}, {"label": "Compensation Discussion", "values": [1004, 1005]}, {"label": "Next Steps||ACTION_ITEMS||", "values": []}]'}, 'id': 'toolu_bdrk_01XXXXXXXXXXXXXXXXXXXXXX', 'type': 'tool_call'}]
```

The declared schema requires `items` to be a JSON array. The model instead emits the fully-correct array content, but as a **string** — sometimes bare (as above), sometimes wrapped one level deeper under its own key (`'{"items": [...]}'`).

A second shape, seen after adding a second top-level field specifically to test whether increasing field count avoids the above:

```python
class Output(BaseModel):
    schema_version: str
    items: List[Entry]
```

```
2 validation errors for Output
schema_version
  Field required [type=missing, input_value={'items': {'schema_version': 'v1', 'Onboarding Timeline': [1001, 1002, 1003], ...}}, input_type=dict]
items.schema_version
  Input should be a valid list [type=list_type, input_value='v1', input_type=str]
```
Here the model nests the *sibling* field (`schema_version`) inside `items` instead of keeping it a top-level sibling — same collapsing tendency, different shape.

### Expected behavior

`out["parsed"]` should be a valid `Output` instance whenever the model's tool-call arguments contain the correct data (which they do — the content itself is accurate, only mis-encoded). At minimum, `with_structured_output()` should not silently accept a malformed tool call from Bedrock that doesn't match the declared JSON schema for that field's type.

### Additional context

Aggregated from our own call-log observability (30-day window, several independent Pydantic schemas sharing this shape — a `List[Model]` field holding a variable number of multi-field items — in production):

- 528 failures matching the stringification shapes above, **100% on `eu.anthropic.claude-sonnet-5`** via this exact `ChatBedrock` code path; 0 observed on other models routed through the same call site in the same deployment.
- Failure rate tracks **per-item payload size/complexity inside the list**, not field count or list-vs-dict: a `Dict[str, str]` field with a handful of short entries almost never fails (<0.01%); a `List[Model]` of many items each carrying a large nested sub-list or long free text fails at 3-10%+ of calls; the identical schema used for a *shrinking* transform (merging/deduplicating an already-generated list) fails far less than one *generating* a fresh list from raw input (0.02% vs 3.9%, same schema).
- Ruled out on our end: not a double-serialization bug in our own code (we're on the non-streaming `.invoke()` path, not `astream()`, which has a related but distinct tracked issue: #827). Not fixed by adding a field to break up a "single dominant field" schema — see the second shape above, found while testing that exact mitigation.
- The only reliable client-side mitigation we've found is a `model_validator(mode="before")` that unwraps a stringified/self-wrapped field before Pydantic validates — it works, but doesn't reduce how often the model produces the malformed shape in the first place, and doesn't stop it from moving elsewhere (see the sibling-field case). Bedrock's native grammar-constrained structured output (`method="json_schema"`, GA Feb 2026) would presumably prevent this at the decoding level, but isn't currently reachable through `ChatBedrock.with_structured_output()` — is there a roadmap for it to prefer that path for eligible models/schemas (related to #883), falling back to forced tool-calling otherwise?

michaelnchin:
Fixes #1221 

Adding handling for a reported case where models can return list/object type tool call arguments as a JSON string instead of the format actually specified via `with_structured_output`. 

While only reported for `ChatBedrock` / `InvokeModel` API, this ostensibly applies to `ChatBedrockConverse` as well, so added a shared `_repair_stringified_tool_call_message` mitigation helper that will apply to the structured output logic in both chat models.


michaelnchin (commented):
Integ tests passing: https://github.com/langchain-ai/langchain-aws/actions/runs/33713391186

PR #1244, opened 2026-08-28
fix(aws): fix json_schema structured output via ChatBedrock Converse passthrough

michaelnchin:
Addresses: https://github.com/langchain-ai/langchain-aws/issues/1221#issuecomment-5447158409

Adding `strict` as a parameter `ChatBedrockConverse`'s `with_structured_output` method to prevent failures when the Converse fallback in `ChatBedrock` same method try to pass that field.


michaelnchin (commented):
Integ CI passing (ignoring 2 unrelated flake xAI standard tests): https://github.com/langchain-ai/langchain-aws/actions/runs/33568871366

PR #1249, opened 2026-09-01
feat(aws): Add ChatOpenAIBedrock for bedrock-runtime OpenAI endpoint

rishabhagrawal1:
## Summary

Adds `ChatOpenAIBedrock`, an OpenAI-compatible chat model for the **bedrock-runtime** endpoint (`bedrock-runtime.{region}.amazonaws.com/openai/v1`), alongside the existing `ChatOpenAIMantle` (bedrock-mantle). This is the [recommended endpoint](https://docs.aws.amazon.com/bedrock/latest/userguide/endpoints.html) for OpenAI models (e.g. the GPT-5.x family) on Amazon Bedrock, and supports the Chat Completions and Responses APIs plus cross-Region inference.

## Why

The OpenAI-compatible surface is now available on `bedrock-runtime`, along with `bedrock-mantle`. Shared logic has been factored into a private `_BaseChatBedrockOpenAI` base; each public class sets only endpoint-specific attributes. `ChatOpenAIMantle`'s public behavior is unchanged.

## What's in it

- **`ChatOpenAIBedrock`** — runtime base URL, `openai-bedrock` tracing provider. Cross-Region inference is selected via the `model` id (e.g. `us.openai.gpt-5.6-sol`, `global.openai.gpt-5.6-sol`).
- **Shared base refactor** — region/base-URL resolution, Bedrock API-key auth (static bearer token or short-term keys derived from AWS credentials, following #1203), profile resolution, LangSmith tracing, and Chat Completions/Responses routing now live once in `_BaseChatBedrockOpenAI`.
- **Deduplication into `langchain_aws.utils`** — the cross-Region prefix strip and the guardrail-rejection guard (previously duplicated per #1225) are now shared helpers; `ChatBedrockConverse` and `ChatAnthropicMantle` are routed through them. Behavior unchanged.

## Behavior notes

- **Guardrails** are rejected on the OpenAI-compatible path for both endpoints (bedrock-mantle silently ignores them; no OpenAI model on bedrock-runtime accepts them yet), consistent with #1225. Use `ChatBedrockConverse` / `ChatAnthropicBedrock` for guardrails.
- **Tool calling with GPT-5.x reasoning models** is rejected on the Chat Completions path while reasoning is active (a model-level behavior, same on both endpoints). Bind tools via the Responses API (`use_responses_api=True`) or `reasoning_effort="none"`. Open-weight models (gpt-oss) are unaffected. Documented on both classes.

## Areas that need careful review

- The `_BaseChatBedrockOpenAI` extraction restructures `ChatOpenAIMantle` (same file) — its public surface and behavior are unchanged; the full pre-existing Mantle unit suite still passes.
- The dedup deliberately edits `anthropic.py` (`ChatAnthropicMantle`) and `bedrock_converse.py` (`_get_base_model`) so they consume the shared `utils` helpers. That's why an "add ChatOpenAIBedrock" PR touches those files. Behavior is preserved.

## Testing

- `make lint` / `make test` green (**1290 passed**); new-class (`openai.py`) coverage 94.64%.
- New unit suite (`ChatModelUnitTests` + endpoint/auth/guardrail/stream-routing/base-guard cases) and integration suite (standard suite via the Responses API + credential-derived auth + global cross-Region + Responses, skipped without creds).
- Verified live against a Bedrock account (public `langchain_aws` imports, credential-derived auth): chat, Responses API, streaming, `us.`/`global.` GPT-5.6 profiles, guardrail rejection, and a **LangGraph `create_react_agent` tool-call loop** (via the Responses API). `ChatOpenAIMantle` and `ChatBedrockConverse` re-verified live after the refactor.

### DIFF
--- a/libs/aws/langchain_aws/__init__.py
+++ b/libs/aws/langchain_aws/__init__.py
@@ -18,6 +18,7 @@ if TYPE_CHECKING:
         ChatAnthropicBedrock,
         ChatAnthropicMantle,
         ChatBedrockNovaSonic,
+        ChatOpenAIBedrock,
         ChatOpenAIMantle,
     )
     from langchain_aws.document_compressors.rerank import BedrockRerank
@@ -94,6 +95,7 @@ __all__ = [
     "ChatBedrock",
     "ChatBedrockConverse",
     "ChatBedrockNovaSonic",
+    "ChatOpenAIBedrock",
     "ChatOpenAIMantle",
     "SagemakerEndpoint",
     "AmazonKendraRetriever",
@@ -136,6 +138,10 @@ def __getattr__(name: str) -> Any:
         "ChatBedrockNovaSonic": (
             "langchain_aws.chat_models",
             'pip install "langchain-aws[nova-sonic]"',
+        ),
+        "ChatOpenAIBedrock": (
+            "langchain_aws.chat_models",
+            'pip install "langchain-aws[openai]"',
         ),
         "ChatOpenAIMantle": (
             "langchain_aws.chat_models",
--- a/libs/aws/langchain_aws/chat_models/__init__.py
+++ b/libs/aws/langchain_aws/chat_models/__init__.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
         ChatAnthropicMantle,
     )
     from langchain_aws.chat_models.bedrock_nova_sonic import ChatBedrockNovaSonic
-    from langchain_aws.chat_models.openai import ChatOpenAIMantle
+    from langchain_aws.chat_models.openai import ChatOpenAIBedrock, ChatOpenAIMantle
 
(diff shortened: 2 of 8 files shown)


3coins (changes requested):
@rishabhagrawal1 
Code looks good. After rebasing this branch onto current main, `test_profile_resolved_through_cross_region_prefix` fails. Upstream now registers **us.openai.gpt-5.6-sol** as its own profile (distinct name, structured_output=False, no reasoning_effort_levels), so `_get_default_model_profile` returns it via exact match instead of stripping the prefix to the base and the == base.profile assertion no longer holds.

I would suggest either asserting non-empty resolution for a prefixed id (drop the equality to base), or using a model id whose geo variant genuinely isn't registered so the strip path is actually exercised.

PR #1256, opened 2026-09-02
feat(aws): native async for ChatBedrockConverse, ChatBedrock and BedrockEmbeddings

### LINKED ISSUE #663: Async client aioboto3 support on the ChatBedrockConverse and BedrockEmbeddings

roihezkiRaven:
Closes #663.

## Why

`ainvoke`/`astream`/`aembed_*` bridge to the synchronous `boto3` client with `run_in_executor`. The thread is held for the whole model call, so concurrency is capped at the default executor size, `min(32, os.cpu_count() + 4)` — the ceiling @kimnamu measured in #663, and the cause of the latency growth @norman-luengo reported.

Against a local fake `bedrock-runtime` with 500 ms of injected latency, on a 12-core box (16-thread default executor):

| concurrent `ainvoke` | today | this PR | max in flight, before → after |
|---|---|---|---|
| 8 | 0.52 s | 0.61 s | 8 → 8 |
| 32 | 1.12 s | 0.55 s | 16 → 32 |
| 64 | 2.16 s | 0.56 s | 16 → 64 |
| 128 | 4.21 s | 0.64 s | 16 → 128 |

Below the ceiling there is no gain, and slightly more overhead; above it the executor serializes into waves while the native path stays flat. A fake endpoint is used deliberately: against real Bedrock the wall clock is dominated by per-account service quotas, so it measures the account rather than the client.

## Try it

```python
import asyncio

async def main():
    async with ChatBedrockConverse(
        model="us.amazon.nova-micro-v1:0", region_name="us-east-1",
        use_async_transport=True, max_tokens=8,
    ) as model:
        print(await model.ainvoke("hi"))
        async for chunk in model.astream("count to three"):
            print(chunk.text, end="", flush=True)
        await asyncio.gather(*(model.ainvoke("hi") for _ in range(32)))

asyncio.run(main())
```

Integration tests covering the same ground live in `tests/integration_tests/test_async_client.py`; they use Nova Micro and Titan v2 with one-token prompts, so a full run costs a fraction of a cent.

## Why not aiobotocore

The issue asks for `aiobotocore`/`aioboto3` support, but neither can be a dependency of this package:

- `aiobotocore` 3.9.0 (latest) pins `botocore>=1.43.3,<1.43.57`; this package requires `boto3>=1.43.64`. `pip install langchain-aws aiobotocore` has no solution today.
- That is structural, not transient. `aiobotocore`'s ceiling trails botocore by design, while this package's floor moves forward for new AWS APIs (`1.42.42 → 1.43.32 → 1.43.64`, the last in #1200 for DynamoDB vector indexes). The ranges cross and stay crossed.
- `aioboto3` is further behind: its latest release pins `aiobotocore==2.25.1`, i.e. `botocore<1.40.62`.

So `BedrockAsyncClient` signs with `botocore` and sends with `httpx` — the approach [Anthropic's SDK](https://github.com/anthropics/anthropic-sdk-python/blob/main/src/anthropic/lib/bedrock/_client.py) takes, and the one `ChatAnthropicBedrock` already benefits from for Anthropic models only. This covers the rest of the surface (Nova, Titan, Cohere, Llama, embeddings). Event stream framing is parsed by botocore's own `EventStreamBuffer` rather than reimplemented.

**No new dependencies are declared** — `httpx` is already in the tree via `langchain-core`. `async_client` is duck-typed on the `boto3` method names, so anyone who prefers `aiobotocore` can pass their own entered client and manage its `botocore` pin themselves.

## Reading it

Five commits, in dependency order:

1. **`feat: add a native async Bedrock runtime client`** — the new `async_client.py` and `_async_transport.py`, plus a factory in `utils.py`. Purely additive (+919/−3); nothing else in the package changes behavior.
2. **`feat: ... ChatBedrockConverse`** — the bulk of the diff is an extraction: `_generate` and `_stream` duplicated ~50 lines of request building verbatim, now shared. `_agenerate`/`_astream` reuse it.
3. **`feat: ... BedrockEmbeddings`** — same shape: `_embedding_func` split into request building and response parsing so both paths share the per-provider logic.
4. **`feat: ... ChatBedrock and BedrockBase`** — same again for the Invoke API.
5. **`test: cover the native async Bedrock paths`** — 52 unit tests and 9 integration tests.

Commits 2–4 are refactor-then-reuse. The refactors are meant to be behavior-preserving, and the check for that is mechanical: **all 1307 pre-existing unit tests pass untouched**, and each model has a test asserting the sync and async paths send byte-identical requests, so the two cannot drift apart.

Opt-in and additive throughout: with neither `use_async_transport` nor `async_client` set, every async method keeps its current executor behavior.

## Decisions worth a second opinion

- **Retries.** The client retries transient errors with jittered backoff. Attempt counting follows botocore exactly (`max_attempts` counts retries after the initial request, `total_max_attempts` counts every send and wins when both are set), so `max_retries` means the same thing on both transports — thanks to @kimnamu for catching that it did not. This is still not botocore's adaptive mode and shares no retry quota; `retries["mode"]` is ignored and documented as such. I included retries because boto3 retries by default, so shipping none would silently regress the async path against the sync one. Happy to drop them if you would rather retries stayed with the caller.
- **Cross-loop reuse raises.** An `httpx` pool binds to the loop that first drives it. Sharing one client across loops previously hung forever with nothing raised (reproduced: 25 threads each running `asyncio.run` on one shared model — 14 of 75 calls completed, the rest deadlocked in `selectors.select`). It now raises immediately with an actionable message. Louder than before, and deliberate, but it is a behavior choice.
- **Bearer tokens are refused, not ignored.** The built client signs SigV4 only, so `use_async_transport=True` together with a `bedrock_api_key` raises rather than silently authenticating as a different identity. Same approach as #1225.

## Known deviations from boto3

- Bearer-token authentication is unsupported on the async path (see above).
- Responses are the raw JSON, so members botocore would drop as unmodeled may be present — `usage` is a superset of the sync path's.
- `ReasoningContentBlock.redactedContent` stays a base64 `str` rather than `bytes`. It is passed through opaquely so it round-trips, but it differs from the sync path.
- Header-bound Invoke members are transcribed from the service model into `_INVOKE_HEADER_MEMBERS`. @kimnamu verified the mapping against botocore 1.43.75, and a test now asserts it against the live service model so a `uv.lock` bump that adds a header fails CI instead of quietly disabling guardrails.

## Testing done

- 1418 unit tests, `--disable-socket`; `make lint` clean, and `mypy` clean under each of 3.10 / 3.11 / 3.12 / 3.13.
- 9 integration tests against real Bedrock, all passing, including the native `astream` round-trip that confirms real Converse streams deliver `messageStop` through this transport.
- Endpoint resolution is asserted against botocore for FIPS, dual-stack, `AWS_ENDPOINT_URL_BEDROCK_RUNTIME` and the `cn` partition; every one of those assertions was red before the fix.
- Verified against botocore's own service model that Converse/ConverseStream have no header-bound members while the two Invoke operations have eight, and that `InvokeModelWithResponseStream` moves `accept` to `X-Amzn-Bedrock-Accept`.
- Rebased onto `main` past #1287; `_astream` enforces the same stream-completion contract as `_stream`, with the three enforcement tests verified red without it.

---

*Disclaimer: this contribution was developed with the assistance of AI agents (Claude), including adversarial review passes that found several defects in my own first draft (silently dropped guardrail headers, swallowed event-stream error frames, and a cross-loop deadlock). The analysis, benchmarks and code were produced under my direction, and I have reviewed the result.*


roihezkiRaven:
Hi @michaelnchin, is this something that we can push somehow? Would love to get this out with your help :) 
True async is extremely important at scale; we see this as a blocker for multiple high-throughput agents we run.

kimnamu (commented):
Thank you for turning the executor-ceiling numbers from #663 into a real transport, and for the byte-identical sync/async request tests: those made this review mostly a matter of running things. I reviewed the auth, region and serialization surface (`async_client.py`, `_async_transport.py`, the `utils.py` factory), not the model refactors.

**Verified as-is** (botocore 1.43.75 from `uv.lock`, PR head `4f84ed2`):

- `_INVOKE_HEADER_MEMBERS` and the stream variant match the `bedrock-runtime` service model exactly: same 8 members, same header names, `X-Amzn-Bedrock-Accept` and `X-Amzn-Bedrock-Content-Type` on the streaming operation, zero header members on Converse/ConverseStream. Guard for drift suggested inline.
- Model ID encoding is identical to botocore's serializer for `:` and for ARNs (`%3A`, `%2F`), signing name `bedrock`, and `make test` gives `1324 passed, 5 skipped, 1 xpassed` locally.

**1. Endpoint resolution diverges from the sync client.** `_base_url` hardcodes `bedrock-runtime.{region}.amazonaws.com`, while `create_aws_client` lets botocore resolve it. Measured through the factory `use_async_transport=True` uses:

| case | `create_aws_client` (sync) | `create_bedrock_async_client` at PR head | with the inline suggestion |
|---|---|---|---|
| `us-gov-west-1` + `AWS_USE_FIPS_ENDPOINT=true` | `bedrock-runtime-fips.us-gov-west-1.amazonaws.com` | `bedrock-runtime.us-gov-west-1.amazonaws.com` | matches |
| `AWS_ENDPOINT_URL_BEDROCK_RUNTIME=https://bedrock.example.internal:8443` | `bedrock.example.internal:8443` | `bedrock-runtime.us-east-1.amazonaws.com` | matches |
| `AWS_ENDPOINT_URL=http://localhost:4566` | `localhost:4566` | `bedrock-runtime.us-east-1.amazonaws.com` | matches |
| `AWS_USE_DUALSTACK_ENDPOINT=true` | `bedrock-runtime.us-east-1.api.aws` | `bedrock-runtime.us-east-1.amazonaws.com` | matches |
| `cn-north-1` | `...amazonaws.com.cn` | `...amazonaws.com` | matches |
| no region anywhere | `NoRegionError` at construction | `.None.amazonaws.com`, error at first request | `NoRegionError` |

FIPS-only accounts and anyone pointing the SDK at a proxy or LocalStack would silently send to the public endpoint. The factory already builds a `boto3.Session`, so the smallest fix is to let it build the client too and read `.meta.endpoint_url` / `.meta.region_name` (suggestion on `utils.py`). Reusing `self.client.meta` from the model would avoid the extra client, but `_setup_async_client` runs before `self.client` exists and the factory is public, so resolving it in the factory covers both.

The unit suite cannot see this today because the autouse `mock_boto3_client` fixture swaps `boto3.Session` for a `MagicMock`. This regression test restores the real class for its scope and is red on PR head, green with the suggestion:

```python
import boto3.session
import pytest
from pydantic import SecretStr

from langchain_aws.utils import create_bedrock_async_client

_CREDS = {"aws_access_key_id": SecretStr("k"), "aws_secret_access_key": SecretStr("s")}
_FIPS = "https://bedrock-runtime-fips.us-gov-west-1.amazonaws.com"
_OVERRIDE = "https://bedrock.example.internal:8443"
_CN = "https://bedrock-runtime.cn-north-1.amazonaws.com.cn"


@pytest.mark.parametrize(
    ("region", "env", "expected"),
    [
        ("us-gov-west-1", {"AWS_USE_FIPS_ENDPOINT": "true"}, _FIPS),
        ("us-east-1", {"AWS_ENDPOINT_URL_BEDROCK_RUNTIME": _OVERRIDE}, _OVERRIDE),
        ("cn-north-1", {}, _CN),
    ],
)
def test_built_client_follows_botocore_endpoint_resolution(
    monkeypatch: pytest.MonkeyPatch, region: str, env: dict, expected: str
) -> None:
    """FIPS, env overrides and partitions must match what boto3 would use."""
    # The autouse fixture swaps `boto3.Session` for a MagicMock; endpoint
    # resolution is offline, so use the real class here.
    monkeypatch.setattr("boto3.Session", boto3.session.Session)
    for key, value in env.items():
        monkeypatch.setenv(key, value)
    client = create_bedrock_async_client(region_name=region, **_CREDS)
    assert client._base_url == expected
```

```
PR head:            3 failed in 0.40s
  - https://bedrock-runtime-fips.us-gov-west-1.amazonaws.com
  + https://bedrock-runtime.us-gov-west-1.amazonaws.com
with suggestion:    5 passed in 0.88s   (incl. the two header-map tests below)
full make test:     1329 passed, 5 skipped, 1 xpassed
ruff check / ruff format --check / mypy: clean
```

**2. `retries["max_attempts"]` is one attempt short of boto3** (not blocking). botocore treats `max_attempts` as retries after the initial request (`botocore/args.py:600`, "actually means max *retry* attempts"), and `ChatBedrockConverse.max_retries` maps straight onto it. Measured with a permanent 500 and `asyncio.sleep` patched out:

| `max_attempts` | boto3 total attempts | `BedrockAsyncClient` at PR head | with suggestion |
|---|---|---|---|
| 0 | 1 | 1 | 1 |
| 1 | 2 | 1 | 2 |
| 2 | 3 | 2 | 3 |
| 4 | 5 | 4 | 5 |

So `max_retries=2` retries twice on the sync path and once on the async path, and `total_max_attempts` is silently ignored. Suggestion inline on `_max_attempts`, plus a one-token change to `test_retries_are_exhausted_then_raise` so it keeps asserting two sends.

Both suggestions applied together keep every existing test green. Clean, opt-in, and additive throughout; I would like to see it land.

_Disclosure: an AI agent (Claude) helped reproduce and draft the measurements above; I reviewed and ran everything before posting._

libs/aws/langchain_aws/utils.py
```diff
@@ -304,6 +302,151 @@ def create_aws_client(
...
+            "Please check that the specified profile name and/or its credentials are "
+            f"valid. Service error: {e}"
+        ) from e
+
+    return BedrockAsyncClient(
+        region_name=_resolve_region_name(region_name) or session.region_name,
+        credentials=session.get_credentials(),
+        endpoint_url=endpoint_url,
+        config=config,
+    )
```
kimnamu, line 406: Appreciate that the factory already goes through `boto3.Session` for credentials: the endpoint can ride the same session. Letting botocore build the client and reading `.meta` picks up FIPS/dual-stack variants, partitions and the `AWS_ENDPOINT_URL[_BEDROCK_RUNTIME]` overrides that `create_aws_client` honors on the sync path. Measured before/after in the review body (six cases, all match afterwards); `NoRegionError` now surfaces at construction like boto3 instead of `bedrock-runtime.None.amazonaws.com` at first request.

```suggestion
    # Resolve the endpoint through botocore so FIPS/dual-stack variants, partitions
    # and AWS_ENDPOINT_URL[_BEDROCK_RUNTIME] overrides match the synchronous client.
    meta = session.client(
        "bedrock-runtime",
        region_name=_resolve_region_name(region_name),
        endpoint_url=endpoint_url,
        config=config,
    ).meta

    return BedrockAsyncClient(
        region_name=meta.region_name,
        credentials=session.get_credentials(),
        endpoint_url=meta.endpoint_url,
        config=config,
    )
```

libs/aws/langchain_aws/async_client.py
```diff
@@ -0,0 +1,770 @@
...
+        return f"https://bedrock-runtime.{self._region_name}.amazonaws.com"
+
+    @property
+    def _max_attempts(self) -> int:
+        """Total attempts per request, from the botocore retry config."""
+        retries = getattr(self._config, "retries", None) or {}
+        try:
+            return max(1, int(retries.get("max_attempts", _DEFAULT_MAX_ATTEMPTS)))
+        except (TypeError, ValueError):
+            return _DEFAULT_MAX_ATTEMPTS
```
kimnamu, line 324: Thanks for spelling out the retry semantics in the docstring; one detail to align with botocore. In `botocore/args.py::_compute_retry_max_attempts` (1.43.75, line 600) `max_attempts` "actually means max *retry* attempts", i.e. total = `max_attempts + 1`, and `total_max_attempts` wins when both are set. Measured with a permanent 500: `max_attempts=2` gives boto3 3 sends and this client 2 (table in the review body). Not blocking.

```suggestion
        retries = getattr(self._config, "retries", None) or {}
        try:
            if "total_max_attempts" in retries:
                return max(1, int(retries["total_max_attempts"]))
            # botocore's `max_attempts` counts retries after the initial request.
            max_attempts = int(retries.get("max_attempts", _DEFAULT_MAX_ATTEMPTS - 1))
            return max(1, max_attempts + 1)
        except (TypeError, ValueError):
            return _DEFAULT_MAX_ATTEMPTS
```

libs/aws/tests/unit_tests/test_async_client.py
```diff
@@ -0,0 +1,640 @@
...
+
+async def test_retries_are_exhausted_then_raise() -> None:
+    throttled = _FakeResponse(
+        status_code=429,
+        json_body={"message": "slow down"},
+        content=b'{"message": "slow down"}',
+        headers={"x-amzn-errortype": "ThrottlingException"},
+    )
+    client, http_client = _client(
+        throttled, config=mock.MagicMock(retries={"max_attempts": 2})
```
kimnamu, line 462: Thanks for covering the exhausted path explicitly. With the `_max_attempts` change above, `max_attempts=1` is the config that yields exactly two sends (initial + one retry), matching boto3's `1 -> 2` row, so the assertion below stays `== 2` untouched.

```suggestion
        throttled, config=mock.MagicMock(retries={"max_attempts": 1})
```

libs/aws/langchain_aws/async_client.py
```diff
@@ -0,0 +1,770 @@
...
+_DEFAULT_POOL_CONNECTIONS = 10
+_DEFAULT_MAX_ATTEMPTS = 5
+_CREDENTIAL_CACHE_SECONDS = 60.0
+_MAX_BACKOFF_SECONDS = 20.0
+
+# Members of the Invoke operations that travel as headers rather than in the JSON
+# body, taken from the `bedrock-runtime` service model. Converse and ConverseStream
+# have no header-bound members, so their kwargs all belong in the body. Dropping
+# these would silently disable guardrails, so they are mapped explicitly.
+_INVOKE_HEADER_MEMBERS: Mapping[str, str] = {
```
kimnamu, line 45: Thanks for flagging this map as the spot to check; I did, against the `bedrock-runtime` service model in botocore 1.43.75, and it matches exactly (8 members, same header names, `X-Amzn-Bedrock-Accept` for the streaming op, no header members on Converse/ConverseStream). Since it is transcribed, a two-assert test pins it so a `uv.lock` bump that adds a header shows up in CI. Runs offline under `--disable-socket`; removing `serviceTier` from the map makes it fail.

```python
from botocore.session import get_session

from langchain_aws.async_client import (
    _INVOKE_HEADER_MEMBERS,
    _INVOKE_STREAM_HEADER_MEMBERS,
)


def _header_members(operation: str) -> dict:
    model = get_session().get_service_model("bedrock-runtime")
    members = model.operation_model(operation).input_shape.members
    return {
        name: member.serialization["name"]
        for name, member in members.items()
        if member.serialization.get("location") == "header"
    }


def test_invoke_header_members_match_service_model() -> None:
    """Catch drift between the transcribed header maps and botocore's model."""
    assert _header_members("InvokeModel") == dict(_INVOKE_HEADER_MEMBERS)
    assert _header_members("InvokeModelWithResponseStream") == dict(
        _INVOKE_STREAM_HEADER_MEMBERS
    )


def test_converse_has_no_header_members() -> None:
    """Converse kwargs are all body-bound, as `converse()` assumes."""
    assert _header_members("Converse") == {}
    assert _header_members("ConverseStream") == {}
```

```
green: 2 passed in 0.37s
red (serviceTier line removed): 1 failed, 1 passed
E  {'serviceTier': 'X-Amzn-Bedrock-Service-Tier'}
```

gouveags:
@roihezkiRaven quick heads-up for the rebase: #1287 has now merged and adds `messageStop` completion validation to `ChatBedrockConverse`. The native `_astream()` path should preserve the same behavior: missing `messageStop` is incomplete, while missing metadata alone is valid.

@michaelnchin tagging you for visibility since you helped settle that contract. Happy to re-review the async path once it’s updated, or help if anything from #1287 is unclear.

libs/aws/langchain_aws/utils.py
roihezkiRaven: Applied, with one adjustment: the `session.client(...)` call sits **inside** the existing `try/except BotoCoreError`, so a missing region surfaces as the same `ValueError("Could not load credentials ...")` that `create_aws_client` raises rather than a bare `NoRegionError`. Identical error surface on both transports.

Your regression test is in as `test_async_client_endpoint_matches_botocore_resolution` in `tests/unit_tests/test_utils.py`, parametrized over FIPS, dual-stack, `AWS_ENDPOINT_URL_BEDROCK_RUNTIME` and `cn-north-1`, plus `test_async_client_without_a_region_fails_at_construction`. I confirmed all four are red on the previous head:

```
- https://bedrock-runtime-fips.us-gov-west-1.amazonaws.com
+ https://bedrock-runtime.us-gov-west-1.amazonaws.com
- https://bedrock-runtime.us-east-1.api.aws
+ https://bedrock-runtime.us-east-1.amazonaws.com
- https://bedrock.example.internal
+ https://bedrock-runtime.us-east-1.amazonaws.com
```

Thank you for catching this one — silently sending a FIPS-only account to the public endpoint is exactly the kind of divergence that makes an opt-in transport untrustworthy.

libs/aws/langchain_aws/async_client.py
roihezkiRaven: Applied. I read `_compute_retry_max_attempts` in botocore 1.43.75 and confirmed both halves: `max_attempts` counts retries after the initial request, and `total_max_attempts` is popped over it when both are set.

This mattered more than the table suggests — `ChatBedrockConverse.max_retries` feeds straight into `retries.max_attempts` (`bedrock_converse.py:1080`), so the same `max_retries=2` meant three sends on the sync path and two here. `total_max_attempts` was also ignored outright.

Covered by `test_max_attempts_follows_botocore_semantics`, parametrized over your table plus the `total_max_attempts` precedence case and the default.

libs/aws/tests/unit_tests/test_async_client.py
roihezkiRaven: Applied as `max_attempts: 1`, assertion left at `== 2`.

libs/aws/langchain_aws/async_client.py
roihezkiRaven: Added as `test_invoke_header_members_match_the_service_model` and `test_converse_operations_have_no_header_members`. I verified the drift check independently: 8 modelled members, and dropping `serviceTier` makes it red.

Thank you for actually checking this against the service model — it was the item I was least able to verify myself, and it is the one where drift would be silent and security-relevant.

roihezkiRaven:
Rebased onto `main` (now at `6165ac8`) and addressed all review feedback. Head is `5b06047`, still five commits.

### @kimnamu — both findings applied

Replies are on the individual threads. Summary:

| Finding | Status |
|---|---|
| Endpoint resolution diverges from the sync client | Applied. Endpoint now resolved through the session's own `bedrock-runtime` client `meta`. The call sits inside the existing `BotoCoreError` guard, so a missing region raises the same `ValueError` the sync factory raises. |
| `retries["max_attempts"]` one attempt short | Applied. `total_max_attempts` wins; `max_attempts` counts retries after the initial request. |
| Header-map drift test | Added, both assertions. |

Every test you supplied is in, and I confirmed each is red on the previous head before applying the fix.

### @gouveags — #1287 ported to the native async path

`_astream` now tracks `messageStop` and raises the same `ConnectionError` after normal exhaustion, with the same semantics you settled:

- Completion requires `messageStop`; metadata stays optional.
- The check runs **after** the `try/finally`, so a caller that stops reading early gets `GeneratorExit` at the `yield` and no spurious error — matching `_stream`.
- The stream is still closed on an incomplete response.

The message is a shared module constant so the two paths cannot drift. Five async regressions mirror your sync coverage: missing `messageStop` for text and for a partial tool call, complete-without-metadata, close-on-incomplete, and early-consumer-close. All three enforcement tests are red without the `_astream` change.

The rebase surfaced this as a real gap rather than a hypothetical: `test_astream_falls_back_to_executor_without_async_client` went red immediately, since the fallback delegates to `_stream`. Three canned async fixtures predated the contract and now model complete event sequences.

### Verification

```
make lint     ruff + ruff format + mypy, 153 source files — clean
make test     1418 passed, 3 skipped, 1 xpassed
integration   9 passed against real Bedrock (us-east-1)
```

The live `astream` round-trip is the one that matters for #1287: it confirms real Converse streams deliver `messageStop` through this transport, so the new check does not false-positive against the service.

Rebase conflicts were two, both mechanical: #1287's `messageStop` tracking landed in the loop this PR extracts into `_build_stream_chunk`, and an import block in `chat_models/bedrock.py`. Nothing else in the 26 upstream commits touched this surface.

@michaelnchin this is green and conflict-free whenever you have time. Happy to split it or adjust anything.

gouveags:
All good for me!

roihezkiRaven:
@gouveags 
Thanks! Any idea who can help make this live?

gouveags:
@michaelnchin has been really helpful!

kimnamu:
@roihezkiRaven thank you for applying all three items so carefully, and for the extra parametrized coverage (`total_max_attempts` precedence, `AWS_USE_DUALSTACK_ENDPOINT`, the no-region case) that went beyond what I suggested.

> Every test you supplied is in, and I confirmed each is red on the previous head before applying the fix.

I re-ran my original measurements against the current head `5663b4e` (after the `main` merge) and everything I flagged now matches the sync client. Nothing left from my side on the auth/region/serialization surface I reviewed; I did not re-review the model refactors.

| Review item | At `4f84ed2` | At `5663b4e` |
|---|---|---|
| Endpoint resolution via `create_bedrock_async_client` | 6 of 7 cases diverged from `boto3` | 7 of 7 match, incl. no region: same `ValueError` with `NoRegionError` cause as `create_aws_client` |
| `retries["max_attempts"]` total sends for N = 0/1/2/4 | 1/1/2/4 (boto3: 1/2/3/5) | 1/2/3/5 |
| `_INVOKE_HEADER_MEMBERS` drift guard | none | `test_invoke_header_members_match_the_service_model` red with `serviceTier` dropped, green otherwise |

```
make test   1425 passed, 3 skipped, 1 xpassed
make lint   ruff check / ruff format --diff / mypy (153 source files): clean
```

<details>
<summary>Measurements at 5663b4e</summary>

Endpoint seen by `boto3.Session().client("bedrock-runtime").meta.endpoint_url` vs `create_bedrock_async_client(...)._base_url`; each case runs in a subprocess with `AWS_*` cleared and only static test keys set.

| case | boto3 (sync path) | `create_bedrock_async_client(...)._base_url` |
|---|---|---|
| us-east-1 default | `https://bedrock-runtime.us-east-1.amazonaws.com` | same |
| us-gov-west-1 + `AWS_USE_FIPS_ENDPOINT=true` | `https://bedrock-runtime-fips.us-gov-west-1.amazonaws.com` | same |
| us-east-1 + `AWS_USE_DUALSTACK_ENDPOINT=true` | `https://bedrock-runtime.us-east-1.api.aws` | same |
| us-east-1 + `AWS_ENDPOINT_URL_BEDROCK_RUNTIME=https://bedrock.example.internal:8443` | `https://bedrock.example.internal:8443` | same |
| us-east-1 + `AWS_ENDPOINT_URL=http://localhost:4566` | `http://localhost:4566` | same |
| cn-north-1 default | `https://bedrock-runtime.cn-north-1.amazonaws.com.cn` | same |
| no region anywhere | `create_aws_client`: `ValueError: Could not load credentials ...` (`__cause__=NoRegionError`) | same type, message and cause |

Retry attempts on a permanent `500 InternalServerException` (`before-send` hook for boto3, `httpx.MockTransport` for the async client, backoff sleep patched out):

| `retries={"max_attempts": N}` | boto3 total HTTP attempts | `BedrockAsyncClient` total HTTP attempts |
|---|---|---|
| 0 | 1 | 1 |
| 1 | 2 | 2 |
| 2 | 3 | 3 |
| 4 | 5 | 5 |

Header drift guard with the `serviceTier` line removed from `_INVOKE_HEADER_MEMBERS`:

```
E   AssertionError: assert {...} == {...}
E     Left contains 1 more item:
E     {'serviceTier': 'X-Amzn-Bedrock-Service-Tier'}
```

Model ID URL encoding is unchanged and still identical to botocore for `:` and ARN ids (`%3A`, `%2F`).

</details>

_Disclosure: Claude (via Claude Code) helped run the measurements above; I reviewed and ran everything before posting._

michaelnchin (commented):
Hi @roihezkiRaven, thanks for working on this!

I greatly appreciate the effort you've put into it this. However, AWS is currently finalizing a brand new [natively async Python SDK](https://github.com/aws/aws-sdk-python/tree/develop/clients/aws-sdk-bedrock-runtime) to replace Boto, and this is expected to be ready for GA in the near future. I'm really hesitant to accept a completely separate implementation now as it would run counter to that effort + maintenance concerns before the official async SDK arrives.

In regards to why we haven't adopted the new SDK already - we actually do use it for the `ChatBedrockNovaSonic` model, strictly in an experimental capacity (since that model specifically needs bidirectional streaming, which Boto doesn't support), but the pre-prod nature of the package also means that rapid upstream changes have already broken the model without warning on a few occasions. Since ChatBedrockConverse and the other classes mentioned are on a more critical path given that they constitute the majority of `langchain-aws`' current user traffic, we're waiting for the SDK to stabilize before adopting it more generally across the package. Will let you know when we are able to start with this.

roihezkiRaven:
Hi @michaelnchin, thanks for the comment. Happy to hear there is going to be a new native async client. I guess this PR is no longer relevant, so I assume we can close it.

PR #1260, opened 2026-09-04
fix(aws): allow auto tool_choice for DeepSeek V3 models

### LINKED ISSUE #1259: `ChatBedrockConverse` prevents DeepSeek V3 from returning text whenever tools are bound (introduced by #688)
Since #688, `ChatBedrockConverse` rewrites a missing `tool_choice` to `any` for `deepseek.v3*` models and omits `auto` from their capability table. If tools are in use, the model is always forced to call a tool and can never answer in text, and asking for `auto` explicitly is refused by lanchain-aws code before any request is made. Bedrock itself accepts `auto` (and no `toolChoice`) on both DeepSeek V3.1 and V3.2 today.

Versions: langchain-aws 1.7.5 (also reproduced on 1.6.0 and 1.7.4), langchain-core 1.6.1, boto3 1.43.82. Region us-east-1 unless noted.

## The Converse API accepts tool selection as documented

With tools, no `toolChoice` (Bedrock's default is `auto`), asked for text:

```python
import boto3
rt = boto3.client("bedrock-runtime", region_name="us-east-1")
tool = {"toolSpec": {"name": "get_weather", "description": "Get the weather for a city.",
        "inputSchema": {"json": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}}}
r = rt.converse(modelId="deepseek.v3.2",
                messages=[{"role": "user", "content": [{"text": "Reply with only the text OK. Do not use any tools."}]}],
                inferenceConfig={"maxTokens": 100}, toolConfig={"tools": [tool]})
print(r["stopReason"], r["output"]["message"]["content"])
# end_turn [{'text': 'OK'}]
```

Asked for the weather instead, the same request returns a valid `toolUse` block.

## Through langchain-aws `auto` is incorrectly refused

```python

@tool
def get_weather(city: str) -> str:
    """Get the weather for a city."""
    return f"Sunny in {city}"

llm = ChatBedrockConverse(model="deepseek.v3.2", provider="deepseek", region_name="us-east-1", max_tokens=100)
print(llm.supports_tool_choice_values)                      # ('any', 'tool')  <- no 'auto'

msg = llm.bind_tools([get_weather]).invoke("Reply with only the text OK. Do not use any tools.")
print(msg.tool_calls)                                        # [{'name': 'get_weather', ...}]  <- forced tool call
# request body sent: toolConfig.toolChoice == {"any": {}}

llm.bind_tools([get_weather], tool_choice="auto")
# ValueError: Model deepseek.v3.2 does not currently support tool_choice of type auto.
```

<details><summary>root cause:</summary>
Both come from `_resolve_tool_choice` in `bedrock_converse.py`:

```python
if not tool_choice:
    if "deepseek.v3" in self._get_base_model():
        return _format_tool_choice("any")      # None -> any
    return None
...
if tool_choice_type in supported:              # 'auto' not in ('any', 'tool') -> raise
```
</details>

## Other providers are unaffected

Same code, `model="us.anthropic.claude-haiku-4-5-20251001-v1:0"`:

```python
llm.bind_tools([get_weather]).invoke("Reply with only the text OK. Do not use any tools.").content
# 'OK'   (no toolChoice sent; caps ('auto', 'any', 'tool'))
```

Not a problem for some other model families: Raw Converse behaves the same for Claude, Grok 4.6, GPT-5.6 and Qwen3, and when `toolChoice` not provided as a parameter, we get text when asked and a tool call when appropriate.

# Impact

On DeepSeek v3 with tools configured, all model calls go out with `tool_choice='any'`, preventing text responses.

# Proposed fix

1. Remove the `deepseek.v3` branch in `_resolve_tool_choice`, and we get Bedrock defaults.
2. Add `"auto"` to the DeepSeek entries in the `supports_tool_choice_values` table.

<details>
<summary>Further details: measurements and origin</summary>

#927 doesn't resolve DeepSeek. Workaround on current releases: pass `supports_tool_choice_values=("auto", "any")` at construction and set `tool_choice="auto"` explicitly (we used a custom middleware to do this).

#688 said: "DeepSeek-V3.1 does not work with the `auto` and named `tool` choice values. While these options will not be rejected on Converse API side, only `any` will produce a valid tool call output." and "DeepSeek-V3.2-Exp supports `tool` choice in addition to `any`, but `auto` is still not supported." I cannot say whether that held in October 2025; it does not hold now.

Raw `Converse`, prompt "What's the weather in Boston? Use the tool.", valid `toolUse` (correct name, non-empty `city`) per 3 trials:

| model | region | no toolChoice | auto | tool |
|---|---|---|---|---|
| deepseek.v3-v1:0 (V3.1) | us-west-2 | 3/3 | 3/3 | 3/3 |
| deepseek.v3.2 | us-west-2 | 3/3 | 3/3 | 3/3 |
| deepseek.v3.2 | us-east-1 | 3/3 | 3/3 | 3/3 |

Prompt "Reply with only the text OK.": `end_turn` with a text block under no toolChoice and under `auto`; `tool_use` under `any`, as expected.

Two-turn loop through `ChatBedrockConverse.bind_tools` (tool call, tool result, second call): the second call is sent with `toolChoice: {"any": {}}` and returns another tool call rather than the sentence asked for.

Full reproducer with an outbound-request tap, showing the `toolChoice` value actually sent for each case:

```python
"""Demo: langchain-aws forces `toolChoice: any` on DeepSeek V3 whenever tools are bound."""
import json
import boto3
from langchain_core.messages import HumanMessage, ToolMessage

MODEL, REGION = "deepseek.v3.2", "us-east-1"

sent: list[dict] = []
_orig_client = boto3.Session.client
def _tapped_client(self, *args, **kwargs):
    client = _orig_client(self, *args, **kwargs)
    if (args[0] if args else kwargs.get("service_name")) == "bedrock-runtime":
        client.meta.events.register("before-call.bedrock-runtime.Converse",
                                    lambda params, **_: sent.append(json.loads(params["body"])))
    return client
boto3.Session.client = _tapped_client

@tool
def get_weather(city: str) -> str:
    """Get the weather for a city."""
    return f"Sunny, 22C in {city}"

def tool_choice_sent():
    return sent[-1].get("toolConfig", {}).get("toolChoice") if sent else None

def describe(msg):
    return f"TOOL CALL: {msg.tool_calls[0]['name']}" if msg.tool_calls else f"TEXT: {msg.content!r:.40}"

llm = ChatBedrockConverse(model=MODEL, provider="deepseek", region_name=REGION, max_tokens=200)
ASK = "Reply with only the text OK. Do not use any tools."

sent.clear(); r = llm.invoke(ASK)
print("1. no tools:            ", tool_choice_sent(), describe(r))
bound = llm.bind_tools([get_weather])
sent.clear(); r = bound.invoke(ASK)
print("2. tools, no choice:    ", tool_choice_sent(), describe(r))
try:
    llm.bind_tools([get_weather], tool_choice="auto").invoke(ASK)
except ValueError as exc:
    print("3. tools, auto:          REFUSED:", str(exc)[:80])
msgs = [HumanMessage("What's the weather in Boston? Use the tool, then answer in one sentence.")]
first = bound.invoke(msgs); msgs.append(first)
for c in first.tool_calls:
    msgs.append(ToolMessage(content=get_weather.invoke(c["args"]), tool_call_id=c["id"]))
sent.clear(); final = bound.invoke(msgs)
print("4. after tool result:   ", tool_choice_sent(), describe(final))
```

Output:

```
1. no tools:             None TEXT: 'OK'
2. tools, no choice:     {'any': {}} TOOL CALL: get_weather
3. tools, auto:          REFUSED: Model deepseek.v3.2 does not currently support tool_choice of type auto. ...
4. after tool result:    {'any': {}} TOOL CALL: get_weather
```

</details>

jpfcabral:
## Why

`ChatBedrockConverse` forced `tool_choice="any"` for `deepseek.v3*` models whenever tools were bound, and omitted `"auto"` from their capability table. This had two user-facing consequences once tools were in use:

- Every request went out with `toolChoice: {"any": {}}`, so DeepSeek V3 could **never answer in text** — it was always forced into a tool call, even when explicitly asked for a plain text reply.
- Requesting `tool_choice="auto"` was rejected by langchain-aws with a `ValueError` before any request was even sent.

Bedrock's Converse API accepts `auto` (and no `toolChoice`, whose default is `auto`) for both DeepSeek V3.1 and V3.2 today, so the restriction is no longer correct. It originated in #688, when `auto` reportedly did not produce valid tool output for these models; that no longer holds.

## What changed

- Removed the `deepseek.v3` branch in `_resolve_tool_choice`, so with no explicit `tool_choice` the resolved value is `None` and Bedrock's default (`auto`) applies.
- Added `"auto"` to `supports_tool_choice_values` for DeepSeek V3 (now `("auto", "any", "tool")`), so `auto` is no longer refused.
- Updated the affected unit tests to assert the new behavior.

## Testing

- Unit tests: full `tests/unit_tests/` suite passes (1253 passed, 5 skipped). Updated `test_deepseek_supports_tool_choice_values`, `test_deepseek_v3_bind_tools_tool_choice_variants`, and `test_deepseek_v3_bind_tools_default_tool_choice`.
- Lint: `ruff check`, `ruff format`, and `mypy` all clean on the changed files.
- **Live validation against Bedrock** (`deepseek.v3.2`, `us-east-1`), before vs. after:

  | Case | Before (bug) | After (fix) |
  |------|-------------|-------------|
  | `supports_tool_choice_values` | `('any', 'tool')` | `('auto', 'any', 'tool')` |
  | tools bound, no choice, asked for text | `{'any': {}}` → forced tool call | `None` → text `'OK'` |
  | `tool_choice="auto"` | `ValueError` (refused) | accepted |
  | after a tool result, asked to answer | `{'any': {}}` → another tool call | `None` → text answer |

## Areas worth careful review

- I unified DeepSeek V3.1 (`v3-v1`) and later V3 entries to `("auto", "any", "tool")`. The issue's literal proposal was `("auto", "any")` for V3.1, but the reporter's live data shows named `tool` selection also works on V3.1. My own live validation only covered `deepseek.v3.2` (the only DeepSeek V3 model enabled in the test account/region), so the V3.1 `tool` support relies on the issue's data rather than a test I ran directly.

Closes #1259

---

> [!NOTE]
> This contribution was prepared with the assistance of an AI agent (Kiro). The diagnosis was reproduced live against Bedrock, and the code, tests, and this description were reviewed before submission.

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -1165,10 +1165,9 @@ class ChatBedrockConverse(BaseChatModel):
             elif "nova" in base_model:
                 self.supports_tool_choice_values = ("auto", "any", "tool")
             elif "deepseek" in base_model and "r1-v1" not in base_model:
-                if "v3-v1" in base_model:
-                    self.supports_tool_choice_values = ("any",)
-                else:
-                    self.supports_tool_choice_values = ("any", "tool")
+                # DeepSeek V3 models accept `auto` (Bedrock's default), `any`,
+                # and named `tool` selection on the Converse API.
+                self.supports_tool_choice_values = ("auto", "any", "tool")
             else:
                 self.supports_tool_choice_values = ()
 
@@ -1556,8 +1555,6 @@ class ChatBedrockConverse(BaseChatModel):
                 not enabled (no safe downgrade possible).
         """
         if not tool_choice:
-            if "deepseek.v3" in self._get_base_model():
-                return _format_tool_choice("any")
             return None
 
         formatted = _format_tool_choice(tool_choice)
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -340,10 +340,12 @@ def test_llama_bind_tools_tool_choice_variants(
     "model,expected_values",
     [
         ("us.deepseek.r1-v1:0", ()),
-        ("deepseek.v3-v1:0", ("any",)),
+        ("deepseek.v3-v1:0", ("auto", "any", "tool")),
+        ("deepseek.v3.2", ("auto", "any", "tool")),
         (
             "deepseek.v3-x:0",
             (
+                "auto",
                 "any",
                 "tool",
             ),
@@ -392,20 +394,30 @@ def test_deepseek_v3_bind_tools_tool_choice_variants() -> None:
         "any": {}
     }
 
-    with pytest.raises(ValueError):
-        chat_model.bind_tools([GetWeather], tool_choice="auto")
+    # `auto` is accepted by Bedrock for DeepSeek V3 and must not be refused
+    # (regression test for issue #1259).
+    chat_model_with_auto = chat_model.bind_tools([GetWeather], tool_choice="auto")
+    assert cast(RunnableBinding, chat_model_with_auto).kwargs["tool_choice"] == {
+        "auto": {}
+    }
 
-    with pytest.raises(ValueError):
-        chat_model.bind_tools([GetWeather], tool_choice="GetWeather")
+    chat_model_with_named = chat_model.bind_tools(
+        [GetWeather], tool_choice="GetWeather"
+    )
+    assert cast(RunnableBinding, chat_model_with_named).kwargs["tool_choice"] == {
+        "tool": {"name": "GetWeather"}
+    }
 
 
 def test_deepseek_v3_bind_tools_default_tool_choice() -> None:
     chat_model = ChatBedrockConverse(model="deepseek.v3-v1:0", region_name="us-east-1")  # type: ignore[call-arg]
 
+    # With no explicit tool_choice, the resolved tool_choice must be None so no
+    # toolChoice is sent and Bedrock's default (`auto`) applies, letting the
+    # model answer in text instead of being forced into a tool call
+    # (regression test for issue #1259).
     chat_model_with_tools = chat_model.bind_tools([GetWeather])
-    assert cast(RunnableBinding, chat_model_with_tools).kwargs["tool_choice"] == {
-        "any": {}
-    }
+    assert cast(RunnableBinding, chat_model_with_tools).kwargs["tool_choice"] is None
 
 
 def test__messages_to_bedrock() -> None:


michaelnchin (commented):
Hi @jpfcabral, I noticed DeepSeek V3.1 and V3.2 behaving a bit different than noted in this PR, specifically in regards to `tool` mode, and use with reasoning enabled. 

Will open another PR to replace this one, but thank you for working on this!

PR #1261, opened 2026-09-05
fix(aws): Update DeepSeek V3.x tool choice support on ChatBedrockConverse

### LINKED ISSUE #1259: `ChatBedrockConverse` prevents DeepSeek V3 from returning text whenever tools are bound (introduced by #688)

### LINKED ISSUE #1270: DeepSeek v3.2 forced "tool" choice silently fails for any non-flat schema (should be "any" only)
`ChatBedrockConverse.validate_environment` grants DeepSeek non-R1/non-v3-v1 models (currently just `deepseek.v3.2`) `supports_tool_choice_values = ("any", "tool")`. AWS's own [ToolChoice API reference](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolChoice.html) restricts `tool` (forced specific tool) to "Anthropic Claude 3 and Amazon Nova models" only, and live testing confirms forced `tool` choice is unreliable for DeepSeek v3.2 for any non-trivial (non-flat) schema.

Related to #1259, but a distinct symptom: that issue is about `_resolve_tool_choice` force-rewriting a missing `tool_choice` to `any` (breaking plain-text responses) and refusing `auto`. This issue is about the `tool` (forced-specific) capability grant itself being unreliable, independent of that.

## Repro

Via `ChatBedrockConverse(model="deepseek.v3.2", ...)`, forcing `toolChoice={"tool": {"name": "..."}}` (e.g. via `with_structured_output(schema, method="function_calling")`) reliably fails to produce a tool call for any schema containing a nested `object` or `array` property — the model just answers in plain text (`stopReason: "end_turn"`) instead. Forcing `toolChoice={"any": {}}` with the exact same schema reliably succeeds.

Tested directly against the Bedrock Converse API (via an internal gateway, not through `langchain-aws`) with `deepseek.v3.2`, 3 repro attempts each, reproduced consistently for all of:

**A flat schema (works under `tool`):**
```json
{"type": "object", "properties": {"answer": {"type": "string"}, "justification": {"type": "string"}}, "required": ["answer", "justification"]}
```
`toolChoice={"tool": {"name": "AnswerWithJustification"}}` → `stopReason: "tool_use"`, valid `toolUse` block. 1/1 success.

**A schema with one nested object property (fails under `tool`, 2/2):**
```json
{"type": "object", "properties": {"name": {"type": "string"}, "address": {"type": "object", "properties": {"street": {"type": "string"}, "city": {"type": "string"}}, "required": ["street", "city"]}}, "required": ["name", "address"]}
```
`toolChoice={"tool": {"name": "PersonNestedOnly"}}` → `stopReason: "end_turn"`, plain text, no tool call. 2/2 failures.

**A schema with only a flat array-of-objects property (fails under `tool`, 2/2):**
```json
{"type": "object", "properties": {"friends": {"type": "array", "items": {"type": "object", "properties": {"street": {"type": "string"}, "city": {"type": "string"}}, "required": ["street", "city"]}}}, "required": ["friends"]}
```
Same failure mode, 2/2.

**A schema with only a flat array-of-strings property (fails under `tool`, 2/2):**
```json
{"type": "object", "properties": {"names": {"type": "array", "items": {"type": "string"}}}, "required": ["names"]}
```
Same failure mode, 2/2 — so it's not specific to nested objects, any `array` or `object`-typed property triggers it.

**A schema combining nested object + array-of-objects (fails under `tool`, 3/3; succeeds under `any`, 3/3):**
```json
{"type": "object", "properties": {"name": {"type": "string"}, "address": {"type": "object", "properties": {"street": {"type": "string"}, "city": {"type": "string"}}, "required": ["street", "city"]}, "friends": {"type": "array", "items": {"type": "object", "properties": {"street": {"type": "string"}, "city": {"type": "string"}}, "required": ["street", "city"]}}}, "required": ["name", "address", "friends"]}
```
`toolChoice={"tool": {...}}` → `stopReason: "end_turn"`, plain text, 3/3. `toolChoice={"any": {}}` → `stopReason: "tool_use"`, valid `toolUse` matching the schema, 3/3.

## Proposed fix

Drop `"tool"` from the DeepSeek non-R1/non-v3-v1 entry in `supports_tool_choice_values`:

```python
elif "deepseek" in base_model and "r1-v1" not in base_model:
    self.supports_tool_choice_values = ("any",)
```

(matching what `deepseek.v3-v1:0` already gets, and matching AWS's own documented `tool`-choice restriction to Claude 3 / Nova). This mainly affects `with_structured_output(method="function_calling")`, which picks `tool_choice` by checking `"tool" in supports_tool_choice_values` first — with the current classification, any structured-output schema beyond the most trivial flat one silently returns `parsed=None` for DeepSeek v3.2.

## Versions

langchain-aws 1.6.2 through 1.7.5 (latest) — the DeepSeek branch in `validate_environment` is unchanged across that range.

michaelnchin:
Fixes #1259

#688, merged in Oct. 2025, restricted DeepSeek V3 models to `any` ToolChoice mode (based on Bedrock V3.1's tested behavior and upstream V3.2-Exp's docs). However, since then, V3.1's tool choice behavior has changed, as well as that of the stable V3.2 version Bedrock added in Dec. 2025.

Manually ran `tool_choice` value tests against Converse API to confirm the current situation:
- With thinking **disabled** (model default), both DeepSeek V3.1 and V3.2 behave the same:
   - Both `auto` and `any` mode now work unconditionally
   - Specific `tool` mode, while still accepted by Converse API, is not obeyed by either model.
- With thinking **enabled** (via `{"reasoning_effort": "high"}`):
  - `any` doesn't produce tool calls on either V3.1 or V3.2
  - `auto` doesn't work on V3.1, but **does** on V3.2 (which matches the DeepSeek [docs](https://api-docs.deepseek.com/guides/tool_calls/#thinking-mode))
  - `tool` mode: same as non-thinking.

Based on these observations, updated the following for DeepSeek V3.1/3.2 on ChatBedrockConverse:
- Allowed ToolChoice values goes to `("auto", "any")`
- Removed the forced `any` ToolChoice mode override for these models in `bind_tools`
- Added `any` -> `auto` auto-downgrade in the case that thinking is also enabled.
  - Note: may need further updates for when Bedrock adds DeepSeek V4, which now uses thinking by default; see: https://github.com/deepseek-ai/DeepSeek-V3/issues/1376

### DIFF
--- a/libs/aws/langchain_aws/chat_models/bedrock_converse.py
+++ b/libs/aws/langchain_aws/chat_models/bedrock_converse.py
@@ -84,6 +84,7 @@ from langchain_aws.utils import (
     create_aws_client,
     parse_model_provider,
     reasoning_effort_additional_fields,
+    thinking_enabled_in_params,
     thinking_forced_tool_use_unsupported,
     thinking_in_params,
     trim_message_whitespace,
@@ -1168,10 +1169,12 @@ class ChatBedrockConverse(BaseChatModel):
             elif "nova" in base_model:
                 self.supports_tool_choice_values = ("auto", "any", "tool")
             elif "deepseek" in base_model and "r1-v1" not in base_model:
-                if "v3-v1" in base_model:
-                    self.supports_tool_choice_values = ("any",)
+                if thinking_enabled_in_params(
+                    base_model, self.additional_model_request_fields or {}
+                ):
+                    self.supports_tool_choice_values = ("auto",)
                 else:
-                    self.supports_tool_choice_values = ("any", "tool")
+                    self.supports_tool_choice_values = ("auto", "any")
             else:
                 self.supports_tool_choice_values = ()
 
@@ -1510,10 +1513,17 @@ class ChatBedrockConverse(BaseChatModel):
             "langchain_core.exceptions.OutputParserException if tool calls are not "
             "generated. Consider adjusting your prompt to ensure the tool is called."
         )
-        if thinking_forced_tool_use_unsupported(self._get_base_model()):
+        base_model = self._get_base_model()
+        if "claude" in base_model and thinking_forced_tool_use_unsupported(base_model):
             additional_context = (
                 "For Claude 3/4 models, you can also support forced tool use "
                 "by disabling `thinking`."
+            )
+            admonition = f"{admonition} {additional_context}"
+        if "deepseek.v3" in base_model:
+            additional_context = (
+                "For DeepSeek V3 models, forced tool use is not available while "
+                "`reasoning_effort` is set in `additional_model_request_fields`."
             )
             admonition = f"{admonition} {additional_context}"
         warnings.warn(admonition)
@@ -1559,8 +1569,6 @@ class ChatBedrockConverse(BaseChatModel):
                 not enabled (no safe downgrade possible).
         """
         if not tool_choice:
-            if "deepseek.v3" in self._get_base_model():
-                return _format_tool_choice("any")
             return None
 
         formatted = _format_tool_choice(tool_choice)
@@ -1572,7 +1580,9 @@ class ChatBedrockConverse(BaseChatModel):
 
         # Thinking-enabled models: downgrade to auto instead of failing.
         if (
-            thinking_in_params(self.additional_model_request_fields or {})
+            thinking_enabled_in_params(
+                self._get_base_model(), self.additional_model_request_fields or {}
+            )
             and "auto" in supported
         ):
             warnings.warn(
--- a/libs/aws/langchain_aws/utils.py
+++ b/libs/aws/langchain_aws/utils.py
@@ -548,7 +548,7 @@ def thinking_on_by_default(model: str) -> bool:
 
 
 def thinking_forced_tool_use_unsupported(model: str) -> bool:
-    """Check if the model rejects forced tool use while thinking is enabled."""
+    """Check if the model rejects or ignores forced tool use with thinking enabled."""
     if "claude-opus-4-8" in model:
         return False
     return any(
@@ -558,8 +558,17 @@ def thinking_forced_tool_use_unsupported(model: str) -> bool:
             "claude-sonnet-4",
             "claude-opus-4",
             "claude-haiku-4",
+            "deepseek.v3",
         )
     )
+
+
+def thinking_enabled_in_params(model: str, params: dict) -> bool:
+    """Check if thinking is explicitly enabled in the request."""
+    if "deepseek.v3" in model:
+        effort = params.get("reasoning_effort")
+        return bool(effort) and str(effort).lower() not in ("none", "low", "medium")
+    return thinking_in_params(params)
 
 
 def reasoning_effort_additional_fields(base_model: str, effort: str) -> Dict[str, Any]:
--- a/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
+++ b/libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py
@@ -340,14 +340,8 @@ def test_llama_bind_tools_tool_choice_variants(
     "model,expected_values",
     [
         ("us.deepseek.r1-v1:0", ()),
-        ("deepseek.v3-v1:0", ("any",)),
-        (
-            "deepseek.v3-x:0",
-            (
-                "any",
-                "tool",
-            ),
-        ),
+        ("deepseek.v3-v1:0", ("auto", "any")),
+        ("deepseek.v3.2", ("auto", "any")),
     ],
 )
 def test_deepseek_supports_tool_choice_values(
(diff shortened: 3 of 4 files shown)


michaelnchin (commented):
Integ workflow passing: https://github.com/langchain-ai/langchain-aws/actions/runs/33953950464

PR #1264, opened 2026-09-06
fix(model-profiles): declare Claude 5 structured output support

### LINKED ISSUE #1263: AutoStrategy does not detect provider-native structured output support for Claude Sonnet 5
### Submission checklist

- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Package (Required)

- [x] langchain
- [ ] langchain-openai
- [ ] langchain-anthropic
- [ ] langchain-classic
- [ ] langchain-core
- [ ] langchain-model-profiles
- [ ] langchain-tests
- [ ] langchain-text-splitters
- [ ] langchain-chroma
- [ ] langchain-deepseek
- [ ] langchain-exa
- [ ] langchain-fireworks
- [ ] langchain-groq
- [ ] langchain-huggingface
- [ ] langchain-mistralai
- [ ] langchain-nomic
- [ ] langchain-ollama
- [ ] langchain-openrouter
- [ ] langchain-perplexity
- [ ] langchain-qdrant
- [ ] langchain-xai
- [ ] Other / not sure / general

### Related Issues / PRs

_No response_

```python
from langchain.agents.factory import _supports_provider_strategy

# Simulating a Bedrock-style Claude Sonnet 5 model identifier
model_name = "bedrock/us.anthropic.claude-sonnet-5"

# This should be True for a Claude Sonnet 5 model (native structured output support),
# but returns False because no regex in FALLBACK_MODELS_WITH_STRUCTURED_OUTPUT matches it.
result = _supports_provider_strategy(model_name)
print(result)  # Actual: False | Expected: True
```

```shell

```

### Description

AutoStrategy incorrectly falls back to ToolStrategy for Claude Sonnet 5 model names when model profile capability data are unavailable.

I am using LangChain's structured output support and relying on AutoStrategy to automatically select between ProviderStrategy and ToolStrategy.

The selection happens in _get_bound_model(). For AutoStrategy, LangChain calls:

_supports_provider_strategy(request.model, tools=request.tools)

If this returns True, LangChain uses ProviderStrategy; otherwise, it uses ToolStrategy.

_supports_provider_strategy() first checks model.profile["structured_output"]. However, when model profile data are unavailable, it falls back to FALLBACK_MODELS_WITH_STRUCTURED_OUTPUT.

The fallback regex list currently recognizes:

claude-sonnet-4-5
claude-sonnet-4-6

but does not recognize:

claude-sonnet-5

As a result, a model name such as:

bedrock/us.anthropic.claude-sonnet-5

does not match any fallback pattern and _supports_provider_strategy() returns False.

The provider prefix is not the problem. The regex boundary (^|[/:.]) correctly allows the . before claude in:

bedrock/us.anthropic.claude-sonnet-5
                    ^

The problem is specifically that the fallback list has no pattern matching the claude-sonnet-5 model family.
Expected behavior

When model profile capability information is unavailable, supported Claude Sonnet 5 model names should be recognized by the fallback capability detection, allowing AutoStrategy to select ProviderStrategy for provider-native structured output.

Actual behavior

For:

bedrock/us.anthropic.claude-sonnet-5

_supports_provider_strategy() returns False, causing:

AutoStrategy -> ToolStrategy

instead of:

AutoStrategy -> ProviderStrategy
Suggested fix

Add an appropriate Claude Sonnet 5 fallback regex, for example:

r"(^|[/:.])claude-sonnet-5(?:-\d{8})?(?:-v\d(?::\d)?)?($|[/:])",

This would support names such as:

claude-sonnet-5
claude-sonnet-5-YYYYMMDD
bedrock/us.anthropic.claude-sonnet-5

The same naming/version conventions used by the existing Claude fallback patterns can therefore be preserved.

### System Info

System Information
------------------
> OS:  Windows
> OS Version:  10.0.26200
> Python Version:  3.13.12 (main, Mar 24 2026, 22:56:40) [MSC v.1944 64 bit (AMD64)]

Package Information
-------------------
> langchain_core: 1.6.0
> langchain: 1.3.15
> langchain_community: 0.4.1
> langsmith: 0.11.1
> deepagents: 0.7.8
> langchain_anthropic: 1.6.1
> langchain_classic: 1.0.8
> langchain_google_genai: 4.3.5
> langchain_litellm: 0.7.0
> langchain_mcp_adapters: 0.3.1
> langchain_protocol: 0.0.18
> langchain_quickjs: 0.3.5
> langchain_text_splitters: 1.1.2
> langgraph_sdk: 0.4.2

Optional packages not installed
-------------------------------
> deepagents-cli

Other Dependencies
------------------
> aiohttp: 3.14.3
> anthropic: 0.120.0
> anyio: 4.14.2
> bsdiff4: 1.2.6
> cachetools: 6.1.0
> cryptography: 48.0.1
> dataclasses-json: 0.6.7
> distro: 1.9.0
> filetype: 1.2.0
> google-genai: 1.75.0
> httpx: 0.28.1
> httpx-sse: 0.4.3
> jsonpatch: 1.33
> langgraph: 1.2.11
> litellm: 1.91.1
> mcp: 1.28.1
> numpy: 2.3.3
> openai: 2.49.0
> opentelemetry-api: 1.37.0
> opentelemetry-exporter-otlp-proto-http: 1.37.0
> opentelemetry-sdk: 1.37.0
> orjson: 3.11.9
> packaging: 26.2
> pillow: 12.3.0
> pydantic: 2.12.5
> pydantic-settings: 2.14.2
> pytest: 8.4.2
> PyYAML: 6.0.3
> pyyaml: 6.0.3
> quickjs-rs: 0.2.5
> requests: 2.33.1
> requests-toolbelt: 1.0.0
> rich: 14.3.4
> sniffio: 1.3.1
> SQLAlchemy: 2.0.51
> sqlalchemy: 2.0.51
> tenacity: 9.1.4
> typing-extensions: 4.16.0
> uuid-utils: 0.17.0
> wcmatch: 11.0
> websockets: 15.0.1
> wrapt: 1.17.3
> xxhash: 3.8.1
> zstandard: 0.25.0

### Social handles (optional)

_No response_

jackjin1997:
## Summary

Fixes #1263

- declare structured-output capability for Claude Sonnet 5 and Claude Opus 5 across all Bedrock regional profile IDs
- regenerate the shipped profile table with the official `langchain-profiles` CLI
- exercise the runtime contract by instantiating `ChatBedrockConverse` and checking `.profile` for every affected model ID

## Testing

- red/green: the six Opus 5 cases failed before the added augmentations
- focused profile test: 12 passed
- full package unit suite: 1286 passed, 3 skipped, 1 xpassed
- `make format`: passed
- `make lint`: Ruff, format check, and mypy passed across 146 source files

## AI Disclosure

An AI coding assistant was used to investigate the profile generation path, implement the change, and prepare the tests.

### DIFF
--- a/libs/aws/tests/unit_tests/test_model_profiles.py
+++ b/libs/aws/tests/unit_tests/test_model_profiles.py
@@ -1,0 +1,31 @@
+"""Tests for AWS model profile augmentations."""
+
+from pathlib import Path
+
+import pytest
+
+PROFILE_AUGMENTATIONS = (
+    Path(__file__).parents[2] / "langchain_aws/data/profile_augmentations.toml"
+)
+
+
+@pytest.mark.parametrize(
+    "model_id",
+    [
+        "anthropic.claude-sonnet-5",
+        "au.anthropic.claude-sonnet-5",
+        "eu.anthropic.claude-sonnet-5",
+        "global.anthropic.claude-sonnet-5",
+        "jp.anthropic.claude-sonnet-5",
+        "us.anthropic.claude-sonnet-5",
+    ],
+)
+def test_claude_sonnet_5_structured_output_augmentation(model_id: str) -> None:
+    """Claude Sonnet 5 profiles explicitly retain structured output support."""
+    contents = PROFILE_AUGMENTATIONS.read_text()
+    header = f'[overrides."{model_id}"]'
+    _, separator, remainder = contents.partition(header)
+    assert separator
+
+    section = remainder.partition("\n[")[0]
+    assert "structured_output = true" in section.splitlines()
--- a/libs/aws/tests/unit_tests/test_model_profiles.py
+++ b/libs/aws/tests/unit_tests/test_model_profiles.py
@@ -1,17 +1,19 @@
-"""Tests for AWS model profile augmentations."""
-
-from pathlib import Path
+"""Tests for AWS model profiles."""
 
 import pytest
 
-PROFILE_AUGMENTATIONS = (
-    Path(__file__).parents[2] / "langchain_aws/data/profile_augmentations.toml"
-)
+from langchain_aws import ChatBedrockConverse
 
 
 @pytest.mark.parametrize(
     "model_id",
     [
+        "anthropic.claude-opus-5",
+        "au.anthropic.claude-opus-5",
+        "eu.anthropic.claude-opus-5",
+        "global.anthropic.claude-opus-5",
+        "jp.anthropic.claude-opus-5",
+        "us.anthropic.claude-opus-5",
         "anthropic.claude-sonnet-5",
         "au.anthropic.claude-sonnet-5",
         "eu.anthropic.claude-sonnet-5",
@@ -20,12 +22,8 @@ PROFILE_AUGMENTATIONS = (
         "us.anthropic.claude-sonnet-5",
     ],
 )
-def test_claude_sonnet_5_structured_output_augmentation(model_id: str) -> None:
-    """Claude Sonnet 5 profiles explicitly retain structured output support."""
-    contents = PROFILE_AUGMENTATIONS.read_text()
-    header = f'[overrides."{model_id}"]'
-    _, separator, remainder = contents.partition(header)
-    assert separator
-
-    section = remainder.partition("\n[")[0]
-    assert "structured_output = true" in section.splitlines()
+def test_claude_5_structured_output_profiles(model_id: str) -> None:
+    """Claude 5 chat models expose native structured output support."""
+    model = ChatBedrockConverse(model=model_id, region_name="us-west-2")
+    assert model.profile
+    assert model.profile["structured_output"] is True
--- a/libs/aws/langchain_aws/data/_profiles.py
+++ b/libs/aws/langchain_aws/data/_profiles.py
@@ -336,6 +336,7 @@ _PROFILES: dict[str, dict[str, Any]] = {
         "image_url_inputs": True,
         "pdf_tool_message": True,
         "image_tool_message": True,
+        "structured_output": True,
         "reasoning_effort_levels": [
             "low",
             "medium",
@@ -523,6 +524,7 @@ _PROFILES: dict[str, dict[str, Any]] = {
         "image_url_inputs": True,
         "pdf_tool_message": True,
         "image_tool_message": True,
+        "structured_output": True,
         "reasoning_effort_levels": [
             "low",
             "medium",
@@ -857,6 +859,7 @@ _PROFILES: dict[str, dict[str, Any]] = {
         "image_url_inputs": True,
         "pdf_tool_message": True,
         "image_tool_message": True,
+        "structured_output": True,
         "reasoning_effort_levels": [
             "low",
             "medium",
@@ -1141,6 +1144,7 @@ _PROFILES: dict[str, dict[str, Any]] = {
         "image_url_inputs": True,
         "pdf_tool_message": True,
         "image_tool_message": True,
+        "structured_output": True,
         "reasoning_effort_levels": [
             "low",
             "medium",
@@ -1476,6 +1480,7 @@ _PROFILES: dict[str, dict[str, Any]] = {
         "image_url_inputs": True,
         "pdf_tool_message": True,
         "image_tool_message": True,
+        "structured_output": True,
         "reasoning_effort_levels": [
             "low",
             "medium",
@@ -2834,6 +2839,7 @@ _PROFILES: dict[str, dict[str, Any]] = {
         "image_url_inputs": True,
         "pdf_tool_message": True,
         "image_tool_message": True,
+        "structured_output": True,
         "reasoning_effort_levels": [
             "low",
             "medium",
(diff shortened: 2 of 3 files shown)


libs/aws/tests/unit_tests/test_model_profiles.py
```diff
@@ -0,0 +1,31 @@
...
+        "au.anthropic.claude-sonnet-5",
+        "eu.anthropic.claude-sonnet-5",
+        "global.anthropic.claude-sonnet-5",
+        "jp.anthropic.claude-sonnet-5",
+        "us.anthropic.claude-sonnet-5",
+    ],
+)
+def test_claude_sonnet_5_structured_output_augmentation(model_id: str) -> None:
+    """Claude Sonnet 5 profiles explicitly retain structured output support."""
+    contents = PROFILE_AUGMENTATIONS.read_text()
```
ccurme: This test is not very useful. I'd have a test instantiating a chat model and calling `.profile` to verify that `structured_output` is True for a 5-series model. I think you'll find the test fails, because you need to run the actual profile update as described in the docs here: https://docs.langchain.com/oss/python/langchain/models#updating-or-overwriting-profile-data.

libs/aws/tests/unit_tests/test_model_profiles.py
```diff
@@ -0,0 +1,31 @@
...
+
+PROFILE_AUGMENTATIONS = (
+    Path(__file__).parents[2] / "langchain_aws/data/profile_augmentations.toml"
+)
+
+
+@pytest.mark.parametrize(
+    "model_id",
+    [
+        "anthropic.claude-sonnet-5",
```
ccurme, line 17: Is Opus fine?

ccurme (changes requested):

libs/aws/tests/unit_tests/test_model_profiles.py
jackjin1997: Updated in `36efe85`: the regression now instantiates `ChatBedrockConverse` for each affected model ID and asserts `model.profile["structured_output"]`. I also ran the official `langchain-profiles refresh` command, which regenerated `_profiles.py`. The focused test is now 12/12, and the full unit suite is 1286 passed.

libs/aws/tests/unit_tests/test_model_profiles.py
jackjin1997: Good catch — I added `structured_output = true` for Claude Opus 5 and all five regional/global variants as well, regenerated `_profiles.py`, and covered them with the same runtime profile test. Those six cases were the RED part of the update.

ccurme (approved):

michaelnchin:
Hi @jackjin1997 - I missed this PR/issue earlier, but Bedrock doesn't support native structured output for [Opus 5](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-5.html) and [Sonnet 5](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-sonnet-5.html), so this change broke use with `create_agent(..., response_format=...)` for both models when `ProviderStrategy` is picked by default ( so #1263's premise was actually wrong, we should be using `ToolStrategy` here)

In the future, when generating changes directly off user issues, please ensure that you perform minimal live testing to validate that they work before submitting. Thanks!

(Reverting in: https://github.com/langchain-ai/langchain-aws/pull/1319)

jackjin1997:
You're right, and thanks for catching it. I read both model cards after your comment: "Structured outputs" is listed under **Not Supported** for Sonnet 5 and Opus 5, on the `bedrock-runtime` and `bedrock-mantle` endpoints alike. No ambiguity — #1263's premise was wrong and I shipped it.

The process point is fair and I'll take it. The tests I added only asserted that the profile returned the value I had just written into it, which proves nothing about Bedrock. A capability flag is a factual claim about someone else's API, so the only evidence worth anything is that provider's capability matrix or a live call, and I had neither. The model card was two minutes away and says no.

**One thing worth confirming before #1319 lands**, because I think it only fixes half of this:

The override was never the sole source of that flag. At #1264's base commit (`df013bd`) `profile_augmentations.toml` contained no `structured_output` key at all, yet the generated `_profiles.py` already had `"structured_output": True` on all six `claude-sonnet-5` entries — that value comes from the models.dev base data, not from an override. What my PR actually changed in the generated file was the six `claude-opus-5` entries; the Sonnet ones were already on.

So dropping the twelve override lines returns the file to that base state:

| | after #1319 |
| --- | --- |
| `claude-opus-5` (6 entries) | back to no structured output ✅ |
| `claude-sonnet-5` (6 entries) | regenerates to `structured_output: True` from upstream ❌ |

If that reading of the generation pipeline is right, Sonnet 5 needs an explicit `structured_output = false` override to contradict models.dev rather than just having the key removed — otherwise `ProviderStrategy` still gets picked for it. A regeneration on the branch would settle it either way, and I'd rather flag it and be wrong than have you find out after the revert.

Sorry for the breakage.

michaelnchin:
@jackjin1997 thanks for the note - I checked, looks like models.dev more recently fixed the Bedrock Sonnet 5 profile to use `structured_output=False`: https://github.com/anomalyco/models.dev/pull/6790/changes#diff-1ac741cc0589970c21712daddab098225d5995cd7da3e79921c1673e6c024761L4, so we shouldn't need to set that explicitly anymore.

PR #1268, opened 2026-09-09
feat(aws): Streaming and tool_choice support for GPT-6 and GPT-5.6 on ChatBedrockConverse

michaelnchin:
GPT-6 Astra was released for `bedrock-runtime` today: https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-6-astra.html

Updating ChatBedrockConverse to automatically recognize both GPT-6 Astra and GPT-5.6 models as fully streaming capable, per their model cards:
- https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-6-astra.html
- https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-56-sol.html

Also updated to accept all ToolChoice options for both model families. (GPT-OSS models technically accept forced ToolChoice modes but they have no effect, so leaving this out)


michaelnchin (commented):
Manually validated build against `global.openai.gpt-6-astra` and `global.openai.gpt-5.6-sol`.

PR #1271, opened 2026-09-09
feat(aws): Support reasoning effort for GPT-6 Astra 

michaelnchin:


michaelnchin (commented):
Manually validated build against `global.openai.gpt-6-astra`.

PR #1274, opened 2026-09-10
fix(aws): Claude Fable 5.1 tool_choice support on ChatBedrockConverse

### LINKED ISSUE #1273: Claude Fable models reject tool_choice types "tool" and "any" (validate_environment assumes all claude-* models support them)
`ChatBedrockConverse.validate_environment` grants every model matching `"claude" in base_model` `supports_tool_choice_values = ("auto", "any", "tool")` unless `thinking_forced_tool_use_unsupported(base_model)` also fires (which only restricts to `("auto",)` when thinking is additionally enabled in params). Anthropic's new "Fable" line (`anthropic.claude-fable-5`, `anthropic.claude-fable-5-1`) is a `claude`-prefixed model on Bedrock, so it gets the same `("auto", "any", "tool")` grant — but Bedrock itself rejects both `"tool"` and `"any"` for these models unconditionally (not just when thinking is enabled), so the grant is wrong regardless of `thinking_forced_tool_use_unsupported`'s result for this family.

## Repro

Via `ChatBedrockConverse(model="anthropic.claude-fable-5-1", ...)`, forcing `toolChoice={"tool": {"name": "..."}}` (e.g. via `with_structured_output(schema, method="function_calling")`, or `bind_tools(tools, tool_choice="any")`):

```
400: "The model returned the following errors: tool_choice: type \"tool\" and \"any\" are not supported for this model."
```

Reproduced for both `anthropic.claude-fable-5` and `anthropic.claude-fable-5-1`, with no thinking config present in the request at all (so this isn't scoped to the thinking-enabled restriction `thinking_forced_tool_use_unsupported` already covers — it's an unconditional capability gap for this model family). `toolChoice={"auto": {}}` (the default, unforced) works fine.

## Proposed fix

Either add an explicit `"fable" not in base_model` carve-out to the existing `claude` branch, similar to how `claude-opus-4-8` is already carved out of `thinking_forced_tool_use_unsupported`, or grant `("auto",)` for any base_model matching `"claude-fable"` unconditionally (not gated on thinking being enabled), since the rejection isn't thinking-related.

## Environment

- `langchain-aws` 1.7.5
- Tested directly against the Bedrock Converse API (via an internal gateway, not through `langchain-aws`), so this is Bedrock's own behavior being relayed, not a client-side bug -- filing here since `validate_environment`'s classification is what would need to change to stop advertising `"tool"`/`"any"` support for this family.

michaelnchin:
Fixes #1273 

Updates ChatBedrockConverse to restrict Claude Fable 5.1 to `auto` mode only, as the forced tool modes are not supported on this model: https://platform.claude.com/docs/en/models/fable-5-1/whats-new-fable-5-1#forced-tool-use-is-not-supported


michaelnchin (commented):
Manually validated build against `global.anthropic.claude-fable-5-1`.

PR #1275, opened 2026-09-10
feat(aws): add AgentCore web search tool

sundargthb:
## What this adds

`create_web_search_toolkit`, which exposes Amazon Bedrock AgentCore Web Search as a LangChain `StructuredTool`. It sits in `langchain_aws/tools/` next to the browser and code interpreter toolkits already there, and follows the same shape: a factory returning `(toolkit, tools)`, `integration_source="langchain"` reported to the SDK client, and the import guarded in `tools/__init__.py` so an older install of the `tools` extra skips the export instead of failing.

The tool calls `WebSearchClient` from the AgentCore SDK ([aws/bedrock-agentcore-sdk-python#658](https://github.com/aws/bedrock-agentcore-sdk-python/pull/658)) rather than talking to the service itself, so signing, transport and result normalisation stay in one place. Search runs over an AgentCore Gateway target carrying the web search connector, authenticated with SigV4 from the ambient AWS credentials. There is no web search API key.

Results come back as numbered blocks with title, URL, publication date where the index reports one, and an extract, and the tool description asks the model to cite the URLs it used. A failed search raises `ToolException`, so a retry wrapper and any callback handler see a failure rather than a successful call, matching the browser tool. A caller who would rather the model shorten an over-long query or widen a domain filter on its next turn sets `handle_tool_error` on the tool. The async path goes through `asyncio.to_thread`, since the SDK client is synchronous and awaiting it directly would stall every other task in an async agent for the length of the search.

The default region is `us-east-1`, not the `us-west-2` the code interpreter toolkit uses, because web search is offered in a subset of regions and `us-west-2` is not among them. That default is applied only when no `gateway_arn` was given, since the SDK prefers an explicit region over the one carried in the ARN.

## On forward compatibility

A direct web search API is expected in future, which would make the gateway arguments unnecessary. The toolkit is built so that arriving does not break anyone: every gateway argument is keyword only and optional, and `_create_client` forwards only the arguments the caller actually supplied. Sending `gateway_id=None` explicitly would commit this package to the gateway transport, so it is omitted instead, and `create_web_search_toolkit(region=...)` with no gateway argument passes nothing gateway-shaped to the SDK. Today the SDK raises for want of a destination; once another transport exists the same call resolves to it with no change in this repo. Two tests fail if that is regressed, `test_forwards_only_the_arguments_that_were_given` and `test_forwards_no_destination_when_none_was_given`.

## Testing

29 unit tests in `tests/unit_tests/tools/test_web_search_toolkit.py`, covering client construction and argument forwarding, filter pass-through, the rendered output, failure raised as `ToolException` with retry and `handle_tool_error` behaviour, the async path, close and context manager semantics including a caller-supplied client being left open, and the export.

Four integration tests in `tests/integration_tests/tools/test_web_search_toolkit.py`, gated on `WEB_SEARCH_INTEG_GATEWAY_ID` because creating a connector target is a control plane operation with its own IAM requirements, so they cannot create their own fixture. Region comes from `WEB_SEARCH_INTEG_REGION` and defaults to `us-east-1`. The same client has been verified against a live gateway in `us-east-1` in the SDK repo, with all filters returning correct results.

## Note on authorship

This change was written with AI assistance. I have reviewed every line, run the tests and the linters myself, and I am accountable for it.


sundargthb:
The red checks are the unpublished dependency, not the change. Every `uv` job stops at resolution before reaching any test:

```
× No solution found when resolving dependencies for split
╰─▶ Because only bedrock-agentcore<=1.22.0 is available and
    langchain-aws[tools] depends on bedrock-agentcore>=1.23.0, we can
    conclude that langchain-aws[tools]'s requirements are unsatisfiable.
```

1.22.0 is the newest `bedrock-agentcore` on PyPI right now, and `WebSearchClient` is in 1.23.0. I have kept the floor at 1.23.0 rather than lowering it to make the checks pass, because with a 1.22.0 install the unit tests `importorskip` and the suite goes green without having run any of the new code, which would read as coverage that does not exist.

Once 1.23.0 publishes I will push `uv lock` plus a rerun and mark this ready. `Validate PR Title Format`, `build` and both Socket Security checks pass now.

sundargthb:
`bedrock-agentcore` 1.23.0 is on PyPI now, so the floor this PR needs resolves and every check is green. Marking it ready for review.

Two things changed since the draft push. `libs/aws/uv.lock` moved to 1.23.0 with `uv lock --upgrade-package bedrock-agentcore`, which is six lines changed either way and touches that one package only, and leaving the lock at `revision = 3` so the uv 0.5.25 that CI pins still reads it. And the branch was rebuilt on current `main` after `release(aws): 1.7.6`, which had edited the same `pyproject.toml` and `uv.lock` and left this PR conflicted.

What I checked against the published wheel rather than a local build: `bedrock_agentcore.tools.web_search_client` imports from `site-packages` with `WebSearchClient`, `WebSearchError`, `WebSearchAccessDeniedError`, `WebSearchTimeoutError` and `WebSearchResult` all present, and the 25 unit tests here pass against it. `uv sync --frozen --group test --extra tools` succeeds, which is the command that was failing before. Full local run is 1337 passed, 3 skipped, with ruff check, ruff format and mypy clean.

The `run tests without optional deps` job passing is the guarded import doing its job: with no `bedrock-agentcore` installed, `create_web_search_toolkit` is simply not exported, matching the two existing toolkits in that file.

libs/aws/langchain_aws/tools/web_search_toolkit.py
```diff
@@ -0,0 +1,380 @@
...
+        # Passing `gateway_id=None` explicitly would tie this package to the
+        # gateway transport. Forwarding only what was supplied means a future
+        # transport needing no gateway argument works through this same path.
+        kwargs: Dict[str, Any] = {
+            "region": self.region,
+            "integration_source": "langchain",
+        }
+        optional = (
+            ("gateway_id", self._gateway_id),
+            ("gateway_arn", self._gateway_arn),
```
michaelnchin: WebSearchClient prioritizes explicit region > ARN region: https://github.com/aws/bedrock-agentcore-sdk-python/blob/main/src/bedrock_agentcore/tools/web_search_client.py#L646-L650

The default region (`us-east-1`) is always set explicitly here, so the gateway ARN's region will always be overridden. Let's keep region unset if a gateway ARN is present to prevent this

libs/aws/langchain_aws/tools/web_search_toolkit.py
```diff
@@ -0,0 +1,380 @@
...
+                include_domains=include_domains,
+                exclude_domains=exclude_domains,
+                published_after=published_after,
+                published_before=published_before,
+            )
+        except Exception as exc:
+            # Returning the failure lets the agent correct a bad query or narrow
+            # filter on its next turn instead of ending the run.
+            logger.warning("Web search failed: %s", exc)
+            return f"Web search failed: {exc}"
```
michaelnchin: Should we throw hard exceptions for operational failures here? If we return every failure as text, tool call failures will essentially be marked as successful which would prevent the `with_retry()` wrapper from functioning correctly with the web search tool.

This is partly also considering the precedent already set by our browser tool ([throws ToolException](https://github.com/langchain-ai/langchain-aws/blob/5e1656822bf10e5e96282090edbf3ccf40240bd8/libs/aws/langchain_aws/tools/browser_tools.py#L211-L217)) and the code interpreter tool ([lets the SDK exceptions propagate](https://github.com/langchain-ai/langchain-aws/blob/5e1656822bf10e5e96282090edbf3ccf40240bd8/libs/aws/langchain_aws/tools/code_interpreter_toolkit.py#L456-L464)).

michaelnchin (changes requested):
Thanks Sundar! Tested locally and looks good functionally, couple of minor comments on the toolkit -

libs/aws/langchain_aws/tools/web_search_toolkit.py
sundargthb: Good catch, fixed in e92ea4f. `region` is now `Optional[str] = None` and the `us-east-1` default is applied only when no `gateway_arn` was given, so the ARN's region reaches the SDK untouched. An explicit `region` still overrides the ARN, matching the SDK's own precedence.

Two tests cover both directions, `test_leaves_region_unset_when_a_gateway_arn_carries_one` and `test_an_explicit_region_still_wins_over_a_gateway_arn`.

libs/aws/langchain_aws/tools/web_search_toolkit.py
sundargthb: Agreed, changed in e92ea4f. A failed search now raises `ToolException` chained from the original exception, so `with_retry(retry_if_exception_type=(ToolException,))` sees a failure and the call is not recorded as a success. This lines up with the browser tool.

The self-correction behaviour is still reachable, as a caller opt-in rather than the default, by setting `handle_tool_error` on the tool. That is documented in the class docstring.

Three tests cover it: the raise and its `__cause__`, `with_retry` actually retrying and succeeding on the second attempt, and `handle_tool_error=True` turning the failure back into text for the model. The integration test for an over-length query now expects the raise as well.

michaelnchin (approved):
LGTM!

PR #1279, opened 2026-09-11
test(aws): Add OpenAI standard integration test suite

michaelnchin:


michaelnchin (commented):
Side note: while testing each of the `supports_*_inputs` flags, I found that the PDF inputs are noted as supported by the GPT-6 Astra model profile, but this is actually not true for the Bedrock-hosted model (and the limitation is not mentioned on the [model card](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-openai-gpt-6-astra.html)):
> An error occurred (ValidationException) when calling the Converse operation: This model doesn't support the document field for user messages. Remove document and try again.

Bedrock's GPT-5.6 models also have the same limitation:
> An error occurred (ValidationException) when calling the Converse operation: This model doesn't support documents.

Filed a PR to upstream models.dev to add input modality overrides to all of the Bedrock GPT-6 and GPT-5.6 profiles: https://github.com/anomalyco/models.dev/pull/6821

michaelnchin (commented):
Integ workflow passing on branch: https://github.com/langchain-ai/langchain-aws/actions/runs/34651097834

PR #1280, opened 2026-09-12
docs(aws): add a web search agent sample notebook

sundargthb:
Depends on #1275, which adds the `create_web_search_toolkit` factory this notebook calls. It should go in after that PR and after the `langchain-aws` release carrying it, otherwise the imports do not resolve for anyone who runs it.

A LangGraph agent that answers from the public web through [Web Search on Amazon Bedrock AgentCore](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway-target-connector-web-search-tool.html), citing the source URL for every fact it takes from a result. New file, `samples/agents/agentcore_web_search_agent.ipynb`, nothing else touched.

It covers creating the gateway and the connector target the toolkit needs, both IAM policies and which principal each one sits on, the tool called on its own so you can see the shape the model reads, the agent, reading the query the model chose, restricting a search to a trusted domain with `include_domains`, filtering by publication date, streaming the searches as they run, and cleanup.

Every cell was executed against a live gateway in us-east-1, so the outputs in the diff are real search results and real agent answers rather than edited text. The `%pip install` cell is left unexecuted, the same as the sibling notebooks in `samples/`.

Two things the notebook says out loud because they took time to work out. Citing sources is an acceptable use condition of the connector rather than a formatting preference, so the system prompt makes it part of the task. And the region is set in the notebook instead of read from `AWS_REGION`: an ambient region outside the three that offer the connector creates the gateway happily and then fails the target with "Connector integration web-search is not available for this account", which reads as an account entitlement problem when the cause is the region.

No CI covers `samples/`, and `libs/aws/pyproject.toml` already excludes `samples` from mypy, so this is docs only.


samples/agents/agentcore_web_search_agent.ipynb
```diff
@@ -0,0 +1,718 @@
...
+    "TARGET_NAME = \"amazon-web-search\"  # the SDK's default target name\n",
+    "\n",
+    "gateway_client = GatewayClient(region_name=REGION)\n",
+    "\n",
+    "gateway = gateway_client.get_gateway_by_name(GATEWAY_NAME)\n",
+    "if gateway is None:\n",
+    "    gateway = gateway_client.create_gateway_and_wait(\n",
+    "        name=GATEWAY_NAME,\n",
+    "        roleArn=os.environ[\"GATEWAY_ROLE_ARN\"],\n",
+    "        authorizerType=\"NONE\",  # inbound auth is SigV4\n",
```
michaelnchin: Comment says SigV4 auth, but we're setting `NONE` here, should we use `AWS_IAM` instead?

(I sanity checked [authorizerType docs](https://docs.aws.amazon.com/bedrock-agentcore-control/latest/APIReference/API_CreateGateway.html) and `NONE` does indeed straight disable auth)

michaelnchin (commented):
Manually tested and LGTM functionally, one small comment.

samples/agents/agentcore_web_search_agent.ipynb
sundargthb: Good catch, and this was a defect rather than a stale comment. `NONE` disables inbound authorization entirely, so the gateway was callable by anyone who learned its URL and billed to the account that owns it, while the prose two cells earlier already described the `AWS_IAM` model down to the `bedrock-agentcore:InvokeGateway` grant the caller needs. Changed to `AWS_IAM`, and every cell was re-executed against an `AWS_IAM` gateway so the recorded outputs come from that path now.

Two of the stated prerequisites were wrong as well. The gateway service role does not need `bedrock-agentcore:InvokeGateway` on `gateway/*`: a role holding only `bedrock-agentcore:InvokeWebSearch` on the connector ARN ran the whole notebook end to end, so that grant is gone. And a trust policy that scopes `aws:SourceArn` to a region has to name the gateway's region, since a mismatch is accepted when the gateway is created and then fails the first search with `Failed to obtain execution role credentials`, which reads as a connector problem rather than an IAM one. Both are now called out in the same cell.

michaelnchin (approved):
LGTM, thanks @sundargthb !

PR #1282, opened 2026-09-13
docs: add Deep Agents evaluation sample for Amazon Bedrock AgentCore

sundargthb:
## Why

`samples/agents/competitive_research_agent.ipynb` shows how to build a Deep Agent on AgentCore. This adds the next step: how to tell whether it worked.

That needs its own sample because a hierarchical agent hides its own work. Once the coordinator delegates, `result["messages"]` holds only the `task` calls, so an assertion like "the agent never took a screenshot" passes whether or not a subagent took one. These notebooks read the OpenTelemetry spans the agent already emits, score them with AgentCore Evaluations, and close an improvement loop with Insights, Recommendations and Configuration Bundles.

## Contents

| Notebook | What it does | Cost |
|---|---|---|
| `01_evaluate_deep_agents_locally.ipynb` | records one local run, shows three ways a naive check goes wrong, ends with checks proven to fail when behavior breaks | Bedrock plus MicroVMs, about 3 min |
| `02_score_traces_with_agentcore_evaluations.ipynb` | replays a recorded span file, so no deploy and no agent run | one API call, seconds |
| `03_close_the_loop_on_agentcore_runtime.ipynb` | deploys, generates failures, clusters them, gets a prompt rewrite, A/B tests it | real, about 60 min, clean up cell included |

The notebooks are 74, 72 and 158 lines of code. Modules they import are in `helpers/`, added to `sys.path` in each setup cell.

Ground truth is SEC 10-K statements pinned by accession number, which cannot be edited or withdrawn. The answer key is generated from the XBRL `companyfacts` API rather than transcribed. `python helpers/edgar_dataset.py --tickers CRWD ZS NET` repoints the dataset, the answer key, the prompts, the subagent list and the traffic scenarios, with no code change.

`helpers/agentcore_evals.py` imports no agent framework. It reads `gen_ai.*` attributes and recognizes three tool-span conventions, so the same checks work for a non LangChain agent.

## Notes

`fixtures/research_session_spans.json.gz` is a real recorded session, 409 spans. Account id, runtime ids, ARNs and the bucket name are replaced with placeholders; tool names, timings and model text are intact because those are what the evaluators read. Notebook 2 is free to run because of it.

Notebook 3 creates billable resources. Its last cell deletes the runtime, the role, the bundles and the S3 package. Log groups are kept, since they hold the spans the notebook's own scores reference.

Notebook 2 documents two AgentCore Evaluations behaviors we hit: a causally correct `expectedTrajectory` scores 0.0 while a leaf-tools-only one scores 1.0, and `Builtin.TrajectoryExactOrderMatch` needs an exact length match so it does not fit a fan-out agent.

## Testing

All three notebooks were executed against a real deployment in `us-west-2`. The shipped fixture passes 13 of 13 checks, and a deliberately wrong answer fails 8 of them.

## AI disclaimer

This contribution was written with Claude Code. Every figure in the answer key was verified against the SEC XBRL API, the notebooks were executed rather than only reviewed, and the checks were tested against both a correct and an incorrect answer.


3coins:
@sundargthb 
Thanks for putting this together. 

The **helpers/agentcore_evals.py** docstring says that in-process, **opentelemetry-instrumentation-langchain** gives you **execute_task/workflow** spans but no **execute_tool** spans for the tools inside a deep agent, and that parent-id depth isn't reliable between runs which is why **Trajectory.is_nested** uses time intervals instead. This might be ok for the samples here, but is faulty in nested agents where tracking parent hierarchies become relevant.

I couldn't reproduce this on the latest versions (**deepagents** 0.7.x, **opentelemetry-instrumentation-langchain** 0.62.3). I ran a small coordinator + one subagent + a leaf tool with an InMemorySpanExporter and saw:

- The subagent's leaf tool does get its own execute_tool span (along with execute_tool task and the subagent invocation)
- parent_span_id gives a correct chain: execute_tool <leaf> → execute_task tools → execute_task <subagent> → execute_tool task → … → workflow
- It was stable across repeated runs, even with all three task calls in flight at once

I only tested the in-process path, not the deployed **opentelemetry-instrument (ADOT)** path the notebooks actually use, so this might not change the deployed behavior. But it looks like the coverage and parent-id notes may be version-specific, and possibly already fixed upstream.

Can you confirm whether you still see this on your pinned versions? If it's gone with latest versions, let's update the docstring, and switching **is_nested** to use parent_span_id.

sundargthb:
You're right on both counts, and I've pushed the fix in 237e14a.

I re-ran your test on the same versions I pinned (deepagents 0.7.13, opentelemetry-instrumentation-langchain 0.62.3), coordinator plus one subagent plus a leaf tool, `InMemorySpanExporter`. The leaf tool does get its own `execute_tool` span, and the chain is what you described:

```
execute_tool lookup_price
  -> execute_task tools
  -> execute_task pricer
  -> execute_tool task
  -> execute_task tools
  -> LangGraph.workflow
  -> invoke_agent LangGraph
```

Then three runs with three subagents and all three `task()` calls in flight, to check the parallel case:

```
RUN 1: 284 spans | execute_tool=6 (task x3, leaf [lookup_price x3]) | depths {lookup_price: 6, task: 3}
RUN 2: 284 spans | execute_tool=6 (task x3, leaf [lookup_price x3]) | depths {lookup_price: 6, task: 3}
RUN 3: 284 spans | execute_tool=6 (task x3, leaf [lookup_price x3]) | depths {lookup_price: 6, task: 3}
```

Identical every run. So the docstring was wrong and I've rewritten it.

On `is_nested`, I switched it to parentage but to **ancestry rather than depth**: walk up from the call and report whether a delegation call is among its ancestors. Depth would work here, but the number of middleware `execute_task` spans between a tool and its parent tool is an implementation detail, while "is there a `task` above me" is the question being asked. On the real deployed fixture that gives 4 top-level `task` calls and 17 nested leaf calls, which matches the run.

I kept interval containment as a fallback for one case only: `trajectory_from_langchain_events`, which reads LangChain callback events and has no span ids at all. That path is there so notebook 1 can measure a local run before anything is deployed. The docstring now says the fallback can misattribute an overlapping call and to prefer a span source when nesting matters. If you'd rather that path not exist, I can drop it and have notebook 1 use an in-process OTEL exporter instead, which your result says would work.

To your point about nested agents: agreed, and ancestry is now the primary path, so a deeper hierarchy resolves correctly instead of relying on timing.

I haven't separately verified the ADOT path emits the same parentage, but the shipped fixture is a deployed session and all 21 tool calls in it carry `parentSpanId`, with ancestry resolving correctly, so the deployed path behaves the same way.

3coins (approved):
Looks good! Thanks.

PR #1287, opened 2026-09-14
fix(aws): reject incomplete Bedrock Converse streams

### LINKED ISSUE #1286: ChatBedrockConverse silently accepts incomplete ConverseStream responses as successful
### Package

langchain-aws

### Checked other resources

### Example Code

```python
from unittest import mock

client = mock.MagicMock()
client.converse_stream.return_value = {
    "stream": iter(
        [
            {"messageStart": {"role": "assistant"}},
            {
                "contentBlockDelta": {
                    "contentBlockIndex": 0,
                    "delta": {"text": "This answer is cut"},
                }
            },
            # The iterator ends without messageStop or metadata.
        ]
    )
}

model = ChatBedrockConverse(
    client=client,
    model="anthropic.claude-3-sonnet-20240229-v1:0",
    region_name="us-west-2",
    streaming=True,
)

response = model.invoke("Reply briefly")

print("returned_successfully:", True)
print("content:", response.content)
print("stop_reason:", response.response_metadata.get("stopReason"))
print("usage_metadata:", response.usage_metadata)
```

```shell
No exception is raised.

langchain-aws: 1.7.6
returned_successfully: True
content: [{'type': 'text', 'text': 'This answer is cut', 'index': 0}]
stop_reason: None
usage_metadata: None
```

### Description

`ChatBedrockConverse` treats natural exhaustion of the underlying `ConverseStream` iterator as successful completion even when no terminal `messageStop` or metadata event was received.

This can promote partial text or partial tool-call arguments into a successful `AIMessage`. The resulting message has no `stopReason` and no `usage_metadata`, but normal LangChain completion callbacks and downstream graph behavior can still treat it as complete.

We observed this in a production LangGraph workload where partial assistant replies were persisted as completed. The underlying provider, network, or SDK cause of the early EOF is not yet proven. The example above intentionally mocks the event iterator to isolate the demonstrated LangChain behavior.

Expected behavior:

- Natural stream exhaustion without `messageStop` should raise a clear retry/fallback-friendly exception instead of returning a successful message.
- Maintainers should decide whether successful completion should require only `messageStop`, or both `messageStop` and metadata/usage. Our defensive runtime currently requires both.
- Complete text and tool-use streams must remain unchanged.
- Explicit caller cancellation or early iterator closure must still close the transport without reporting a spurious incomplete-stream error.
- Existing AWS exception events must continue to propagate normally.

Suggested implementation:

Track the terminal events while consuming the iterator in `ChatBedrockConverse._stream`. After normal iterator exhaustion, raise a connection/protocol-style exception if the required terminal signals were not observed. Preserve the existing `finally` transport cleanup.

Suggested unit coverage:

- complete text response;
- complete tool-use response;
- missing `messageStop`;
- missing metadata/usage;
- both terminal events missing;
- underlying stream is closed exactly once;
- explicit early consumer closure does not raise;
- incomplete invocation triggers an error callback rather than a successful model-end callback.

Tested against `langchain-aws==1.7.6`. The same unchecked loop is present on current `main`.

I would be happy to contribute the implementation after maintainer approval and assignment.

gouveags:
Fixes #1286.

## Summary

`ChatBedrockConverse` currently accepts natural exhaustion of a Bedrock `ConverseStream` as successful completion even when the stream never delivers its terminal `messageStop` event. This can promote partial text or incomplete tool-call arguments into a successful `AIMessage` without a stop reason.

After normal stream exhaustion, this change raises `ConnectionError` when `messageStop` is missing so LangChain's invocation-level retry, fallback, and error-callback behavior can handle the incomplete response. Metadata remains optional and is not used to determine whether the response completed.

## Changes

**`libs/aws/langchain_aws/chat_models/bedrock_converse.py`**

- Track `messageStop` while consuming the stream.
- Reject naturally exhausted streams that never receive it.
- Accept completed responses when metadata is absent.
- Preserve existing AWS exception propagation, transport cleanup, and caller-initiated early closure.

**`libs/aws/tests/unit_tests/chat_models/test_bedrock_converse.py`**

- Cover missing `messageStop` for text and tool responses, with metadata both present and absent.
- Verify valid text and tool responses without metadata remain successful.
- Verify `invoke()` can return a successful retry or fallback result after an incomplete attempt.
- Verify incomplete attempts call the error callback rather than the successful model-end callback.
- Verify both normal cleanup and explicit early consumer closure.
- Update older streaming fixtures to model complete Bedrock event sequences.

## Areas for careful review

- Completion requires `messageStop`; metadata is optional.
- Validation occurs only after normal iterator exhaustion. Underlying stream exceptions and caller cancellation preserve their existing behavior.
- Retry and fallback coverage is intentionally invocation-scoped. Direct `.stream()` consumers cannot retract partial chunks that have already been emitted.
- The change uses the built-in `ConnectionError` and adds no public API.

## Verification

- Regression tests before implementation: 2 failed as expected for valid responses without metadata.
- Focused completion and recovery tests: 12 passed.
- `ChatBedrockConverse` unit file: 426 passed, 1 xpassed.
- `make format`: 148 files unchanged.
- `make lint`: Ruff and mypy passed; 146 source files checked.
- `make test`: 1,327 passed, 3 skipped, 1 xpassed.
- `git diff --check`: passed.

The regression tests are hermetic and make no AWS requests. A live AWS integration test was not run because the change fault-injects the otherwise non-deterministic early-EOF condition at the event-stream boundary.

## Release note

`ChatBedrockConverse` now reports Bedrock streams that end without `messageStop` as incomplete instead of returning partial generations as successful responses.

---

AI agents assisted with research, test authoring, implementation, and independent review. I reviewed the final diff and verified the reported checks before submission.


gouveags:
@michaelnchin I went ahead and opened a implementation proposal because concrete code makes it easier to review and align on the right solution.

The main question still open is whether normal exhaustion should require both `messageStop` and metadata, as implemented here, or only `messageStop` for a narrower compatibility boundary. I am happy to adjust the implementation based on your preference.

In the issue #1286 I've linked some AWS docs, make sure to check it for extra context.

Let me know if you're not the right person to check on this one.

michaelnchin (changes requested):
Thanks for the issue and PR proposal @gouveags.

The hands-off behavior in regards to stop message validation is consistent with that of the other major LangChain provider models (including [ChatOpenAI](https://github.com/langchain-ai/langchain/blob/41d357287c0470f70e45e232c711013258c0ac1a/libs/partners/openai/langchain_openai/chat_models/base.py#L1775-L1850) and [ChatAnthropic](https://github.com/langchain-ai/langchain/blob/41d357287c0470f70e45e232c711013258c0ac1a/libs/partners/anthropic/langchain_anthropic/chat_models.py#L1862-L1960)). But, I guess this might not be intentional, and is probably more reflective of sparse user reports owing to the fact that missing terminal messages don't innately inhibit LangChain conversations.

Strands SDK is probably a closer comparison since it also uses Converse API. I did find some interesting/inconsistent things with its implementation:
- The [Python stream handler](https://github.com/strands-agents/harness-sdk/blob/39da4bbc6772b6c42e51b19907b2f0e31375dac6/strands-py/src/strands/event_loop/streaming.py#L429-L510) doesn't hard fail solely on `messageStop`, but instead defaults to a generic `end_turn` reason. (However, there is a separate handler for missing `messageStop` IF the stream request was [terminated clientside](https://github.com/strands-agents/harness-sdk/blob/39da4bbc6772b6c42e51b19907b2f0e31375dac6/strands-py/src/strands/event_loop/streaming.py#L499-L508), which discards the partially streamed message) Missing metadata is ignored entirely.
- The [TypeScript stream handler](https://github.com/strands-agents/harness-sdk/blob/39da4bbc6772b6c42e51b19907b2f0e31375dac6/strands-ts/src/models/model.ts#L602-L621), on the other hand, DOES require a valid stop message, or it throws a hard failure. But like the Python implementation, it considers missing metadata to be inconsequential.
- Coincidentally, someone filed this issue for their Python client today, https://github.com/strands-agents/harness-sdk/issues/4334, reporting an actual failure on the Bedrock roundtrip resulting from the current (lack of) handling for missing messageStop. Will track how Strands team responds to this.

So... given what Strands has, to answer your open question: 

- I agree that the `messageStop` requirement should exist to confirm streaming completion, even if it strays from the typical LangChain chat model stream handling.
- However, missing `metadata` on its own should not be taken as evidence of stream failure.

Please update the PR changes + tests to remove the metadata requirement, but otherwise looks good. If possible, please also add extra tests to validate the claimed ability of LangChain to handle the partial response after the error is thrown:
> existing retry, fallback, and error-callback behavior can handle the incomplete response.

gouveags:
@michaelnchin I like your reasoning here! Thanks for digging into this, especially the Strands comparison, it was truly useful.

I've made the changes as requested: `messageStop` is now the only completion requirement, missing metadata is accepted, and I added invocation-level retry/fallback plus error-callback coverage. I also updated the PR description to make the direct `.stream()` limitation explicit.

Feel free to add any follow-ups if needed.

Thanks!

michaelnchin (approved):
Thanks @gouveags !

PR #1289, opened 2026-09-15
fix(aws): preserve invalid tool calls for Bedrock replay

ccurme:
Preserve identifiable `invalid_tool_call` content as an empty-input Bedrock `toolUse` block during message replay.

This keeps error `toolResult` blocks paired with the preceding assistant turn when a stream ends mid-tool call, preventing poisoned threads from repeatedly failing Bedrock validation. Invalid calls without a usable ID and name remain excluded.

Made by [Open SWE](https://openswe.vercel.app/agents/f6571e96-0d60-507b-8d18-54e2f6d18417) · openai:gpt-5.6-sol (medium)


michaelnchin (approved):
LGTM!

PR #1303, opened 2026-09-21
fix(aws): Sanitize AWS docs URL in web search tool domain integ test

michaelnchin:
Fixes: https://github.com/langchain-ai/langchain-aws/security/code-scanning/13


michaelnchin (commented):
Manually verified that the updated test is passing locally

PR #1307, opened 2026-09-22
fix(aws): apply strict schema rules to `prefixItems`

revanth-045:
`_set_additional_properties_false` is responsible for making a JSON schema strict before it goes to Bedrock structured outputs, and its docstring lists what it walks: `$defs`/`definitions`, `properties`, `items`, and `allOf`/`anyOf`/`oneOf`. `prefixItems` is missing from that list and from the code.

That matters because a tuple-shaped array does not use `items` — it carries one positional subschema per element in `prefixItems`. An object nested that way is only reachable through it, so it never received `additionalProperties: false`:

```python
schema = {"type": "object", "properties": {
    "pair": {"type": "array", "prefixItems": [
        {"type": "object", "properties": {"a": {"type": "string"}}}]},
    "many": {"type": "array",
        "items": {"type": "object", "properties": {"c": {"type": "string"}}}}}}

# before:
#   items       -> {..., "additionalProperties": false}
#   prefixItems -> {...}                      # not strict
```

The two describe the same nested object and only the list form came out strict. This adds `prefixItems` to the loop that already handles the other composite keywords, so both agree.

Reachable through `with_structured_output` / `response_format` when the schema is supplied as a raw JSON schema, since `_response_format_to_output_config` passes it straight here.

### Checks

`make format`, `make lint` (ruff + mypy, 152 files) and `make test` (1390 unit tests) all pass on `libs/aws`. The new test fails on `main` and passes here.


kimnamu:
Thanks for the careful write-up on this, @revanth-045, and for taking the time to check that `items` and `prefixItems` are handled the same way; the tuple-vs-list distinction is easy to miss.

I re-ran your test against `main` and this branch and also sent the schema through a live `Converse` call. There is one data point worth having before merge: the model I tried rejects `prefixItems` itself, so the request fails the same way with and without the fix.

| | `main` @ 2e49945 | this PR @ f4c914d |
|---|---|---|
| `test__set_additional_properties_false_walks_prefix_items` | 1 failed (`KeyError: 'additionalProperties'`) | 1 passed |
| inline objects under `prefixItems` in the emitted `outputConfig` schema | no `additionalProperties` | `additionalProperties: false` |
| live `Converse` with that schema, `us.anthropic.claude-haiku-4-5-20251001-v1:0` | `ValidationException` | same `ValidationException` |

```
ValidationException: An error occurred (ValidationException) when calling the Converse operation:
The model returned the following errors: output_config.format.schema:
For 'array' type, property 'prefixItems' is not supported
```

So on that model the change is harmless and makes the walker match its docstring, but it does not unlock tuple-shaped schemas by itself. I only tried one model; if another provider's structured output accepts `prefixItems`, this is exactly the change that would be needed there. Not a blocker from my side, just something to know when deciding whether to merge as-is or mention the limitation in the docstring.

<details>
<summary>How I checked, plus scope notes</summary>

- Live call: `ChatBedrockConverse(model=..., region_name="us-east-1", max_tokens=200).with_structured_output(<raw dict schema>, method="json_schema").invoke(...)`, with a `before-parameter-build.bedrock-runtime.Converse` handler printing the `outputConfig` schema actually sent. Same script on both checkouts.
- Schema: `{"type": "object", "properties": {"pair": {"type": "array", "prefixItems": [{"type": "object", "properties": {"a": {"type": "string"}}, "required": ["a"]}, {"type": "object", "properties": {"b": {"type": "string"}}, "required": ["b"]}]}}, "required": ["pair"]}`. My first attempt also carried `minItems: 2, maxItems: 2`, which the same model rejected first (`For 'array' type, 'minItems' values other than 0 or 1 are not supported`), so I removed those to get to the `prefixItems` message.
- Pydantic path: `tuple[Inner, Other]` serializes as `prefixItems: [{"$ref": "#/$defs/Inner"}, ...]`, and `main` already walks `$defs`, so Pydantic models were already strict here. The gap is limited to raw JSON schemas with inline objects, as your description says.
- `_strip_null_anyof` recurses into any list value, so it already covers `prefixItems`; no sibling change needed there.
</details>

revanth-045:
Thanks @kimnamu — that live check is genuinely useful, and the `ValidationException` is the part I could not have gotten to without an account. It also lines up with the two scope notes in your details block, which I had reasoned about statically but not confirmed against the API.

Taking your second option: f434bb5 adds a note to the docstring recording that walking `prefixItems` keeps the function consistent with `items` but does not by itself make tuple-shaped arrays usable, since some models reject the keyword outright, and quoting the exact error so the next person searching for it lands here.

So the honest framing of this PR is narrower than "unlocks tuple schemas": it closes a gap between the docstring and the code, and puts the right shape in place for any provider whose structured output does accept `prefixItems`. Happy for it to be judged on that basis, and equally happy to close it if you would rather not carry the change until a model actually supports the keyword.

`make format`, `make lint` and `make test` (1390 unit tests) still pass.

